PR #12, opened 2017-03-21
Feature/itbl 1090 fcm upgrade

davidtruong:
Registration and disabling with Firebase works. Backwards compatible with GCM.

IP: Still needs to add in testing.


davidtruong:
@andrewboni @dontgitit @: Ready to review

Included unit tests and verified different variants of the GCM/FCM server credentials. I added additional fields in the userProfile to indicate which send platform the user registered with to verify the upgrade.

The SDK upgrade is backwards compatible with the GCM. The previously registered GCM tokens are automatically disabled the first time the user registers for FCM.

I'm currently writing up some tech documentation on the migration strategy. Notes here: https://iterable.quip.com/y1BVAlRJzu16

PR #14, opened 2017-07-14
Feature/itbl 2397 in app consume

davidtruong:


iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -594,18 +594,14 @@ public void trackInAppOpen(int campaignId, int templateId, String messageId) {
...
-    public void trackInAppClick(int campaignId, int templateId, String messageId, int buttonIndex) {
+    public void trackInAppClick(String messageId, int buttonIndex) {
         JSONObject requestJSON = new JSONObject();
 
         try {
             requestJSON.put(IterableConstants.KEY_EMAIL, _email);
-            requestJSON.put(IterableConstants.KEY_CAMPAIGN_ID, campaignId);
-            requestJSON.put(IterableConstants.KEY_TEMPLATE_ID, templateId);
             requestJSON.put(IterableConstants.KEY_MESSAGE_ID, messageId);
             requestJSON.put(IterableConstants.ITERABLE_IN_APP_BUTTON_INDEX, buttonIndex);
```
dontgitit: unrelated, but is this supposed to matter/affect anything? seems to be unused in backend code

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -616,6 +612,24 @@ public void trackInAppClick(int campaignId, int templateId, String messageId, in
...
+
+        try {
+            addEmailOrUserIdToJson(requestJSON);
+            requestJSON.put(IterableConstants.KEY_MESSAGE_ID, messageId);
+        }
+        catch (JSONException e) {
+            e.printStackTrace();
+        }
+
+        sendPostRequest(IterableConstants.ENDPOINT_INAPP_CONSUME, requestJSON);
```
dontgitit: this doesn't really make sense; you're sending the request even if adding args fails?

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
davidtruong: It's not needed so I'm removing it from the api call for consistency.

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
davidtruong: You're right. I'll update all the calls to create the Json from a [hashmap](https://developer.android.com/reference/org/json/JSONObject.html#JSONObject(java.util.Map)) to avoid the try/catch block jsonObject exception.

dontgitit: it's fine as is, just move the `sendPostRequest` into the `try`?

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
davidtruong: Will fix in a later commit ticket: https://iterable.atlassian.net/browse/ITBL-2713

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
davidtruong: Will fix in a later commit ticket: https://iterable.atlassian.net/browse/ITBL-2713

PR #16, opened 2017-07-19
Fix for button click not working with an empty action.

davidtruong:
Simple fix to remove the null check.


davidtruong:
The empty check shouldn’t be there. It caused a bug where buttons without a defined action wouldn’t close the dialog.

-DT

PR #17, opened 2017-07-21
Android notifications with images.

davidtruong:


iterableapi/src/main/java/com/iterable/iterableapi/IterableNotification.java
```diff
@@ -28,6 +32,37 @@ protected IterableNotification(Context context) {
...
+            public void run() {
+                if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.JELLY_BEAN) {
+                    final RemoteViews bigContentView = notification.bigContentView;
+                    try {
+                        Class picassoClass = Class.forName(IterableConstants.PICASSO_CLASS);
+                        if (picassoClass != null) {
+                            Picasso.with(mContext).load(imageUrl).into(bigContentView, bigIconId, requestCode, notification);
+                        }
+                    } catch (ClassNotFoundException e) {
+                        IterableLogger.w(TAG, "ClassNotFoundException: Check that picasso is added " +
```
dontgitit: is it ok to ask people to rely on that?

davidtruong: Technically this isn't needed since we embed that reference in our SDK export. However I thought it would be good to have it in there just in case that changes in the future.

PR #19, opened 2017-10-18
<In-Progress> Feature/itbl 2763 html in app notifcation

davidtruong:


davidtruong:
Starting the review now, still need to wrap up some unit tests.

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -526,22 +527,160 @@ public void disablePush(String iterableAppId, String projectNumber, String pushS
      * @param clickCallback
      */
     public void spawnInAppNotification(final Context context, final IterableHelper.IterableActionHandler clickCallback) {
+        String htmlString;
+
+        htmlString = "<!DOCTYPE HTML PUBLIC \"-//W3C//DTD XHTML 1.0 Transitional //EN\" \"http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd\"><html xmlns=\"http://www.w3.org/1999/xhtml\" xmlns:v=\"urn:schemas-microsoft-com:vml\" xmlns:o=\"urn:schemas-microsoft-com:office:office\"><head><!--[if gte mso 9]><xml><o:OfficeDocumentSettings><o:AllowPNG/><o:PixelsPerInch>96</o:PixelsPerInch></o:OfficeDocumentSettings></xml><![endif]--><meta http-equiv=\"Content-Type\" content=\"text/html; charset=utf-8\"/><meta name=\"viewport\" content=\"width=device-width\"/><!--[if !mso]><!--><meta http-equiv=\"X-UA-Compatible\" content=\"IE=edge\"/><!--<![endif]--><title>Simple</title><style type=\"text/css\" id=\"media-query\">\n" +
```
dontgitit: is there a better place to put/format this?

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -558,6 +697,8 @@ public void getInAppMessages(int count, IterableHelper.IterableActionHandler onC
         try {
             addEmailOrUserIdToJson(requestJSON);
             requestJSON.put(IterableConstants.ITERABLE_IN_APP_COUNT, count);
+            requestJSON.put(IterableConstants.KEY_PLATFORM, IterableConstants.ITBL_PLATFORM_ANDROID);
+            requestJSON.put(IterableConstants.ITBL_KEY_SDK_VERSION, "0.0.0");
```
dontgitit: probably want something other than `0.0.0`... and can you pull the value from some other config or something? seems weird to have it hardcoded here

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppHTMLNotification.java
```diff
@@ -0,0 +1,322 @@
...
+    }
+
+    /**
+     * Returns the gravity for the given padding
+     * @param padding
+     * @return
+     */
+    int getLocation(Rect padding) {
+        int gravity = Gravity.CENTER_VERTICAL;
+        if (padding.top  == 0 && padding.bottom < 0) {
```
dontgitit: extra whitespace

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppManager.java
```diff
@@ -219,6 +247,38 @@ public static JSONObject getNextMessageFromPayload(String payload) {
...
+
+        return rect;
+    }
+
+    /**
+     * Retrieves the padding percentage
+     * @param jsonObject
+     * @return
+     */
+    static int decodePadding(JSONObject jsonObject) {
```
dontgitit: this seems to return a display percentage?

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppHTMLNotification.java
```diff
@@ -0,0 +1,322 @@
...
+        });
+    }
+
+    /**
+     * Returns the gravity for the given padding
+     * @param padding
+     * @return
+     */
+    int getLocation(Rect padding) {
+        int gravity = Gravity.CENTER_VERTICAL;
```
dontgitit: what about horizontal?

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppHTMLNotification.java
```diff
@@ -0,0 +1,322 @@
...
+            }
+        });
+    }
+
+    /**
+     * Returns the gravity for the given padding
+     * @param padding
+     * @return
+     */
+    int getLocation(Rect padding) {
```
dontgitit: the whole percentage/padding/-1 thing seems kind of jank

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
davidtruong: This should actually be removed. Missed it from the diff.

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppHTMLNotification.java
davidtruong: The horizontal position gets handled through the offset calculation in resize. I'll add some documentation to that.

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppHTMLNotification.java
davidtruong: Any issues with this function specifically?
Basically 0 represents a fixed dialog location, and -1 represents automatic resizing for that side of the dialog.

I think that decodePadding could be cleaned up so that we pass down -1 directly from the server rather than the Autoexpand/displayOption.

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppManager.java
davidtruong: @JonSEng: For V2 What are your thoughts on changing the server API to just return -1 rather than displayOption/AutoExpand?
Changing that on the server is still backwards compatible with the SDK since I get the percentage.

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
davidtruong: Removed

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
davidtruong: Fixed

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppHTMLNotification.java
davidtruong: Fixed

PR #25, opened 2018-01-05
Feature/itbl 3821 os notification payload

davidtruong:
Updated the SDK to match the backend pushService. OS based notifications should not get rendered by the SDK when no body field is present in the data payload.

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableNotification.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableNotification.java
@@ -243,4 +243,19 @@ public class IterableNotification extends NotificationCompat.Builder {
 
         return isGhostPush;
     }
+
+    /**
+     * Returns if the given notification has an empty body
+     * @param extras
+     * @return
+     */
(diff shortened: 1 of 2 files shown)


iterableapi/src/main/java/com/iterable/iterableapi/IterableNotification.java
```diff
@@ -243,4 +243,19 @@ static boolean isGhostPush(Bundle extras) {
...
+
+    /**
+     * Returns if the given notification has an empty body
+     * @param extras
+     * @return
+     */
+    static boolean isEmptyBody(Bundle extras) {
+        String notificationBody = "";
+        if (extras.containsKey(IterableConstants.ITERABLE_DATA_KEY)) {
+            String iterableData = extras.getString(IterableConstants.ITERABLE_DATA_KEY);
```
dontgitit: what's this? doesn't seem like you ever use it

davidtruong: removed excess variable

PR #28, opened 2018-03-06
Android image load fix

davidtruong:
Fixed up loading sequence for android notifications which can have a null bigContentView.

Changed logic for handling the title payload.


iterableapi/src/androidTest/java/com/iterable/iterableapi/IterableNotificationTest.java
```diff
@@ -51,6 +52,27 @@ public void testGhostPush() throws Exception {
...
+        notif.putString(IterableConstants.ITERABLE_DATA_BODY, body);
+
+        getContext().getApplicationInfo().icon = android.R.drawable.sym_def_app_icon;
+
+        IterableNotification iterableNotification = IterableNotification.createNotification(getContext(), notif, Application.class);
+        IterableNotification.postNotificationOnDevice(appContext, iterableNotification);
+        assertEquals("IterableAPI", iterableNotification.mContentTitle);
+//        It looks like mNotificationManager.notify(iterableNotification.requestCode, iterableNotification.build());
+//        is the culprit here for the flaky tests. This thread is spun up by the android system. Unless we do dependency injection and mock the notificationManager, it'll be hard to make this unflake.
+        Thread.sleep(1000);
```
davidtruong: Eventually these sleep tests will get removed in favor of mocked calls to create the notification and make network calls.

iterableapi/src/main/java/com/iterable/iterableapi/IterableNotification.java
```diff
@@ -8,25 +8,30 @@
...
+
 /**
  * Created by David Truong dt@iterable.com
  */
 public class IterableNotification extends NotificationCompat.Builder {
     static final String TAG = "IterableNotification";
     private boolean isGhostPush;
     private String imageUrl;
+    private String title;
+    private String content;
```
dontgitit: it doesn't look like this is ever used?

dontgitit: or rather, you use it here: `.setSummaryText(this.content)`
but i don't see where you initialize it?

iterableapi/src/main/java/com/iterable/iterableapi/IterableNotification.java
```diff
@@ -37,32 +42,30 @@ protected IterableNotification(Context context, String channelId) {
...
-                        if (picassoClass != null) {
-                            Picasso.with(mContext).load(imageUrl).into(bigContentView, bigIconId, requestCode, notification);
-                        }
-                    } catch (ClassNotFoundException e) {
-                        IterableLogger.w(TAG, "ClassNotFoundException: Check that picasso is added " +
-                                "to the build dependencies", e);
-                    }
-                }
+        InputStream in;
+        if (this.imageUrl != null) {
```
dontgitit: i'm confused, why is everything in here? shouldn't only the bitmap stuff be in here? why is title/content in here as well? won't this make it so that it only loads a notification if the notification comes with an image?

or does all that happen in `createNotification`? if it happens in `createNotification`, then why is this method resetting the `bigContentTitle` and `summaryText`?

iterableapi/src/main/java/com/iterable/iterableapi/IterableNotification.java
davidtruong: These only get used if an image is passed onto the build method to use for the BigPictureStyle notification.

iterableapi/src/main/java/com/iterable/iterableapi/IterableNotification.java
davidtruong: You're right, I should move all of the style logic into build method so it is more clear. 
I created the style for the  text based notification in createNotification, but I merely passed on the params for the BigPictureStyle which was created in the build method.

iterableapi/src/main/java/com/iterable/iterableapi/IterableNotification.java
davidtruong: Removing these unnecessary params

dontgitit: so they weren't unnecessary, they were plain wrong, because you were setting stuff to `null` right?

dontgitit: nvmd i forgot you were setting them in the builder

iterableapi/src/main/java/com/iterable/iterableapi/IterableNotification.java
```diff
@@ -8,25 +8,31 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.HttpURLConnection;
+import java.net.MalformedURLException;
+import java.net.URL;
+
 /**
  * Created by David Truong dt@iterable.com
  */
 public class IterableNotification extends NotificationCompat.Builder {
```
dontgitit: should probably rename this to `IterableNotificationBuilder` since it's a `NotificationCompat.Builder`

iterableapi/src/main/java/com/iterable/iterableapi/IterableNotification.java
```diff
@@ -8,25 +8,31 @@
...
     static final String TAG = "IterableNotification";
     private boolean isGhostPush;
     private String imageUrl;
+
+    //This parameter necessary for backwards compatability < Android 23
```
dontgitit: `// This parameter is necessary for backwards compatibility < Android 23`

(fix whitespace, add `is`, typo in `compatability`)

iterableapi/src/main/java/com/iterable/iterableapi/IterableNotification.java
dontgitit: is there a `setImage` method or something? it's still strange to me that we call `setTicker`, `setContentTile`, etc in `createNotification`, and then suddenly all the `image`-related stuff happens here instead?

davidtruong: There is no setImage method, that image is attached to the notification via the style.

The image load is done inside of the build function because that is run asynchronously, whereas the createNotification is run on the main thread.

dontgitit: how does the notification get shown then? (re asynchronous) - can it be shown before the image is fetched?

davidtruong: The notification object gets created in the main thread in IterablePushReceiver, but the build() gets executed in a background thread and added to the device in postNotificationOnDevice. This is done because you can't make network calls on the main thread.

On the background thread the image load is guaranteed to process prior to adding it to the device.

dontgitit: ah, that makes sense

iterableapi/src/main/java/com/iterable/iterableapi/IterableNotificationBuilder.java
```diff
@@ -8,61 +8,90 @@
...
-                        Class picassoClass = Class.forName(IterableConstants.PICASSO_CLASS);
-                        if (picassoClass != null) {
-                            Picasso.with(mContext).load(imageUrl).into(bigContentView, bigIconId, requestCode, notification);
-                        }
-                    } catch (ClassNotFoundException e) {
-                        IterableLogger.w(TAG, "ClassNotFoundException: Check that picasso is added " +
-                                "to the build dependencies", e);
-                    }
-                }
+        InputStream in;
```
dontgitit: any reason this is declared here and not on line 77 where it's used?

iterableapi/src/main/java/com/iterable/iterableapi/IterableNotificationBuilder.java
```diff
@@ -8,61 +8,90 @@
...
+        if (this.imageUrl != null) {
+            try {
+                URL url = new URL(this.imageUrl);
+                HttpURLConnection connection = (HttpURLConnection) url.openConnection();
```
dontgitit: is this cast necessary?

iterableapi/src/main/java/com/iterable/iterableapi/IterableNotificationBuilder.java
```diff
@@ -8,61 +8,90 @@
...
+                IterableLogger.e(TAG, e.toString());
+            } catch (IOException e) {
+                IterableLogger.e(TAG, e.toString());
             }
-        });
-        return notification;
+        }
+
+        //The default style if the BigPictureStyle isn't set.
+        if (this.mStyle == null) {
```
dontgitit: what's `mStyle`? is that like `getStyle` or something?

davidtruong: Exactly its the internal style. I used that because there is no public getStyle accessor.

iterableapi/src/main/java/com/iterable/iterableapi/IterableNotificationBuilder.java
davidtruong: no particular reason

dontgitit: care to move it then? 😂

iterableapi/src/main/java/com/iterable/iterableapi/IterableNotificationBuilder.java
davidtruong: Checking to see if I can use the base class for URLConnection instead.

davidtruong: Looks like I could just use URLConnection, since i don't need to update the headers.

dontgitit: 👍 
why'd you have the cast then?

davidtruong: Copy paste from my other HttpUrlConnection code, which i figured did the same thing. Didn't realize I didn't need all of the functionality.

dontgitit: ah aight. let's also fix that one if it doesn't need it

iterableapi/src/main/java/com/iterable/iterableapi/IterableNotificationBuilder.java
```diff
@@ -8,61 +8,89 @@
...
+    protected IterableNotificationBuilder(Context context, String channelId) {
         super(context, channelId);
     }
 
+    /**
+     * Sets the image url
+     * @param imageUrl
+     */
+    public void setImageUrl(String imageUrl) {
+        if (imageUrl != null) {
```
dontgitit: any particular reason for the null check? and why is there one here, but not in `setExpandedContent` below?

davidtruong: Removed, it will be null either way with or without the check.

iterableapi/src/main/java/com/iterable/iterableapi/IterableNotificationBuilder.java
davidtruong: Done and pushed.

iterableapi/src/main/java/com/iterable/iterableapi/IterableNotificationBuilder.java
```diff
@@ -8,61 +8,85 @@
...
+                URLConnection connection = url.openConnection();
+                connection.setDoInput(true);
+                connection.connect();
+                Bitmap myBitmap = BitmapFactory.decodeStream(connection.getInputStream());
+                this.setStyle(new NotificationCompat.BigPictureStyle()
+                        .bigPicture(myBitmap)
+                        .setSummaryText(expandedContent));
+            } catch (MalformedURLException e) {
+                IterableLogger.e(TAG, e.toString());
+            } catch (IOException e) {
```
dontgitit: `IOException` and `MalformedURLException` seem reasonable... do we also need to do anything in case the url provided isn't to an image file? for example, maybe they messed up pasting the url, or maybe it's in some weird format that `BitmapFactory` doesn't support? (btw, what formats does it support? i hope not just `.bmp`? looks like your old code used `Picasso` to load images, why'd that change?)

davidtruong: I removed Picasso because it loads the image into the the BigContentView after the notification is already created. However with the newer Android OS that view is not guaranteed to exist. So we now have to load the image prior to building the notification.

I'll take a look at adding in some more error handling for that case where it is not an image.

dontgitit: alright. we should probably add that error handling before release, don't want app crashes. is there a list somewhere of formats that `BitmapFactory` supports?

davidtruong: According to the documentation if the stream is null or can't be decoded it returns null. So it should work as is. Passing null to bigPicture just shows a blank image.
https://developer.android.com/reference/android/graphics/BitmapFactory.html

davidtruong:
@dontgitit: I think the BitmapFactory question was the last of the PR comments. If you approve it I can release it tomorrow morning.

iterableapi/src/main/java/com/iterable/iterableapi/IterableNotificationBuilder.java
dontgitit: ah, that's cool. maybe we shouldn't set `bigPictureStyle` if we can't load the image, and just proceed to the regular text style?

dontgitit:
yeah, i think maybe we don't want to render an empty image since i'm assuming it looks ugly

iterableapi/src/main/java/com/iterable/iterableapi/IterableNotificationBuilder.java
davidtruong: Perfect, was thinking the same thing.

PR #29, opened 2018-03-27
[ITBL-2713] Don't send the request if constructing JSON request body failed

vbabenkoru:


iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -451,12 +451,13 @@ public void updateEmail(String newEmail) {
             try {
                 requestJSON.put(IterableConstants.KEY_CURRENT_EMAIL, _email);
                 requestJSON.put(IterableConstants.KEY_NEW_EMAIL, newEmail);
+
+                sendPostRequest(IterableConstants.ENDPOINT_UPDATE_EMAIL, requestJSON);
+                _email = newEmail;
```
dontgitit: @davidtruong @vbabenkoru does this need to happen in some sort of success callback of `sendPostRequest`? what if it fails?

davidtruong: @dontgitit: Yes it is on our radar. @vbabenkoru and I agreed that this will get fixed up in the next PR when we work on adding in callbacks handlers.

PR #32, opened 2018-04-02
[ITBL-4329] Add purchase tracking

vbabenkoru:


iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -390,6 +391,35 @@ public void track(String eventName, int campaignId, int templateId, JSONObject d
         }
     }
 
+    public void trackPurchase(double total, List<CommerceItem> items) {
```
davidtruong: Can you make sure to add comments to the functions.

vbabenkoru:
Added method documentation

iterableapi/src/main/java/com/iterable/iterableapi/CommerceItem.java
```diff
@@ -0,0 +1,27 @@
+package com.iterable.iterableapi;
+
+import org.json.JSONException;
+import org.json.JSONObject;
+
+public class CommerceItem {
```
davidtruong: Can you also add in documentation for the CommerceItem?

vbabenkoru: Done

iterableapi/src/main/java/com/iterable/iterableapi/CommerceItem.java
```diff
@@ -0,0 +1,52 @@
+package com.iterable.iterableapi;
+
+import org.json.JSONException;
+import org.json.JSONObject;
+
+import java.util.List;
+
+/**
+ * Represents a product. These are used by the commerce API; see {@link IterableApi#trackPurchase(double, List, JSONObject)}
```
dontgitit: there are a bunch of optional fields, what about those?

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -390,6 +391,46 @@ public void track(String eventName, int campaignId, int templateId, JSONObject d
         }
     }
 
+    /**
+     * Tracks a purchase.
+     * @param total total purchase amount
+     * @param items list of purchased items
+     */
+    public void trackPurchase(double total, List<CommerceItem> items) {
+        trackPurchase(total, items, null);
```
dontgitit: how about using `new JSONObject()` so that you can ignore the `if` check on line 423? BTW i think server side we are ok with receiving `dataFields: null`

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -390,6 +391,46 @@ public void track(String eventName, int campaignId, int templateId, JSONObject d
...
+    }
+
+    /**
+     * Tracks a purchase.
+     * @param total total purchase amount
+     * @param items list of purchased items
+     * @param dataFields a `JSONObject` containing any additional information to save along with the event
+     */
+    public void trackPurchase(double total, List<CommerceItem> items, JSONObject dataFields) {
```
dontgitit: you should also support an override with `campaignId`

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
vbabenkoru: This would be weird if we omit the parameter when dataFields is null but send dataFields: {} when we just call the method without dataFields.

vbabenkoru: I'd expect the API to behave the same when parameter is null vs not specified.

dontgitit: yeah, i guess that's true

iterableapi/src/main/java/com/iterable/iterableapi/CommerceItem.java
vbabenkoru: The implementation was copied from the iOS version to bring feature parity to Android. Should I close this and refactor both?

vbabenkoru: Yes, that sounds reasonable. Feature parity first, API expansion later.

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
davidtruong: Chatted with Victor. We'll push this release with feature parity to iOS and then go back and improve it to support the optional api aparms.

PR #33, opened 2018-04-06
[ITBL-4453] Handle response status codes, add onSuccess & onFailure callbacks

vbabenkoru:
+ move to JUnit4 tests, improve retry logic.

### DIFF
--- a/iterableapi/src/androidTest/java/com/iterable/iterableapi/IterableApiResponseTest.java
+++ b/iterableapi/src/androidTest/java/com/iterable/iterableapi/IterableApiResponseTest.java
@@ -1,0 +1,258 @@
+package com.iterable.iterableapi;
+
+import android.os.AsyncTask;
+import android.support.test.InstrumentationRegistry;
+import android.support.test.filters.MediumTest;
+import android.support.test.filters.SmallTest;
+import android.support.test.runner.AndroidJUnit4;
+
+import org.hamcrest.CoreMatchers;
+import org.json.JSONObject;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+import java.io.IOException;
+import java.net.HttpURLConnection;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+
+import okhttp3.mockwebserver.MockResponse;
+import okhttp3.mockwebserver.MockWebServer;
+import okhttp3.mockwebserver.RecordedRequest;
+import okhttp3.mockwebserver.SocketPolicy;
+
+import static junit.framework.Assert.assertEquals;
+import static junit.framework.Assert.assertNotNull;
+import static junit.framework.Assert.assertTrue;
+import static org.junit.Assert.assertThat;
+
+@RunWith(AndroidJUnit4.class)
+@MediumTest
+public class IterableApiResponseTest {
+
+    private MockWebServer server;
+
+    @Before
+    public void setUp() {
+        createIterableApi();
+        server = new MockWebServer();
+        IterableApi.overrideURLEndpointPath(server.url("").toString());
+    }
+
+    @After
+    public void tearDown() throws IOException {
+        server.shutdown();
+        server = null;
+    }
+
+    private void createIterableApi() {
+        IterableApi.sharedInstanceWithApiKey(InstrumentationRegistry.getTargetContext(), "fake_key", "test_email");
+    }
+
+    private void stubAnyRequestReturningStatusCode(int statusCode, JSONObject data) {
+        String body = null;
+        if (data != null)
+            body = data.toString();
+        stubAnyRequestReturningStatusCode(statusCode, body);
+    }
+
+    private void stubAnyRequestReturningStatusCode(int statusCode, String body) {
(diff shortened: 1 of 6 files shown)


iterableapi/src/main/java/com/iterable/iterableapi/IterableRequest.java
```diff
@@ -115,19 +107,79 @@ protected String doInBackground(IterableApiRequest... params) {
...
+
+                // Read the response body
+                try {
+                    BufferedReader in;
+                    if (responseCode < 400) {
+                        in = new BufferedReader(
+                                new InputStreamReader(urlConnection.getInputStream()));
+                    } else {
+                        in = new BufferedReader(
+                                new InputStreamReader(urlConnection.getErrorStream()));
```
dontgitit: the output comes from a different "stream" for errors?!

iterableapi/src/main/java/com/iterable/iterableapi/IterableRequest.java
```diff
@@ -115,19 +107,79 @@ protected String doInBackground(IterableApiRequest... params) {
...
-                    java.util.Scanner scanner = new java.util.Scanner(errorStream).useDelimiter("\\A");
-                    requestResult = scanner.hasNext() ? scanner.next() : "";
-                    IterableLogger.d(TAG, "Invalid Request for: " + iterableApiRequest.resourcePath);
-                    IterableLogger.d(TAG, requestResult);
-                } else if (iterableApiRequest.requestType == IterableApiRequest.REDIRECT) {
-                    if (responseCode >= 300) {
+
+                    handleFailure(errorMessage, jsonResponse);
+                } else if (responseCode == 200) {
+                    if (requestResult.length() > 0 && error == null) {
```
dontgitit: seems `requestResult.ANYTHING` will throw null pointer if the code fails before it gets to https://github.com/Iterable/iterable-android-sdk/pull/33/files#diff-d211efd4c4061bd677c652daa51418fcR129 ?

iterableapi/src/main/java/com/iterable/iterableapi/IterableRequest.java
```diff
@@ -179,13 +255,24 @@ protected void setRetryCount(int count) {
...
+        this.apiKey = apiKey;
+        this.resourcePath = resourcePath;
+        this.json = json;
+        this.requestType = requestType;
+        this.successCallback = onSuccess;
+        this.failureCallback = onFailure;
+    }
 
-    public IterableApiRequest(String apiKey, String resourcePath, JSONObject json, String requestType, IterableHelper.IterableActionHandler callback){
+    public IterableApiRequest(String apiKey, String resourcePath, JSONObject json, String requestType, final IterableHelper.IterableActionHandler callback) {
```
davidtruong: It looks like this function can be deprecated in favor of the successCallback. Then the get requests & the getAndTrackDeeplink can pull the result string from the SuccessHandler.

We can probably remove this function since it is only called internally.

iterableapi/src/androidTest/java/com/iterable/iterableapi/IterableApiResponseTest.java
```diff
@@ -0,0 +1,258 @@
...
+@MediumTest
+public class IterableApiResponseTest {
+
+    private MockWebServer server;
+
+    @Before
+    public void setUp() {
+        createIterableApi();
+        server = new MockWebServer();
+        IterableApi.overrideURLEndpointPath(server.url("").toString());
```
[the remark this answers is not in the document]
vbabenkoru: We still need to override the endpoint in tests somehow..

iterableapi/src/main/java/com/iterable/iterableapi/IterableRequest.java
davidtruong: What if we change the underlying logic for getAndTrackDeeplink?

public static void getAndTrackDeeplink(String uri, IterableHelper.IterableActionHandler onCallback) { 
Replace 
IterableApiRequest request = new IterableApiRequest(null, uri, null, IterableApiRequest.REDIRECT, onCallback);

with 
IterableApiRequest request = new IterableApiRequest(null, uri, null, IterableApiRequest.REDIRECT, onSuccess, onFailure);

Then in the onSuccess or onFailure you can call the getAndTrackDeeplink callback. That way we could expose the onSuccess and onFailure if needed. 
public static void getAndTrackDeeplink(String uri, IterableHelper.SuccessHandler onSuccess, IterableHelper.FailureHandler onFailure) {

vbabenkoru: Success and failure handlers take JSONObjects as arguments, so that won't work with getAndTrackDeeplink.

davidtruong: I guess that's true. I forgot that I was overwriting the requestResult with the urlConnection headerField since the redirect URL is not in the response body. 

https://github.com/Iterable/iterable-android-sdk/blob/73910c4f95fcb39977ca19bacbf64bdcbf3d6f27/iterableapi/src/main/java/com/iterable/iterableapi/IterableRequest.java#L174

PR #35, opened 2018-04-13
[ITBL-4445] updateEmail: only update the current email address of the IterableAPI instance when the call is successful

vbabenkoru:


iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -483,18 +483,23 @@ public void sendPush(String email, int campaignId, Date sendAt, JSONObject dataF
...
                 requestJSON.put(IterableConstants.KEY_NEW_EMAIL, newEmail);
 
-                sendPostRequest(IterableConstants.ENDPOINT_UPDATE_EMAIL, requestJSON);
-                _email = newEmail;
+                sendPostRequest(IterableConstants.ENDPOINT_UPDATE_EMAIL, requestJSON, new IterableHelper.SuccessHandler() {
+                    @Override
+                    public void onSuccess(JSONObject data) {
+                        _email = newEmail;
+                    }
+                }, null);
```
dontgitit: what's this last `null` param?

virusman: onError. We only update the email in onSuccess.

vbabenkoru: onError. We only update the email in onSuccess.

iterableapi/src/androidTest/java/com/iterable/iterableapi/IterableApiRequestsTest.java
```diff
@@ -67,4 +71,32 @@ public void testTrackPurchaseWithDataFields() throws Exception {
...
+
+        RecordedRequest request1 = server.takeRequest();
+        assertEquals("/" + IterableConstants.ENDPOINT_UPDATE_EMAIL, request1.getPath());
+        assertEquals("{\"currentEmail\":\"test_email\",\"newEmail\":\"test@example.com\"}", request1.getBody().readUtf8());
+        Thread.sleep(100); // We need the callback to run to verify the internal email field change
+
+        server.enqueue(new MockResponse().setResponseCode(400).setBody("{}"));
+
+        // Check that we handle failures properly
+        IterableApi.sharedInstance.updateEmail("invalid_mail!!123");
```
davidtruong: Do we actually fail on invalid emails? We should explore the full logic because your initial createIterableApi uses an invalid email.

PR #42, opened 2018-06-27
[ITBL-4563] Extract redirect handing code from IterableRequest

vbabenkoru:
**Changed:** getAndTrackDeeplink was previously returning the original app link URL (if the response is 302), the URL passed to it (if it's not an Iterable link), or the error message from the response body (if it returned with code >= 400). It only returns a URL now and doesn't return the error text in case of 400 errors, same as on iOS. Is it a desired change?


davidtruong:
Definitely makes sense, I've been meaning to clean up this code to remove the extra redirect calls into a separate function. 

Since this is now a separate class can you make sure to add documentation to the function?
We should also update the MParticle Kit to maintain parity there. We can talk about that more later.

PR #43, opened 2018-06-28
[ITBL-5290] Store attribution information

vbabenkoru:
Store attribution information for 24 hours after the last push open or deep link click

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
@@ -104,7 +104,7 @@ public class IterableApi {
                 getPreferences(),
                 IterableConstants.SHARED_PREFS_ATTRIBUTION_INFO_KEY,
                 attributionInfo.toJSONObject(),
-                3600 * IterableConstants.SHARED_PREFS_ATTRIBUTION_INFO_EXPIRATION_HOURS
+                3600 * IterableConstants.SHARED_PREFS_ATTRIBUTION_INFO_EXPIRATION_HOURS * 1000
                 );
     }
 
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
@@ -85,6 +85,30 @@ public class IterableApi {
     }
 
     /**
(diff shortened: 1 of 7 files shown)


iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -84,6 +84,30 @@ public String getPayloadData(String key) {
...
+    /**
+     * Stores attribution information.
+     * @param attributionInfo Attribution information object
+     */
+    void setAttributionInfo(IterableAttributionInfo attributionInfo) {
+        IterableUtil.saveExpirableJsonObject(
+                getPreferences(),
+                IterableConstants.SHARED_PREFS_ATTRIBUTION_INFO_KEY,
+                attributionInfo.toJSONObject(),
+                3600 * IterableConstants.SHARED_PREFS_ATTRIBUTION_INFO_EXPIRATION_HOURS
```
davidtruong: Is this correct?
This is passing in hours * 60 min/hr * 60 sec/min which is the # of seconds. 

However saveExpirableValue stores currentTimeMillis() + expirationInterval. So shouldn't expirationInterval be the number of millis?

davidtruong:
Can you also add some tests for the expiration?

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
vbabenkoru: You're right, it will expire in 86 seconds. :) I'm writing tests for the expiration logic, so it should be covered and verified.

vbabenkoru:
Added a test for persistence/expiration

iterableapi/src/main/java/com/iterable/iterableapi/IterableAttributionInfo.java
```diff
@@ -0,0 +1,39 @@
...
+    public IterableAttributionInfo(int campaignId, int templateId, String messageId) {
+        this.campaignId = campaignId;
+        this.templateId = templateId;
+        this.messageId = messageId;
+    }
+
+    public JSONObject toJSONObject() {
+        JSONObject jsonObject = new JSONObject();
+        try {
+            jsonObject.put("campaignId", campaignId);
```
tapashmajumder: Extract consts "campaignId" etc.?

PR #45, opened 2018-07-02
[ITBL-5343] Add handleAppLink method

vbabenkoru:


iterableapi/src/main/java/com/iterable/iterableapi/IterableActionSource.java
```diff
@@ -0,0 +1,12 @@
+package com.iterable.iterableapi;
+
+/**
+ * Enum representing the source of the action: push notification, app link, etc.
+ */
+public enum IterableActionSource {
```
davidtruong: Should we include a source for in-app?

vbabenkoru: Yes, but I'd add this when we figure out a good way to pass in-app links to the handlers and implement that.

PR #47, opened 2018-07-19
[ITBL-4897] Improve FCM support, remove GCM

vbabenkoru:
Ready


iterableapi/src/main/java/com/iterable/iterableapi/IterablePushRegistration.java
```diff
@@ -4,142 +4,117 @@
...
-            } else if (iterablePushRegistrationData.pushRegistrationAction == IterablePushRegistrationData.PushRegistrationAction.DISABLE) {
-                pushRegistrationObject = getDeviceToken(iterablePushRegistrationData.projectNumber, iterablePushRegistrationData.messagingPlatform, iterablePushRegistrationData.iterableAppId, false);
-                if (pushRegistrationObject != null) {
-                    IterableApi.sharedInstance.disablePush(pushRegistrationObject.token);
+    protected Void doInBackground(IterablePushRegistrationData... params) {
+        IterablePushRegistrationData iterablePushRegistrationData = params[0];
+        if (iterablePushRegistrationData.pushIntegrationName != null) {
+            PushRegistrationObject pushRegistrationObject = getDeviceToken();
+            if (pushRegistrationObject != null) {
+                if (iterablePushRegistrationData.pushRegistrationAction == IterablePushRegistrationData.PushRegistrationAction.ENABLE) {
```
davidtruong: It's possible that the user upgrades to FCM and then calls disable.
That means that we should also call disableOldDeviceIfNeeded from the disable case in order to disable the old GCM token. (move disableOldDeviceIfNeeded outside the if/if else block)

iterableapi/src/main/java/com/iterable/iterableapi/IterableFirebaseMessagingService.java
```diff
@@ -0,0 +1,82 @@
...
+public class IterableFirebaseMessagingService extends FirebaseMessagingService {
+
+    static final String TAG = "itblFCMMessagingService";
+
+    @Override
+    public void onMessageReceived(RemoteMessage remoteMessage) {
+        if (remoteMessage.getData().size() > 0) {
+            Map<String,String> messageData = remoteMessage.getData();
+            handlePushReceived(messageData);
+            Log.d(TAG, "Message data payload: " + remoteMessage.getData());
```
davidtruong: Use the IterableLogger

iterableapi/src/main/java/com/iterable/iterableapi/IterableFirebaseMessagingService.java
```diff
@@ -0,0 +1,82 @@
...
+        }
+
+        // Check if message contains a notification payload.
+        if (remoteMessage.getNotification() != null) {
+            Log.d(TAG, "Message Notification Body: " + remoteMessage.getNotification().getBody());
```
davidtruong: Use the IterableLogger

iterableapi/src/main/java/com/iterable/iterableapi/IterablePushRegistration.java
```diff
@@ -4,142 +4,117 @@
 import android.content.SharedPreferences;
 import android.os.AsyncTask;
 
-import com.google.android.gms.gcm.GoogleCloudMessaging;
-import com.google.android.gms.iid.InstanceID;
+//import com.google.android.gms.gcm.GoogleCloudMessaging;
```
davidtruong: Remove commented imports

iterableapi/src/main/java/com/iterable/iterableapi/IterablePushRegistration.java
```diff
@@ -4,142 +4,117 @@
...
-        }
-
-        String deviceToken = null;
-        if (pushRegistrationObject != null) {
-            deviceToken = pushRegistrationObject.token;
+            IterableLogger.e("IterablePush", "iterablePushRegistrationData has not been specified");
         }
-
-        return deviceToken;
+        return null;
```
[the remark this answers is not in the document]
vbabenkoru: The type is `Void`, so it must return an object (`null`).

iterableapi/src/main/java/com/iterable/iterableapi/IterablePushRegistration.java
vbabenkoru: If they upgrade, they'll call `registerForPush` at some point to register the new token anyway.

iterableapi/src/main/java/com/iterable/iterableapi/IterablePushRegistration.java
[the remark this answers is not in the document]
davidtruong: Got it it was a capital `Void`.

iterableapi/src/main/java/com/iterable/iterableapi/IterablePushRegistration.java
davidtruong: My thought was about the old token. What happens if they upgrade to FCM and want to disable pushes on the device without ever attempting to register a new token?

virusman: Yeah, that makes sense. I’ll move this call.

vbabenkoru: Yeah, that makes sense. I’ll move this call.

PR #50, opened 2018-07-26
Set payload data onReceive

steelbrain:
`setPayloadData` was defined but wasn't being used. This makes sure to set the payload data on GCM messages

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterablePushActionReceiver.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterablePushActionReceiver.java
@@ -80,6 +80,7 @@ public class IterablePushActionReceiver extends BroadcastReceiver {
         }
 
         // Automatic tracking
+        IterableApi.sharedInstance.setPayloadData(intent);
         IterableApi.sharedInstance.setNotificationData(notificationData);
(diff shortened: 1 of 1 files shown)


iterableapi/src/main/java/com/iterable/iterableapi/IterablePushActionReceiver.java
```diff
@@ -36,6 +36,7 @@ public void onReceive(Context context, Intent intent) {
 
         String actionName = intent.getAction();
         if (IterableConstants.ACTION_PUSH_ACTION.equalsIgnoreCase(actionName)) {
+            IterableApi.sharedInstance.setPayloadData(intent);
```
vbabenkoru: I think it would be better to move this to line 84, before the calls to `setNotificationData` and `trackPushOpen`.

steelbrain: Good idea! Taken care of

vbabenkoru:
Awesome, thank you for catching that @steelbrain!

steelbrain:
@vbabenkoru Thoughts on a new method `getAllPayloadData` on `IterableAPI` that returns a JSON object with all the extras?

Currently we have to get each payload key one by one and it cannot be used, which doesn't work if keys are dynamic (unless we wrap the dynamic top level keys into a namespace object like "tb")

PR #51, opened 2018-07-26
Add method to get all payload data

steelbrain:
Adds a new method that returns all of payload as a JSONObject. It handles JSON errors the same way other methods do in this file.

Context: https://github.com/Iterable/iterable-android-sdk/pull/50#issuecomment-408253595


vbabenkoru:
Thanks! Isn't it enough to just return a Bundle?

steelbrain:
Bundle works, I did JSON because it's specific to my usecase but I think the conversion could live in consumerland

Updated!

vbabenkoru:
By the way, do you only need this within one app session? In the iOS version of the SDK, we store the push payload for 24 hours.

steelbrain:
Yes the payloads are instantly sent to React Native world where they are dealt with

PR #54, opened 2018-08-15
[ITBL-5598] Disable/re-register device token when email or userId changes

vbabenkoru:


iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -367,9 +367,19 @@ private static IterableApi sharedInstanceWithApiKey(Context currentContext, Stri
...
     public void setEmail(String email) {
+        if (_email != null && _email.equals(email)) {
+            return;
+        }
+
+        if (_email == null && _userId == null && email == null) {
+            return;
+        }
+
+        onLogOut();
```
davidtruong: Since IterablePushRegistration is an async call, would we run into any issues with calling onLogin right after onLogOut?

vbabenkoru: Registration is handled through AsyncTask that is using a serial queue for processing. But it should not matter, since it is enabling and disabling different devices. One edge case, though: if the app switches from a userId to email or from email to userId, and both match the same user, it may encounter a race condition and disable the device after registering.

PR #59, opened 2018-09-17
[ITBL-5995] Store additional data in device fields

vbabenkoru:
Device ID (autogenerated, persisted on first launch) - `deviceId`
App Package Name - `appPackageName`
App version - `appVersion`
App build number - `appBuild`
Iterable SDK version - `iterableSdkVersion`

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
@@ -25,6 +25,7 @@ import java.util.Date;
 import java.util.List;
 import java.util.Map;
 import java.util.TimeZone;
+import java.util.UUID;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
@@ -51,6 +52,7 @@ public class IterableApi {
     private boolean _debugMode;
     private Bundle _payloadData;
     private IterableNotificationData _notificationData;
+    private String _deviceId;
 
 //---------------------------------------------------------------------------------------
 //endregion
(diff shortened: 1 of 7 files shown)


iterableapi/src/main/java/com/iterable/iterableapi/IterableUtil.java
```diff
@@ -12,6 +15,41 @@ public static long currentTimeMillis() {
...
+        return null;
+    }
+
+    /**
+     * Maps a version string like "x.y.z" onto an integer like xxxyyyzzz
+     * Example: "1.2.3" -> 1002003
+     * @param versionString Version string
+     * @return Integer representation of the version
+     */
+    static int convertVersionStringToInt(String versionString) {
```
davidtruong: Do you have a particular use case for providing this conversion?

vbabenkoru: It is not used now, but it may be used by upgrade logic to compare versions and determine which upgrades/migrations to run.

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -1271,6 +1279,17 @@ private String getAdvertisingId() {
         return advertisingId;
     }
 
+    private String getDeviceId() {
```
dontgitit: i think you should rename this to `getOrGenerateDeviceId`, and call it once on SDK initialization. Otherwise, you might have multiple threads run it at the same time, and generate multiple ids

dontgitit: (and replace usage of `getDeviceId()` with `_deviceId` on L1156 since it'll always be set)

PR #62, opened 2018-10-05
[ITBL-6259] Deferred deep linking support

vbabenkoru:
- Added a new config option: `checkForDeferredDeeplink` to enable DDL
- Send a device fingerprint on first launch
- If matched, call the URL handler with the original link

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
@@ -1367,8 +1367,11 @@ public class IterableApi {
         IterableLogger.d(TAG, "handleDDL: " + response);
         try {
             MatchFpResponse matchFpResponse = MatchFpResponse.fromJSONObject(response);
-            IterableAction action = IterableAction.actionOpenUrl(matchFpResponse.destinationUrl);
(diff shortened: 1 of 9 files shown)


iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -1323,6 +1324,57 @@ private void onLogIn() {
...
+
+        } catch (Exception e) {
+            IterableLogger.e(TAG, "Error while checking deferred deep link", e);
+        }
+    }
+
+    private void handleDDL(JSONObject response) {
+        IterableLogger.d(TAG, "handleDDL: " + response);
+        try {
+            MatchFpResponse matchFpResponse = MatchFpResponse.fromJSONObject(response);
```
tapashmajumder: Shouldn't check isMatch from the response here?

vbabenkoru: Yeah, you're right; the fact that the URL handler was not being called was a side effect. :)

vbabenkoru: Fixed

PR #63, opened 2018-10-22
Feature/itbl 6230 call create user for user 

davidtruong:
Attempts to create a user profile when the userId is set.


vbabenkoru:
Verified to work in local dev.

PR #70, opened 2018-12-01
[MOB-54] In-App improvements

vbabenkoru:


iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppManager.java
```diff
@@ -1,279 +1,254 @@
...
+    void scheduleProcessing() {
+        if (canShowInAppAfterPrevious()) {
+            processMessages();
+        } else {
+            new Handler(Looper.getMainLooper()).postDelayed(new Runnable() {
+                @Override
+                public void run() {
+                    processMessages();
+                }
+            }, (getSecondsSinceLastInApp() + 2) * 1000);
```
tapashmajumder: What is the +2 for?

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppManager.java
```diff
@@ -1,279 +1,254 @@
...
+            }
+        }
+        scheduleProcessing();
+    }
 
-                JSONObject textJson = buttonJson.optJSONObject(IterableConstants.ITERABLE_IN_APP_CONTENT);
-                button.setTextColor(getIntColorFromJson(textJson, IterableConstants.ITERABLE_IN_APP_COLOR, Color.BLACK));
-                button.setText(textJson.optString(IterableConstants.ITERABLE_IN_APP_TEXT));
+    private void processMessages() {
+        if (!IterableActivityMonitor.getInstance().isInForeground() || isShowingInApp() || !canShowInAppAfterPrevious()) {
```
tapashmajumder: I think this check can be moved to beginning of scheduleProcessing() method to remove duplication.

If the 'if' condition is fulfilled in scheduleProcessing() and the 'if' condition is not fulfilled here then it may be possible that we will not get the next message.

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppManager.java
vbabenkoru: To account for possible timer inaccuracy, we add something to make sure it is never run before the time passes.

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppManager.java
```diff
@@ -1,279 +1,254 @@
...
+        Activity currentActivity = IterableActivityMonitor.getInstance().getCurrentActivity();
+        // Prevent double display
+        if (currentActivity != null) {
+            if (IterableInAppManager.showIterableNotificationHTML(currentActivity, message.getContent().html, message.getMessageId(), new IterableHelper.IterableActionHandler() {
+                @Override
+                public void execute(String data) {
+                    if (clickCallback != null) {
+                        clickCallback.execute(data);
+                    }
+                    if (data != null && data.startsWith("itbl://")) {
```
tapashmajumder: Is it guaranteed that all urls will be prepended with itbl:// by the UI? For iOS we can get applewebdata://abc urls.

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppManager.java
vbabenkoru: We're checking if the app is still in foreground and if nothing else is being displayed at the moment. We don't know that at the time of scheduling.

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppManager.java
vbabenkoru: `itbl://` has been a convention in the current version of in-app for internal app links, so we can keep that convention and just route it to the custom action handler.

tapashmajumder: If you have a custom action in the inApp html with <a href='action1'>Click Here</a> (no url scheme). When the user clicks this link what will the action show up as? Will it be 'action1' or 'itbl://action1' or something else?

vbabenkoru: I thought we required a URI schema for links, but if it's possible to pass anything, it's a good idea to also pass anything that doesn't have a schema to the custom action handler.

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppManager.java
tapashmajumder: I see. If at display time if the app is in foreground or something is being displayed the processing loop will end here. Is that OK?

vbabenkoru: Yes. We schedule processing whenever the app goes into foreground or after an existing in-app is closed, so it the dropped processing will be picked up at that moment.

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppManager.java
tapashmajumder: In case of iOS the url will automatically come in as `applewebdata://action1` if you forget to up `itbl://action1`. 
For Android you may want to check for either 'itbl://' or the absence of any url scheme' for custom action.

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppManager.java
tapashmajumder: Ok makes sense.

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppManager.java
tapashmajumder: Shouldn't the delay be (IN_APP_DELAY_SECONDS - getSecondsSinceLastInAPP())?

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppManager.java
vbabenkoru: On Android, invalid URLs will just become an empty string. However, this code wasn't handling `itbl://` links properly: the calling method already strips `itbl://` prefix, so we should just check if the URL doesn't contain a scheme, then it is an `itbl://` URL. Fixed that.

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppManager.java
vbabenkoru: Fixed.

PR #82, opened 2019-02-08
[MOB-129] Add documentation for new in-app messages API

vbabenkoru:


README.md
```diff
@@ -91,23 +91,81 @@ Note that `FirebaseInstanceIdService` is deprecated and replaced with `onNewToke
...
+	
+IterableConfig config = new IterableConfig.Builder()
+		.setPushIntegrationName("myPushIntegration")
+		..setInAppHandler(new MyInAppHandler())
+		.build();
+IterableApi.initialize(context, "<your-api-key>", config);
+```
+
+### Getting the local queue of in-app messages
+The SDK keeps the local in-app message queue in sync by polling the server every time the app goes into foreground, and via silent push messages that arrive from Iterable servers to notify the app whenever a new in-app message is added to the queue.
```
bradumbaugh: - I might not use the word "polling" here—maybe "checking?" Probably being overly-picky, but "polling" to me implies regular, repeated checks

README.md
```diff
@@ -91,23 +91,81 @@ Note that `FirebaseInstanceIdService` is deprecated and replaced with `onNewToke
...
+### Migrating in-app messages from the previous version of the SDK
+
+If you are already using in-app messages, you will have to make the following changes to your code:
+
+1. `spawnInAppNotification` is no longer needed and will fail to compile. In-app messages are now displayed automatically. If you need to customize the process, please refer to the sections above. To handle clicks on links in in-app messages, define `urlHandler` on `IterableConfig`.
+2. If you want to handle all in-app messages manually, as before these changes, define an `inAppHandler` on `IterableConfig` that always returns `InAppResponse.SKIP` to never show in-app messages automatically, and use `getInAppManager().getMessages()` to get the messages and `getInAppManager().showMessage(message)` to show a specific message.
+
+
+## Deep Linking
 #### Handling links from push notifications
```
bradumbaugh: This jumps from H2 to H4

README.md
bradumbaugh: - On iOS, @tapashmajumder indicated there's some custom code that needs to be called in the app delegate to tell the SDK about an incoming silent push notification, so that it does the right thing. Is that true on Android, too?

vbabenkoru: Changed to "checking the server queue".

vbabenkoru: No, push message handling is done automatically on Android (with one exception that is mentioned earlier in the section about push).

vbabenkoru:
@bradumbaugh Fixed!

PR #89, opened 2019-03-01
[MOB-72] Add an option to specify notification channel name via manifest metadata

vbabenkoru:


vbabenkoru:
Added docs for this and other fields in metadata.

PR #92, opened 2019-04-02
[MOB-291] Use action:// schema for custom actions, itbl:// for internal actions handled by the SDK

vbabenkoru:
Also, don't remove the schema when calling the click handler.

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppManager.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppManager.java
@@ -1,10 +1,8 @@
 package com.iterable.iterableapi;
 
-import android.app.Activity;
 import android.content.Context;
 
-import android.content.DialogInterface;
-import android.graphics.Rect;
+import android.net.Uri;
 import android.os.Handler;
 import android.os.Looper;
 import android.support.annotation.VisibleForTesting;
@@ -34,6 +32,8 @@ public class IterableInAppManager implements IterableActivityMonitor.AppStateCal
     private final Context context;
     private final IterableInAppStorage storage;
     private final IterableInAppHandler handler;
+    private final IterableInAppDisplayer displayer;
+    private final IterableActivityMonitor activityMonitor;
     private final double inAppDisplayInterval;
 
     private long lastSyncTime = 0;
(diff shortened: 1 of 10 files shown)


iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppManager.java
```diff
@@ -129,9 +129,12 @@ public void execute(String data) {
                         clickCallback.execute(data);
                     }
                     if (data != null && !data.isEmpty()) {
-                        if (!data.contains("://")) {
-                            // This is an itbl:// URL, pass that to the custom action handler
+                        if (data.startsWith("action://")) {
+                            // This is an action:// URL, pass that to the custom action handler
```
tapashmajumder: We should probably parse the action name before calling action handler.

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppManager.java
```diff
@@ -240,6 +243,10 @@ private boolean canShowInAppAfterPrevious() {
         return getSecondsSinceLastInApp() >= inAppDisplayInterval;
     }
 
+    private void handleInternalAction(String url) {
```
davidtruong: What is the purpose of this empty function?

davidtruong: Are you going to handle the itbl://close scenario in this pr? Or are you waiting to figure out the list supported actions?

vbabenkoru: No, this is just a placeholder for a future PR.

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppManager.java
```diff
@@ -240,6 +244,10 @@ private boolean canShowInAppAfterPrevious() {
         return getSecondsSinceLastInApp() >= inAppDisplayInterval;
     }
 
+    private void handleIterableCustomAction(String url) {
+
```
tapashmajumder: Not needed for +1 ... but I think renaming the parameter to action name instead of url will be better here.

PR #96, opened 2019-04-30
[MOB-234] Add support for Save to Inbox

vbabenkoru:
Add support for Inbox feature.

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppManager.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppManager.java
@@ -29,6 +29,10 @@ public class IterableInAppManager implements IterableActivityMonitor.AppStateCal
     static final long MOVE_TO_FOREGROUND_SYNC_INTERVAL_MS = 60 * 1000;
     static final int MESSAGES_TO_FETCH = 100;
 
+    public interface Listener {
+        void onInboxUpdated();
+    }
+
     private final IterableApi api;
     private final Context context;
     private final IterableInAppStorage storage;
@@ -37,6 +41,7 @@ public class IterableInAppManager implements IterableActivityMonitor.AppStateCal
     private final IterableActivityMonitor activityMonitor;
     private final double inAppDisplayInterval;
 
+    private final List<Listener> listeners = new ArrayList<>();
     private long lastSyncTime = 0;
     private long lastInAppShown = 0;
 
@@ -82,6 +87,43 @@ public class IterableInAppManager implements IterableActivityMonitor.AppStateCal
     }
 
     /**
+     * Get the list of inbox messages
+     * @return A {@link List} of {@link IterableInAppMessage} objects stored in inbox
+     */
+    public synchronized List<IterableInAppMessage> getInboxMessages() {
+        List<IterableInAppMessage> filteredList = new ArrayList<>();
+        for (IterableInAppMessage message : storage.getMessages()) {
+            if (!message.isConsumed() && !isMessageExpired(message) && message.isInboxMessage()) {
+                filteredList.add(message);
+            }
+        }
+        return filteredList;
+    }
+
+    /**
+     * Get the count of unread inbox messages
+     * @return Unread inbox messages count
+     */
+    public synchronized int getUnreadInboxMessagesCount() {
+        int unreadInboxMessageCount = 0;
+        for (IterableInAppMessage message : getInboxMessages()) {
+            if (!message.isRead()) {
+                unreadInboxMessageCount++;
+            }
+        }
+        return unreadInboxMessageCount;
+    }
+
+    /**
+     * Set the read flag on an inbox message
+     * @param message Inbox message object retrieved from {@link IterableInAppManager#getInboxMessages()}
+     * @param read Read state flag. true = read, false = unread
+     */
+    public synchronized void setRead(IterableInAppMessage message, boolean read) {
+        message.setRead(read);
+    }
+
+    /**
      * Trigger a manual sync. This method is called automatically by the SDK, so there should be no
      * need to call this method from your app.
      */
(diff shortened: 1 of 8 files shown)


iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppManager.java
```diff
@@ -80,6 +85,43 @@
...
+    }
+
+    /**
+     * Set the read flag on an inbox message
+     * @param message Inbox message object retrieved from {@link IterableInAppManager#getInboxMessages()}
+     * @param read Read state flag. true = read, false = unread
+     */
+    public synchronized void setRead(IterableInAppMessage message, boolean read) {
+        message.setRead(read);
+    }
```
tapashmajumder: Should this fire a change event?

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppManager.java
```diff
@@ -80,6 +85,43 @@
...
+        return unreadInboxMessageCount;
+    }
+
+    /**
+     * Set the read flag on an inbox message
+     * @param message Inbox message object retrieved from {@link IterableInAppManager#getInboxMessages()}
+     * @param read Read state flag. true = read, false = unread
+     */
+    public synchronized void setRead(IterableInAppMessage message, boolean read) {
+        message.setRead(read);
```
tapashmajumder: Should this fire a change event?

vbabenkoru: Good catch, thanks!

vbabenkoru: Fixed

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppManager.java
```diff
@@ -101,9 +144,10 @@ public void execute(String payload) {
                                     messages.add(message);
                                 }
                             }
+
+                            syncWithRemoteQueue(messages);
+                            lastSyncTime = IterableUtil.currentTimeMillis();
                         }
-                        syncWithRemoteQueue(messages);
```
davidtruong: Shouldn't we update the lastSyncTime regardless of there are new messages are set?

How exactly are we using lastSyncTime?

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppManager.java
```diff
@@ -217,7 +265,8 @@ private void processMessages() {
                 IterableLogger.d(TAG, "Response: " + response);
                 message.setProcessed(true);
                 if (response == InAppResponse.SHOW) {
-                    showMessage(message);
+                    boolean consume = !message.isInboxMessage();
+                    showMessage(message, consume, null);
```
davidtruong: Should we create another function for showMessage with showMessage(message, consume)?

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppMessage.java
```diff
@@ -141,8 +198,12 @@ static IterableInAppMessage fromJSONObject(JSONObject messageJson) {
         Trigger trigger = Trigger.fromJSONObject(triggerJson);
         JSONObject customPayload = messageJson.optJSONObject(IterableConstants.ITERABLE_IN_APP_CUSTOM_PAYLOAD);
 
+        Boolean saveToInbox = messageJson.has(IterableConstants.ITERABLE_IN_APP_SAVE_TO_INBOX) ? messageJson.optBoolean(IterableConstants.ITERABLE_IN_APP_SAVE_TO_INBOX) : null;
```
davidtruong: Any reason why you are using a Boolean and setting it to null instead of a boolean and setting it to false

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppManager.java
```diff
@@ -200,6 +247,7 @@ private void syncWithRemoteQueue(List<IterableInAppMessage> remoteQueue) {
             }
         }
         scheduleProcessing();
+        notifyOnChange();
```
tapashmajumder: Not needed for +1. Do we need to check before notifying of change? I think this is something we should take a look at in a different PR when we handle changing messages etc.

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppManager.java
vbabenkoru: I've moved this inside the `if` check because if `jsonArray` does not exist, we got an invalid response from the server, and we shouldn't consider this a successful sync.

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppManager.java
vbabenkoru: Possibly.. As a public API, or specifically for this instance?

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppMessage.java
vbabenkoru: `IterableInAppMessage` objects are serialized and stored into JSON. The Boolean is used because the original JSON may not may not have that field, and Boolean is needed to distinguish between `false` and a non-existent value.

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppManager.java
vbabenkoru: Good point! I'll add the check here.

vbabenkoru: Done

PR #112, opened 2019-06-18
[MOB-475] Fix in-app persistency

vbabenkoru:
Also, trigger in-app queue update on app launch

### DIFF
--- a/iterableapi/src/test/java/com/iterable/iterableapi/IterableInAppFileStorageTest.java
+++ b/iterableapi/src/test/java/com/iterable/iterableapi/IterableInAppFileStorageTest.java
@@ -39,4 +39,29 @@ public class IterableInAppFileStorageTest {
         storage = new IterableInAppFileStorage(RuntimeEnvironment.application);
         assertEquals(0, storage.getMessages().size());
     }
+
+    @Test
+    public void testInAppImplicitSave() throws Exception {
+        IterableInAppFileStorage storage = new IterableInAppFileStorage(RuntimeEnvironment.application);
+        IterableInAppMessage testInAppMessage = InAppTestUtils.getTestInAppMessage();
+        storage.addMessage(testInAppMessage);
+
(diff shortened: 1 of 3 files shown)


iterableapi/src/test/java/com/iterable/iterableapi/IterableInAppFileStorageTest.java
```diff
@@ -39,4 +39,29 @@ public void testInAppPersistence() throws Exception {
         storage = new IterableInAppFileStorage(RuntimeEnvironment.application);
         assertEquals(0, storage.getMessages().size());
     }
+
+    @Test
+    public void testInAppImplicitSave() throws Exception {
+        IterableInAppFileStorage storage = new IterableInAppFileStorage(RuntimeEnvironment.application);
+        IterableInAppMessage testInAppMessage = InAppTestUtils.getTestInAppMessage();
+        storage.addMessage(testInAppMessage);
```
davidtruong: Did you forget to add an assertion for this test case?

vbabenkoru: No, there are assertions below this line. The first three lines make it persist a message, and the code that follows checks that the message and its properties were stored and retrieved correctly across instances of `IterableInAppFileStorage`.

vbabenkoru: Added comments to the test.

davidtruong: What's the point of creating a new IterableInAppFileStorage(RuntimeEnvironment.application); ?

vbabenkoru: It makes it load the previously persisted data into a new instance of `IterableInAppFileStorage` to check if it's been correctly saved and restored from disk.

PR #113, opened 2019-06-19
[MOB-477] Use Api-Key header for the API key instead of api_key

vbabenkoru:


vbabenkoru:
No, our backend supports both header names, just `Api-Key` is RFC-compliant and `api_key` is not.

vbabenkoru:
Actually per RFC headers are case-insensitive, but the common convention for HTTP headers is to use capitalized words and dashes: `Content-Type`, `Content-Length`, `Host`, etc.

PR #114, opened 2019-06-19
[MOB-466] Send platform and SDK version with every API request

vbabenkoru:


iterableapi/src/main/java/com/iterable/iterableapi/IterableConstants.java
```diff
@@ -15,6 +15,8 @@
     public static final String BASE_URL_LINKS = "https://links.iterable.com/";
 
     //API Fields
+    public static final String HEADER_SDK_PLATFORM      = "SDK-Platform";
```
davidtruong: Should the constant match out casing style of other fields? sdkPlatform

vbabenkoru: The other fields are JSON fields in the request body; this one is an HTTP header. The common convention for HTTP headers is to use capitalized words and dashes: `Content-Type`, `Content-Length`, `Host`, etc.

PR #121, opened 2019-07-07
[MOB-530] Support deserializing legacy payload fields persisted by earlier beta versions

vbabenkoru:


iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppMessage.java
```diff
@@ -140,6 +140,12 @@ static IterableInAppMessage fromJSONObject(JSONObject messageJson) {
         JSONObject triggerJson = messageJson.optJSONObject(IterableConstants.ITERABLE_IN_APP_TRIGGER);
         Trigger trigger = Trigger.fromJSONObject(triggerJson);
         JSONObject customPayload = messageJson.optJSONObject(IterableConstants.ITERABLE_IN_APP_CUSTOM_PAYLOAD);
+        if (customPayload == null) {
+            customPayload = contentJson.optJSONObject(IterableConstants.ITERABLE_IN_APP_LEGACY_PAYLOAD);
```
davidtruong: Does this mean that the old payload wasn't getting utilized prior to this addition?
If so, do we even to still support this payload?

vbabenkoru: The API returns the payload in both fields. The data persisted by an older version, `beta1`, only has the old field, `payload` within `content`. Without support for the legacy field, the newer version doesn't deserialize it properly when an app is updated to a version with a newer SDK.

davidtruong: Got it

PR #122, opened 2019-07-07
[MOB-499] Track InAppOpen event on Inbox message open

vbabenkoru:


iterableapi-ui/src/main/java/com/iterable/iterableapi/ui/inbox/InboxMessageFragment.java
```diff
@@ -19,10 +19,12 @@
 
 public class InboxMessageFragment extends Fragment {
     public static final String ARG_MESSAGE_ID = "messageId";
+    public static final String LOADED_ONCE = "loadedOnce";
```
davidtruong: Personally I think this should just be LOADED instead LOADED_ONCE?

vbabenkoru: Updated

PR #124, opened 2019-08-01
[MOB-506] Add messageContext to in-app open and in-app click events

vbabenkoru:


iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -1419,6 +1495,36 @@ private void handleDDL(JSONObject response) {
         setDDLChecked(true);
     }
 
+    private JSONObject getInAppMessageContext(IterableInAppMessage message, IterableInAppLocation location) {
+        JSONObject messageContext = new JSONObject();
+        try {
+            boolean isSilentInbox = message.isInboxMessage() &&
```
davidtruong: Looks like I should add in the isInbox && never trigger type on the backend for consistency

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -1001,6 +1001,52 @@ public void trackInAppOpen(String messageId) {
...
+        } else {
+            IterableLogger.w(TAG, "trackInAppOpen: could not find an in-app message with ID: " + messageId);
+        }
+    }
+
+    /**
+     * Tracks an in-app open.
+     * @param message in-app message
+     */
+    public void trackInAppOpen(IterableInAppMessage message, IterableInAppLocation location) {
```
davidtruong: Is it necessary to have trackInAppOpen(IterableInAppMessage message)? It seems like we could have it pass in the messageId everywhere it is referenced?

I'd expect trackInAppOpen(String messageId ...) to be public rather than trackInAppOpen(IterableInAppMessage message...)
IE: if they are rendering the in-app themselves and calling trackInAppOpen they might only have reference to the messageId.

vbabenkoru: `InAppManager` is always enabled in the SDK, so `IterableInAppMessage` instances are always going to be available, and developers should be using IterableInAppMessage instances when implementing custom rendering of the messages.

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
Ayyanchira: Also, comparing to iOS code, I think 'isSilentInbox' can be a member variable of IterableInAppMessage class as a computed variable instead of being declared inside this function.

PR #130, opened 2019-08-26
[Mob-592] add in app deleted event

Ayyanchira:
Changing method signature has a bit of ripple effect. Please check if the methods are called at appropriate places


iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -1174,6 +1174,22 @@ void trackInAppDelivery(IterableInAppMessage message) {
      * @param messageId
      */
     public void inAppConsume(String messageId) {
+        IterableInAppMessage message = getInAppManager().getMessageById(messageId);
+        if (message != null){
```
vbabenkoru: Space

PR #138, opened 2019-08-30
[MOB-681] Remove api_key query parameter from GET calls, replace with a header

vbabenkoru:


iterableapi/src/androidTest/java/com/iterable/iterableapi/IterableApiRequestsTest.java
```diff
@@ -51,15 +51,18 @@ public void testPostRequestHeaders() throws Exception {
...
         assertEquals(IterableConstants.ITBL_KEY_SDK_VERSION_NUMBER, request.getHeader(IterableConstants.HEADER_SDK_VERSION));
+        assertEquals("fake_key", request.getHeader(IterableConstants.HEADER_API_KEY));
     }
 
     @Test
     public void testGetRequestHeaders() throws Exception {
         IterableApi.sharedInstance.getInAppMessages(1, null);
         RecordedRequest request = server.takeRequest(1, TimeUnit.SECONDS);
         assertTrue(request.getPath().startsWith("/" + IterableConstants.ENDPOINT_GET_INAPP_MESSAGES));
+        assertFalse(request.getPath().contains("api_key"));
```
Ayyanchira: Is this line to check if we are not using the old one?

vbabenkoru: Yeah, I removed the constant so I have use a literal value here.

iterableapi/src/androidTest/java/com/iterable/iterableapi/IterableApiRequestsTest.java
```diff
@@ -75,7 +78,6 @@ public void testTrackPurchase() throws Exception {
         RecordedRequest request = server.takeRequest(1, TimeUnit.SECONDS);
         assertEquals("/" + IterableConstants.ENDPOINT_TRACK_PURCHASE, request.getPath());
         assertEquals(expectedRequest, request.getBody().readUtf8());
-        assertEquals("fake_key", request.getHeader(IterableConstants.HEADER_API_KEY));
```
Ayyanchira: Just curious for this deletion as this would not have caused a problem... would it? Is it because testTrackPurchase has nothing to do with header check?

vbabenkoru: Yeah, I moved all header checks to the relevant tests instead, and this one was in an unrelated test..

PR #140, opened 2019-09-06
[MOB-689] Fix read/unread state update

vbabenkoru:
`DiffUtil` is comparing objects within arrays, but since the in-app messages themselves are mutable, we were effectively comparing the updated `IterableInAppMessages` to themselves. This PR changes `InboxRecyclerViewAdapter` to use an immutable `InboxRow` class instead.

### DIFF
--- a/iterableapi-ui/src/main/java/com/iterable/iterableapi/ui/inbox/InboxRecyclerViewAdapter.java
+++ b/iterableapi-ui/src/main/java/com/iterable/iterableapi/ui/inbox/InboxRecyclerViewAdapter.java
@@ -2,6 +2,7 @@ package com.iterable.iterableapi.ui.inbox;
 
 import android.net.Uri;
 import android.support.annotation.NonNull;
+import android.support.v4.util.ObjectsCompat;
 import android.support.v7.util.DiffUtil;
 import android.support.v7.widget.RecyclerView;
 import android.view.LayoutInflater;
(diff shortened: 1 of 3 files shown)


iterableapi-ui/src/main/java/com/iterable/iterableapi/ui/inbox/InboxRecyclerViewAdapter.java
```diff
@@ -17,18 +18,19 @@
...
+import java.util.ArrayList;
 import java.util.Date;
 import java.util.List;
 
 public class InboxRecyclerViewAdapter extends RecyclerView.Adapter<InboxRecyclerViewAdapter.ViewHolder> {
 
     private static final String TAG = "InboxRecyclerViewAdapter";
 
-    private List<IterableInAppMessage> values;
+    private List<InboxRow> values;
```
Ayyanchira: What are values? Does it represent a message?

Ayyanchira: Just a suggestion, how about variable name to be:- **inboxItems**?

vbabenkoru: @Ayyanchira Done!

PR #146, opened 2019-09-18
MOB 706 Improve Logging in Android SDK

Ayyanchira:
Added method in Iterable Logger to have verbose implementation.
Added a generic method to callout from any Iterable class to understand from which class, method and line number the call is made.
Called the generic functions from few of Iterable API's key functionality like trackInAppOpen, trackInAppClose, showMessage, inappClose etc enabling developers to understand the internal flow of SDK
Added IterableLog statements to Log Request and Response with proper indentation for JSON objects.

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableLogger.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableLogger.java
@@ -9,37 +9,51 @@ public class IterableLogger {
 
     public static void d(String tag, String msg) {
         if (isLoggableLevel(Log.DEBUG)) {
-            Log.d(tag, msg);
+            Log.d(tag, " 💚 " + msg);
         }
     }
 
     public static void d(String tag, String msg, Throwable tr) {
         if (isLoggableLevel(Log.DEBUG)) {
-            Log.d(tag, msg, tr);
+            Log.d(tag, " 💚 " + msg, tr);
+        }
+    }
+
+    public static void v(String tag, String msg) {
+        if (isLoggableLevel(Log.VERBOSE)) {
+            Log.v(tag, msg);
         }
     }
 
     public static void w(String tag, String msg) {
         if (isLoggableLevel(Log.WARN)) {
-            Log.w(tag, msg);
+            Log.w(tag, " 🧡️ "+ msg);
         }
     }
 
     public static void w(String tag, String msg, Throwable tr) {
         if (isLoggableLevel(Log.WARN)) {
-            Log.w(tag, msg, tr);
+            Log.w(tag, " 🧡 " + msg, tr);
         }
     }
 
     public static void e(String tag, String msg) {
         if (isLoggableLevel(Log.ERROR)) {
-            Log.e(tag, msg);
+            Log.e(tag, " ❤️ "+ msg);
         }
     }
 
     public static void e(String tag, String msg, Throwable tr) {
         if (isLoggableLevel(Log.ERROR)) {
-            Log.e(tag, msg, tr);
+            Log.e(tag, " ❤️ "+ msg, tr);
+        }
+    }
+
+    public static void printInfo() {
+        try {
+            Log.v("Iterable Call", " 💛 " + Thread.currentThread().getStackTrace()[3].getFileName() + " => " + Thread.currentThread().getStackTrace()[3].getClassName() + " => " + Thread.currentThread().getStackTrace()[3].getMethodName() + " => Line #" + Thread.currentThread().getStackTrace()[3].getLineNumber());
+        }catch (Exception e) {
+            Log.e("Iterable Call", "Couldn't print info");
         }
     }
 
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableRequest.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableRequest.java
@@ -57,12 +57,12 @@ class IterableRequest extends AsyncTask<IterableApiRequest, Void, String> {
             URL url;
             HttpURLConnection urlConnection = null;
 
+            IterableLogger.v(TAG,">>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>\n");
+            String baseUrl = (iterableApiRequest.baseUrl != null && !iterableApiRequest.baseUrl.isEmpty()) ? iterableApiRequest.baseUrl : iterableBaseUrl;
             try {
-                String baseUrl = (iterableApiRequest.baseUrl != null && !iterableApiRequest.baseUrl.isEmpty()) ? iterableApiRequest.baseUrl : iterableBaseUrl;
                 if (overrideUrl != null && !overrideUrl.isEmpty()) {
                     baseUrl = overrideUrl;
                 }
-
                 if (iterableApiRequest.requestType == IterableApiRequest.GET) {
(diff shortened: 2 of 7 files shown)


iterableapi/src/main/java/com/iterable/iterableapi/IterableLogger.java
```diff
@@ -9,40 +9,50 @@
...
 
     public static void e(String tag, String msg, Throwable tr) {
         if (isLoggableLevel(Log.ERROR)) {
-            Log.e(tag, msg, tr);
+            Log.e(tag, " ❤️ "+ msg, tr);
         }
     }
 
+    public static void printInfo() {
+        Log.v(Thread.currentThread().getStackTrace()[3].getClassName()," 💛 " + Thread.currentThread().getStackTrace()[3].getFileName()+ " => " + Thread.currentThread().getStackTrace()[3].getClassName() + " => " +  Thread.currentThread().getStackTrace()[3].getMethodName() + " => Line #" +Thread.currentThread().getStackTrace()[3].getLineNumber() );
```
vbabenkoru: Is this always safe to do, including production?

iterableapi/src/main/java/com/iterable/iterableapi/IterableRequest.java
```diff
@@ -63,6 +66,9 @@ protected String doInBackground(IterableApiRequest... params) {
                     baseUrl = overrideUrl;
                 }
 
+                IterableLogger.v(TAG,">>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>\nRequest: ");
```
vbabenkoru: We log `GET Request`/`POST Request` after this anyway, I think we can remove `\nRequest: ` from this line.

iterableapi/src/main/java/com/iterable/iterableapi/IterableRequest.java
```diff
@@ -135,6 +150,11 @@ protected String doInBackground(IterableApiRequest... params) {
                 String jsonError = null;
                 try {
                     jsonResponse = new JSONObject(requestResult);
+                    IterableLogger.printInfo();
+                    IterableLogger.v(TAG,"<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<\n" +
+                                    "Response from : " + baseUrl + iterableApiRequest.resourcePath);
+                    IterableLogger.v(TAG, jsonResponse.toString(2));
+                    IterableLogger.v(TAG, "======================================");
                 } catch (Exception e) {
                     IterableLogger.e(TAG, e.getMessage(), e);
```
vbabenkoru: Can we also add the request URL to error logs?

iterableapi/src/main/java/com/iterable/iterableapi/IterableLogger.java
Ayyanchira: I am a bit doubtful too. Stacktrace[2] always gave the immediate invoking point while stack trace[3] gave description of its caller. 
Serves the purpose here, but I think it can fail if the stack trace has something popping up in between two activity, ending up showing some different classname, method name and line number.

vbabenkoru: Yeah I think just wrap this into try/catch to catch any errors if they ever happen.

Ayyanchira: Nice idea 👍

vbabenkoru: Hmm I think this might be too verbose:
`InboxFragment.java => com.iterable.iterableapi.ui.inbox.InboxFragment$SessionManager => onMessageImpressionStarted => Line #139`
I think something like this:
`InboxFragment$SessionManager => onMessageImpressionStarted:139` would contain the same information in a more compact way.

vbabenkoru: hmm actually the full log line is this: `V/com.iterable.iterableapi.ui.inbox.InboxFragment$SessionManager:  💛 InboxFragment.java => com.iterable.iterableapi.ui.inbox.InboxFragment$SessionManager => onMessageImpressionStarted => Line #139`. The tag already contains the entire class name

Ayyanchira: The classname variable in stack trace contains the package name. What we can do is avoid putting the same classname in Tag and instead use some other value (may be just a hard coded string like "Called"/"Invoked"/"IterableMethod"). What do you think?

Ayyanchira: The entire package name can be beneficial (if not repeated) on the same line 🤷‍♂️

iterableapi/src/main/java/com/iterable/iterableapi/IterableRequest.java
```diff
@@ -133,12 +147,19 @@ protected String doInBackground(IterableApiRequest... params) {
                 // Parse JSON
                 JSONObject jsonResponse = null;
                 String jsonError = null;
+
                 try {
                     jsonResponse = new JSONObject(requestResult);
+                    IterableLogger.v(TAG,"<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<\n" +
```
vbabenkoru: This only logs if JSON could not be parsed.. There is another error handling code at lines 196 and 199. If all this code ends up looking the same, maybe extract it into a method?

iterableapi/src/main/java/com/iterable/iterableapi/IterableRequest.java
```diff
@@ -104,6 +116,8 @@ protected String doInBackground(IterableApiRequest... params) {
                     os.close();
                 }
 
+                IterableLogger.v(TAG, "======================================");
+                headerString = null;
```
vbabenkoru: Are you setting this to `null` to let it get garbage collected without waiting for the response?

iterableapi/src/main/java/com/iterable/iterableapi/IterableRequest.java
```diff
@@ -186,6 +207,16 @@ protected String doInBackground(IterableApiRequest... params) {
         return requestResult;
     }
 
+    private void addToHeaderString(HttpURLConnection urlConnection, StringBuilder headerString) {
```
vbabenkoru: Since we always build `headerString` in the caller anyways, what do you think about just taking `HttpURLConnection` and returning a `String` here? And maybe change the name to `buildHeaderString` or something similar..

iterableapi/src/main/java/com/iterable/iterableapi/IterableRequest.java
Ayyanchira: Yeah I was thinking of not having this line at all as we are assigning it a new StringBuilder() for every new request. 
Did it because I remember not getting header information from urlConnection after  making a request and thought of keeping similar functionality after we do request. Will remove the line 👍

iterableapi/src/main/java/com/iterable/iterableapi/IterableRequest.java
Ayyanchira: V.good suggestion 👍

vbabenkoru:
Can you also move `IterableLogger.e(TAG, e.getMessage(), e);` into `logError` to reduce duplication?

PR #151, opened 2019-09-30
[MOB-763] Rename 'unknown' to 'other', don't track in-app delete events for default in-app auto-consume

vbabenkoru:


iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppManager.java
```diff
@@ -199,7 +199,8 @@ public void execute(Uri url) {
         })) {
             setRead(message, true);
             if (consume) {
-                removeMessage(message, IterableInAppDeleteActionType.UNKNOWN, inAppLocation);
+                // Remove the message without tracking
+                removeMessage(message, null, null);
```
Ayyanchira: Should we have an initializer for deleteActionType somewhere? 
Setting it null will not set deleteSource in one or more API calls

```
 if (source != null) {
                requestJSON.put(IterableConstants.ITERABLE_IN_APP_DELETE_ACTION, source.toString());
           }
```

IterableAPI -> #1219

vbabenkoru: That's on purpose, see the comment. We should only track a delete event if the user has explicitly performed some delete action. In the case of auto-consume when we show normal in-apps, the user doesn't explicitly delete the message so we shouldn't pass the context or delete action type to make sure an event is not created.

Ayyanchira: Cool makes sense 👍

PR #153, opened 2019-10-07
[MOB-847] Don't stop Inbox session when an Inbox item is viewed

vbabenkoru:


iterableapi-ui/src/main/java/com/iterable/iterableapi/ui/inbox/InboxFragment.java
```diff
@@ -34,13 +35,20 @@
 
     private final SessionManager sessionManager = new SessionManager();
     private @LayoutRes int itemLayoutId = R.layout.fragment_inbox_item;
+    private boolean sessionStarted = false;
```
Ayyanchira: I think ```sessionInProgress``` would be a better term for this. What do you think?

PR #162, opened 2019-10-21
[MOB 476] Make load/save operation asynchronous

Ayyanchira:
Modified Future method by adding new function which takes in executor and runs callable on it. This allows operation like save, to create singleThreadExecutor to be passed to Future class to perform the tasks.
Save operation currently has only one instance of executor service created and maintained which is reused in case of consequent save calls.


Ayyanchira:
Creating a new ticket which has new architecture

PR #163, opened 2019-10-25
[MOB-873] Only perform automatic push registration when the app goes into foreground

vbabenkoru:


iterableapi/src/main/java/com/iterable/iterableapi/IterableActivityMonitor.java
```diff
@@ -115,6 +115,12 @@ public boolean isInForeground() {
     }
 
     public void addCallback(AppStateCallback callback) {
+        // Don't insert again if the same callback already exists
+        for (WeakReference<AppStateCallback> existingCallback : callbacks) {
+            if (existingCallback.get() == callback) {
+                return;
+            }
+        }
```
Ayyanchira: Just reviewing... So this will add to callBack list only if its unique...
👍

vbabenkoru: Yes; since `initialized` is now non-static (which it should be), and `IterableApi.initialize` can potentially be called multiple times, we have to make sure the same `IterableApi` instance can't subscribe multiple times.

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -1223,6 +1221,26 @@ protected void registerDeviceToken(String email, String userId, String applicati
...
+        @Override
+        public void onSwitchToForeground() {
+            onForeground();
+        }
+
+        @Override
+        public void onSwitchToBackground() {}
+    };
+
+    private void onForeground() {
```
Ayyanchira: onForeground feels like it itself is a callback. How about handleForegroundOperation()/ performForegroundTask() / handlePushRegistration() or something like that..?

vbabenkoru: It is a callback though, just passed from the listener object.

Ayyanchira: convinced 😁

PR #164, opened 2019-10-29
[MOB-899] Store inapp content separately

Ayyanchira:


iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFileStorage.java
```diff
@@ -14,8 +14,8 @@
 import java.util.Map;
 
 public class IterableInAppFileStorage implements IterableInAppStorage, IterableInAppMessage.OnChangeListener {
-    private static final String TAG = "IterableInAppFileStorage";
-
+    private static final String TAG = "InAppFileStorage";
+    String FOLDER_PATH = "InAppContent";
```
vbabenkoru: Since we're creating folders in the common storage shared with the app itself, it's better to prefix everything with an Iterable-specific prefix or store everything inside a folder with an Iterable-specific name so we never clash with the app or any of the libraries used by it.

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFileStorage.java
```diff
@@ -38,13 +38,15 @@ public synchronized IterableInAppMessage getMessage(String messageId) {
     @Override
     public synchronized void addMessage(IterableInAppMessage message) {
         messages.put(message.getMessageId(), message);
+        saveHTML(message.getMessageId(),message.getContent().html);
```
vbabenkoru: Space

vbabenkoru: I'm wondering if we can handle saving HTML within `save()`.. It feels weird that we save the HTML and the metadata on different steps.

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFileStorage.java
```diff
@@ -115,11 +117,82 @@ private void load() {
...
         } catch (Exception e) {
             IterableLogger.e("IterableInAppFileStorage", "Error while saving in-app messages to file", e);
         }
     }
+
+    @Override
+    public void saveHTML(String messageID, String contentHTML) {
+        File folder = createFolderIfNecessary(messageID);
+        if (folder == null) {
+            return;
```
vbabenkoru: This is an error state. Log an error in that case.

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFileStorage.java
```diff
@@ -115,11 +117,82 @@ private void load() {
...
+        }
+
+        File file = new File(folder,"index.html");
+        Boolean result = IterableUtil.instance.writeFile(file,contentHTML);
```
vbabenkoru: IterableUtil.writeFile(...) should work as well. The instance is only exposed for mocking in tests.

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFileStorage.java
```diff
@@ -115,11 +117,82 @@ private void load() {
...
+        Boolean result = folder.mkdir();
+        if (result){
+            return folder;
+        }else {
+            return null;
+        }
+    }
+
+    private File getInAppContentFolder() {
+        File context = IterableUtil.getFileDir(IterableApi.getInstance().getMainActivityContext(),FOLDER_PATH);
```
vbabenkoru: Try not to access global variables (`IterableApi.getInstance().getMainActivityContext()`) if possible.

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFileStorage.java
```diff
@@ -115,11 +117,82 @@ private void load() {
...
+    private File getFileForContent(String messageID) {
+        File folder = getFolderForMessage(messageID);
+        File file = new File(folder,"index.html");
+        return file;
+    }
+
+    @Override
+    public String getHTML(String messageID) {
+        File file = getFileForContent(messageID);
+        String contentHTML = IterableUtil.instance.readFile(file);
```
vbabenkoru: Same as above w/ `instance`.

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppMessage.java
```diff
@@ -192,6 +192,7 @@ Date getExpiresAt() {
     }
 
     public Content getContent() {
+        //TODO: Do the check here
```
vbabenkoru: Looks like this part is not finished.

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppMessage.java
```diff
@@ -259,7 +264,9 @@ static IterableInAppMessage fromJSONObject(JSONObject messageJson) {
         long expiresAtLong = messageJson.optLong(IterableConstants.ITERABLE_IN_APP_EXPIRES_AT);
         Date expiresAt = expiresAtLong != 0 ? new Date(expiresAtLong) : null;
 
-        String html = contentJson.optString(IterableConstants.ITERABLE_IN_APP_HTML);
+
+        String html =  getHTML(contentJson, storageInterface, messageId);
```
vbabenkoru: We need to load the HTML lazily: most of the time we don't even access the HTML content, so we should not be loading it here when deserializing metadata.

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppMessage.java
```diff
@@ -286,6 +293,19 @@ static IterableInAppMessage fromJSONObject(JSONObject messageJson) {
         return message;
     }
 
+    private static String getHTML(JSONObject contentJson, IterableInAppStorage storageInterface, String messageId) {
```
vbabenkoru: I think this method should not do both things conditionally, it's not very clear from its name that it is conditional.

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppMessage.java
Ayyanchira: Forgot to delete TODO. False planning 😅

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppMessage.java
```diff
@@ -286,6 +300,19 @@ static IterableInAppMessage fromJSONObject(JSONObject messageJson) {
...
+        String html;
+        if (storageInterface == null) {
```
vbabenkoru: The logic should not depend on where the object was created, but rather on which data it has. I think it may be better if it checks if the `html` field already exists in the JSON. If you load a metadata file created by an old version of the SDK, it will contain the HTML in the serialized JSON _and_ have `storageInterface` set. Then it will try to load from HTML which doesn't exist, ignoring the HTML that was passed via metadata.

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppMessage.java
```diff
@@ -21,11 +22,12 @@
     private final Trigger trigger;
     private final Boolean saveToInbox;
     private final InboxMetadata inboxMetadata;
-
     private boolean processed = false;
     private boolean consumed = false;
     private boolean read = false;
 
+    static IterableInAppStorage inAppStorageInterface;
```
vbabenkoru: Does this have to be static?

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFileStorage.java
```diff
@@ -113,13 +115,85 @@ private void load() {
...
+        if (!result) {
+            IterableLogger.e(TAG,"Failed to store HTML content");
+        }
+    }
+
+    private File createFolderIfNecessary(String messageID) {
+        File folder = getFolderForMessage(messageID);
+
+        if (folder.isDirectory()) {
+            IterableLogger.v(TAG, "Directory exists already. No need to store again");
```
vbabenkoru: What if the directory exists but there is no index.html?

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppMessage.java
Ayyanchira: Oh yes.. I forgot to delete this method. No more using this :)

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppMessage.java
Ayyanchira: Done :)

iterableapi/src/test/java/com/iterable/iterableapi/IterableInAppMessageTest.java
```diff
@@ -67,4 +76,71 @@ public void testInAppMessageOnChangeListener_consumed() throws Exception {
...
+        JSONObject payload = new JSONObject(IterableTestUtils.getResourceString("inapp_payload_multiple.json"));
+        JSONArray jsonArray = payload.optJSONArray(IterableConstants.ITERABLE_IN_APP_MESSAGE);
+        String storageInterfaceHTML = "HTML from storage interface";
+        if (jsonArray != null) {
+            for (int i = 0; i < jsonArray.length(); i++) {
+                JSONObject messageJson = jsonArray.optJSONObject(i);
+
+                IterableInAppStorage storageInterface = mock(IterableInAppStorage.class);
+                IterableInAppMessage message = IterableInAppMessage.fromJSONObject(messageJson, storageInterface);
+                Mockito.when(storageInterface.getHTML(message.getMessageId())).thenReturn(storageInterfaceHTML).thenReturn("Second Value");
```
vbabenkoru: Import `Mockito.when` and just use `when(...`

iterableapi/src/test/java/com/iterable/iterableapi/IterableInAppMessageTest.java
```diff
@@ -67,4 +76,71 @@ public void testInAppMessageOnChangeListener_consumed() throws Exception {
...
+                assertNotNull(message.getContent().html);
+                assertNotEquals(message.getContent().html, storageInterfaceHTML);
+
+                //Asserting if the storage was accessed immediately or not
+                assertEquals(storageInterface.getHTML(message.getMessageId()), storageInterfaceHTML);
```
vbabenkoru: In `assertEquals`, the first parameter is the expected value, and the second one is the value being checked.

vbabenkoru: If you want to verify that a method was called on the mock, use `verify(...)` (see test methods above).

iterableapi/src/test/java/com/iterable/iterableapi/IterableInAppMessageTest.java
```diff
@@ -67,4 +76,71 @@ public void testInAppMessageOnChangeListener_consumed() throws Exception {
...
+                assertEquals(storageInterface.getHTML(message.getMessageId()), "Second Value");
+            }
+        }
+    }
+
+    @Test
+    public void testStorageIfAccessedWhenRequired() throws Exception {
```
vbabenkoru: `WhenRequired` - when is it required? The test name should describe the situation when we expect something to happen.

iterableapi/src/test/java/com/iterable/iterableapi/IterableInAppMessageTest.java
```diff
@@ -67,4 +76,71 @@ public void testInAppMessageOnChangeListener_consumed() throws Exception {
...
+                JSONObject messageJsonHTMLStripped = new JSONObject(messageJson.toString());
+                JSONObject content = (JSONObject) messageJsonHTMLStripped.get("content");
+                content.remove("html");
+                messageJsonHTMLStripped.put("content", content);
+
+                IterableInAppStorage storageInterface = mock(IterableInAppStorage.class);
+                IterableInAppMessage message = IterableInAppMessage.fromJSONObject(messageJsonHTMLStripped, storageInterface);
+                Mockito.when(storageInterface.getHTML(message.getMessageId())).thenReturn(storageInterfaceHTML);
+                assertNotNull(message.getContent().html);
+                assertEquals(message.getContent().html, storageInterfaceHTML);
```
vbabenkoru: Maybe add a `verify` check here to make sure `getHTML` was called.

iterableapi/src/test/java/com/iterable/iterableapi/IterableInAppMessageTest.java
```diff
@@ -67,4 +76,71 @@ public void testInAppMessageOnChangeListener_consumed() throws Exception {
...
+            for (int i = 0; i < jsonArray.length(); i++) {
+                JSONObject messageJson = jsonArray.optJSONObject(i);
+                IterableInAppMessage message = IterableInAppMessage.fromJSONObject(messageJson, null);
+
+                //Making sure if messageObject actually contains html content
+                assertEquals(message.getContent().html == "", false);
+
+                JSONObject messageBackInJSON = message.toJSONObject();
+                JSONObject contentJSON = messageBackInJSON.getJSONObject("content");
+                String html = contentJSON.optString("html");
```
vbabenkoru: It might be better to do `optString("html", null)` and then `assertNull`.

iterableapi/src/main/java/com/iterable/iterableapi/IterableUtil.java
```diff
@@ -63,6 +64,15 @@ static File getSdkCacheDir(Context context) {
         return instance.getSdkCacheDir(context);
     }
 
+    static File getSDKFileDirectory(Context context) {
```
vbabenkoru: `getSDKFilesDirectory`?

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppMessage.java
```diff
@@ -280,12 +287,14 @@ static IterableInAppMessage fromJSONObject(JSONObject messageJson) {
...
         IterableInAppMessage message = new IterableInAppMessage(messageId,
                 new Content(html, padding, backgroundAlpha), customPayload, createdAt, expiresAt, trigger, saveToInbox, inboxMetadata);
+        message.inAppStorageInterface = storageInterface;
         message.processed = messageJson.optBoolean(IterableConstants.ITERABLE_IN_APP_PROCESSED, false);
         message.consumed = messageJson.optBoolean(IterableConstants.ITERABLE_IN_APP_CONSUMED, false);
         message.read = messageJson.optBoolean(IterableConstants.ITERABLE_IN_APP_READ, false);
         return message;
     }
 
+    //Called when saving the message object to file. Commented lines where we were writing html content to json to be stored in file.
```
vbabenkoru: Lines were removed, so update or remove the comment.

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppMessage.java
```diff
@@ -243,7 +246,10 @@ void setRead(boolean read) {
         onChanged();
     }
 
-    static IterableInAppMessage fromJSONObject(JSONObject messageJson) {
+
+    //This method is called when app is just launched from SYNC inApp. Also called from IterableInAppFileStorage to load json from stored file
```
vbabenkoru: There are not the only situations when it is called.

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppMessage.java
```diff
@@ -192,7 +192,10 @@ Date getExpiresAt() {
     }
 
     public Content getContent() {
-         return content;
+        if (content.html == null || content.html == "") {
```
vbabenkoru: When would it be not null and equal to `""`?

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFileStorage.java
```diff
@@ -113,13 +115,88 @@ private void load() {
...
+        }
+
+        File file = new File(folder, "index.html");
+        Boolean result = IterableUtil.writeFile(file, contentHTML);
+        if (!result) {
+            IterableLogger.e(TAG, "Failed to store HTML content");
+        }
+    }
+
+    private File createFolderIfNecessary(String messageID) {
```
vbabenkoru: This looks too generic.. `createMessageFolderIfNecessary`?

iterableapi/src/test/java/com/iterable/iterableapi/IterableInAppMessageTest.java
Ayyanchira: Wow... that's a great suggestion.. Thanks 👍

iterableapi/src/test/java/com/iterable/iterableapi/IterableInAppMessageTest.java
Ayyanchira: Even I took a while reading and understanding this 😬😅.. Changing it right away

iterableapi/src/test/java/com/iterable/iterableapi/IterableInAppMessageTest.java
Ayyanchira: Thanks... I actually wanted to do that... 😅

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppMessage.java
Ayyanchira: Removing the line. As I initially put for my understanding...

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppMessage.java
Ayyanchira: Hmm... I was not sure while storing if I should keep html as "" or not store html at all in the json. Now that we are not storing html at all.. I think, I should remove "" check

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFileStorage.java
```diff
@@ -86,7 +88,10 @@ private void loadMessagesFromJson(JSONObject jsonData) {
             for (int i = 0; i < messagesJson.length(); i++) {
                 JSONObject messageJson = messagesJson.optJSONObject(i);
                 if (messageJson != null) {
-                    IterableInAppMessage message = IterableInAppMessage.fromJSONObject(messageJson);
+                    IterableInAppMessage message = IterableInAppMessage.fromJSONObject(messageJson, this);
+                    if (message.getContent().html != null) {
```
vbabenkoru: This is going to read from a previously saved HTML every time.

Ayyanchira: Thanks for pointing this out. This would have been a silent bug if unnoticed.

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFileStorage.java
```diff
@@ -112,14 +118,96 @@ private void load() {
         }
     }
 
+    private void save(IterableInAppMessage message) {
```
vbabenkoru: The semantics of the call imply that this only saves one message, but it calls `saveMetadata` that saves metadata of _all_ messages. I think we can remove the argument here and just iterate over all messages.

vbabenkoru: Make sure you keep it thread-safe if you do iterate though.

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppMessage.java
```diff
@@ -243,7 +247,16 @@ void setRead(boolean read) {
         onChanged();
     }
 
-    static IterableInAppMessage fromJSONObject(JSONObject messageJson) {
+    public boolean isLoadedHtmlFromJson() {
+        return loadedHtmlFromJson;
+    }
+
+    public void setLoadedHtmlFromJson(boolean loadedHtmlFromJson) {
```
vbabenkoru: This should not be public API, make it package-private.

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppMessage.java
```diff
@@ -243,7 +247,16 @@ void setRead(boolean read) {
         onChanged();
     }
 
-    static IterableInAppMessage fromJSONObject(JSONObject messageJson) {
+    public boolean isLoadedHtmlFromJson() {
```
vbabenkoru: This should not be public API, make it package-private.

vbabenkoru: Also, `isLoadedHtmlFromJson` sounds a bit weird.. What do you think about `hasLoadedHtmlFromJson`?

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFileStorage.java
```diff
@@ -85,8 +89,13 @@ private void loadMessagesFromJson(JSONObject jsonData) {
         if (messagesJson != null) {
             for (int i = 0; i < messagesJson.length(); i++) {
                 JSONObject messageJson = messagesJson.optJSONObject(i);
+
                 if (messageJson != null) {
-                    IterableInAppMessage message = IterableInAppMessage.fromJSONObject(messageJson);
+                    IterableInAppMessage message = IterableInAppMessage.fromJSONObject(messageJson, this);
+                    if (message.hasLoadedHtmlFromJson()) {
```
vbabenkoru: You don't need to do this, this should already be handled when you save.

Ayyanchira: Right.. Thanks.. 👍

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFileStorage.java
```diff
@@ -112,14 +117,118 @@ private void load() {
         }
     }
 
+    private synchronized void saveMessages() {
+        saveHTMLsInBackground();
+        saveMetadata();
+    }
+
+    private void saveHTMLsInBackground() {
+        Future.runAsync(new Callable<Void>() {
```
vbabenkoru: It should be thread-safe, not explicitly running on another thread. Let's handle background save in another PR.

Ayyanchira: Removing the Future Async part...

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFileStorage.java
```diff
@@ -112,14 +114,102 @@ private void load() {
         }
     }
 
+    private synchronized void saveMessages() {
+        saveHTMLsInBackground();
+        saveMetadata();
+    }
+
+    private synchronized void saveHTMLsInBackground() {
```
vbabenkoru: The name is misleading, it's not running in background so it can be `saveHtmlContent`?

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFileStorage.java
```diff
@@ -38,20 +38,21 @@ public synchronized IterableInAppMessage getMessage(String messageId) {
     @Override
     public synchronized void addMessage(IterableInAppMessage message) {
         messages.put(message.getMessageId(), message);
+        saveMessages();
```
vbabenkoru: Is there a reason why it's running before `message.setOnChangeListener(this);` now?

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppMessage.java
```diff
@@ -243,7 +247,18 @@ void setRead(boolean read) {
         onChanged();
     }
 
-    static IterableInAppMessage fromJSONObject(JSONObject messageJson) {
+    @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+    public boolean hasLoadedHtmlFromJson() {
```
vbabenkoru: This can actually be package-private, not public + restricted to the library group, since these methods are only accessed from within the same package.

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFileStorage.java
Ayyanchira: I don't think I can give a good reason but when removing a message first thing we do is remove the listener... so thought of keeping the symmetry by having it called at last while adding a message.

Ayyanchira: Reverting...

Ayyanchira:
🏁.....🏃‍♂️........................................⫯

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFileStorage.java
```diff
@@ -15,7 +15,7 @@
 
 public class IterableInAppFileStorage implements IterableInAppStorage, IterableInAppMessage.OnChangeListener {
     private static final String TAG = "IterableInAppFileStorage";
-
+    String FOLDER_PATH = "IterableInAppFileStorage";
```
vbabenkoru: This should be `private static final String` since it's a constant.

iterableapi/src/main/java/com/iterable/iterableapi/IterableUtil.java
```diff
@@ -174,6 +200,7 @@ String readFile(File file) {
 
         boolean writeFile(File file, String content) {
             try {
+                Log.e(TAG + "Writing :",content);
```
vbabenkoru: Looks like a temporary debug line?

iterableapi/src/main/java/com/iterable/iterableapi/IterableUtil.java
```diff
@@ -5,6 +5,7 @@
 import android.content.pm.PackageInfo;
 import android.content.pm.PackageManager;
 import android.support.annotation.VisibleForTesting;
+import android.util.Log;
```
vbabenkoru: Also remove the import

Ayyanchira:
🏆🏃‍♂️🏁.............................................⫯

PR #165, opened 2019-10-30
[MOB-904] Handle Empty Thumbnail URL

Ayyanchira:
Log appropriate error when an empty url is obtained from inbox item thumbnail parameter

### DIFF
--- a/iterableapi-ui/src/main/java/com/iterable/iterableapi/ui/BitmapLoader.java
+++ b/iterableapi-ui/src/main/java/com/iterable/iterableapi/ui/BitmapLoader.java
@@ -26,6 +26,11 @@ public class BitmapLoader {
     private static final int DEFAULT_TIMEOUT_MS = 3000;
 
     static public void loadBitmap(final ImageView imageView, final Uri uri) {
+        if (uri == null || uri.getPath() == null || uri.getPath().isEmpty()) {
+            IterableLogger.d("BitmapLoader", "Empty url for Thumbnail in inbox");
+            return;
+        }
+
         Future.runAsync(new Callable<Bitmap>() {
             @Override
             public Bitmap call() throws Exception {


iterableapi-ui/src/main/java/com/iterable/iterableapi/ui/BitmapLoader.java
```diff
@@ -26,6 +26,11 @@
     private static final int DEFAULT_TIMEOUT_MS = 3000;
 
     static public void loadBitmap(final ImageView imageView, final Uri uri) {
+        if (uri.getPath() == "") {
```
vbabenkoru: Can `uri` be `null` at all?

iterableapi-ui/src/main/java/com/iterable/iterableapi/ui/BitmapLoader.java
```diff
@@ -26,6 +26,11 @@
...
+            IterableLogger.e("BitmapLoader", "Empty url for Thumbnail in inbox");
```
vbabenkoru: If there is no image, it's not an error per se, we can maybe just log it as a debug or verbose message.

iterableapi-ui/src/main/java/com/iterable/iterableapi/ui/BitmapLoader.java
Ayyanchira: Yes. If user fails to put anything in thumbnail field creating the template.

iterableapi-ui/src/main/java/com/iterable/iterableapi/ui/BitmapLoader.java
Ayyanchira: Hmm.. sounds right.

iterableapi-ui/src/main/java/com/iterable/iterableapi/ui/BitmapLoader.java
vbabenkoru: `uri == null || uri.getPath().isEmpty()` then?

iterableapi-ui/src/main/java/com/iterable/iterableapi/ui/BitmapLoader.java
```diff
@@ -26,6 +26,11 @@
     private static final int DEFAULT_TIMEOUT_MS = 3000;
 
     static public void loadBitmap(final ImageView imageView, final Uri uri) {
+        if (uri.getPath().isEmpty() || uri == null) {
```
vbabenkoru: This will crash if `uri` is null.

vbabenkoru: Actually `getPath()` can also return null, so it's better to check for `uri == null || uri.getPath() == null || uri.getPath.isEmpty()`

Ayyanchira: Should I remove == null? I checked sending no links in thumbnail and it ends up as "".

Ayyanchira: Ok.. I will have it in a sequence.. Got it. 👍

vbabenkoru: Yes, conditions are evaluated in order. If you're accessing a method on a null object before verifying that it's not null, it will crash. If you check for non-nullity before accessing properties, it won't proceed with checks if the first one fails.

Ayyanchira: Thanks for this Victor! Shame that I didn't know that `OR` operator doesn't even check for`RHS`(Right Hand Side) if the `LHS` is `true`. Somehow I used to feel that both LHS and RHS is accessed/computed regardless and the reason why we had `if let` nesting in Xcode. 😓

PR #167, opened 2019-11-07
[MOB-947] Adding test method for IterableInboxSession initializer

Ayyanchira:
Checked the red lines in codecov and experimenting if it takes the code coverage of the file to 100%

### DIFF
--- a/iterableapi/src/test/java/com/iterable/iterableapi/IterableInboxSessionTest.java
+++ b/iterableapi/src/test/java/com/iterable/iterableapi/IterableInboxSessionTest.java
@@ -1,0 +1,20 @@
+package com.iterable.iterableapi;
+
+import org.junit.Test;
+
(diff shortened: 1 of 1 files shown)


iterableapi/src/test/java/com/iterable/iterableapi/IterableInboxSessionTest.java
```diff
@@ -0,0 +1,22 @@
+package com.iterable.iterableapi;
+
+import org.junit.Test;
+
+import static org.junit.Assert.*;
+
+public class IterableInboxSessionTest {
+
```
vbabenkoru: Leave just one empty line here

Ayyanchira:
> Since it's basically a data class, it is probably more important to test the behavior of the methods that call `new IterableInboxSession()` and check if they track the session data correctly. Having this test is ok though.

PR #170, opened 2019-11-19
[MOB-965] Bug Fix inApp location not being sent properly

Ayyanchira:
Added new public method incorporating inAppLocation. Previous public methods now route to the new one with in-app as a default parameter to location.

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppDisplayer.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppDisplayer.java
@@ -17,7 +17,7 @@ class IterableInAppDisplayer {
         return IterableInAppHTMLNotification.getInstance() != null;
     }
 
-    boolean showMessage(IterableInAppMessage message, final IterableHelper.IterableUrlCallback clickCallback) {
+    boolean showMessage(IterableInAppMessage message, IterableInAppLocation location, final IterableHelper.IterableUrlCallback clickCallback) {
         Activity currentActivity = activityMonitor.getCurrentActivity();
         // Prevent double display
         if (currentActivity != null) {
@@ -27,7 +27,7 @@ class IterableInAppDisplayer {
                     clickCallback,
                     message.getContent().backgroundAlpha,
                     message.getContent().padding,
-                    true);
+                    true, location);
         }
         return false;
     }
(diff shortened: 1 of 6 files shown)


iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppDisplayer.java
```diff
@@ -18,6 +18,11 @@ boolean isShowingInApp() {
     }
 
     boolean showMessage(IterableInAppMessage message, final IterableHelper.IterableUrlCallback clickCallback) {
+        return showMessage(message, IterableInAppLocation.IN_APP, clickCallback);
+    }
+
+
+    boolean showMessage(IterableInAppMessage message, IterableInAppLocation location, final IterableHelper.IterableUrlCallback clickCallback) {
```
vbabenkoru: This class is package-private, so you can just change the method signature to include the extra argument.

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppHTMLNotification.java
```diff
@@ -119,6 +120,8 @@ public void setPadding(Rect insetPadding) {
         this.insetPadding = insetPadding;
     }
 
+    public void setLocation(IterableInAppLocation location) { this.location = location; }
```
vbabenkoru: Use the same formatting (new line after `{`) as elsewhere in the file.

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppDisplayer.java
```diff
@@ -42,10 +42,14 @@ boolean showMessage(IterableInAppMessage message, final IterableHelper.IterableU
...
     public static boolean showIterableNotificationHTML(Context context, String htmlString, String messageId, IterableHelper.IterableUrlCallback clickCallback, double backgroundAlpha, Rect padding) {
-        return showIterableNotificationHTML(context, htmlString, messageId, clickCallback, backgroundAlpha, padding, false);
+        return showIterableNotificationHTML(context, htmlString, messageId, clickCallback, backgroundAlpha, padding, false, IterableInAppLocation.IN_APP);
     }
 
     public static boolean showIterableNotificationHTML(Context context, String htmlString, String messageId, final IterableHelper.IterableUrlCallback clickCallback, double backgroundAlpha, Rect padding, boolean callbackOnCancel) {
+        return showIterableNotificationHTML(context, htmlString, messageId, clickCallback, backgroundAlpha, padding, false, IterableInAppLocation.IN_APP);
+    }
+
+    public static boolean showIterableNotificationHTML(Context context, String htmlString, String messageId, final IterableHelper.IterableUrlCallback clickCallback, double backgroundAlpha, Rect padding, boolean callbackOnCancel, IterableInAppLocation location) {
```
vbabenkoru: Hmm this was moved from another class at some point; you can make these methods non-public, and then there is no need to make another copy of the method, just change the arguments.

PR #171, opened 2019-11-21
[MOB-926] add session id to inbox calls

Ayyanchira:
This architecture is good in terms of scaling. Also enabling other developers to play with sessionId if they end up not using the IterableInboxFragment altogether.

I tried minimizing, switching back and forth the inbox fragment, things are working. But in certain combination, the sessionId was not generated which I will have to check properly.
One thing to note is that the bug where we are not sending appropriate inapp location needs to be fixed for sessionIds to be sent during the request. Currently only in-app consume and trackSession are able to send sessionIds

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
@@ -47,6 +47,7 @@ public class IterableApi {
     private String _deviceId;
 
     private IterableInAppManager inAppManager;
+    private String inboxSessionId;
 
 //---------------------------------------------------------------------------------------
 //endregion
@@ -832,8 +833,8 @@ public class IterableApi {
             requestJSON.put(IterableConstants.KEY_MESSAGE_ID, message.getMessageId());
             requestJSON.put(IterableConstants.KEY_MESSAGE_CONTEXT, getInAppMessageContext(message, location));
             requestJSON.put(IterableConstants.KEY_DEVICE_INFO, getDeviceInfoJson());
-            if (IterableInboxSession.sessionId != null) {
-                requestJSON.put(IterableConstants.KEY_INBOX_SESSION_ID,IterableInboxSession.sessionId);
+            if (location == IterableInAppLocation.INBOX) {
+                addInboxSessionID(requestJSON);
             }
             sendPostRequest(IterableConstants.ENDPOINT_TRACK_INAPP_OPEN, requestJSON);
         }
@@ -899,8 +900,8 @@ public class IterableApi {
             requestJSON.put(IterableConstants.ITERABLE_IN_APP_CLICKED_URL, clickedUrl);
             requestJSON.put(IterableConstants.KEY_MESSAGE_CONTEXT, getInAppMessageContext(message, clickLocation));
             requestJSON.put(IterableConstants.KEY_DEVICE_INFO, getDeviceInfoJson());
-            if (IterableInboxSession.sessionId != null) {
-                requestJSON.put(IterableConstants.KEY_INBOX_SESSION_ID,IterableInboxSession.sessionId);
+            if (clickLocation == IterableInAppLocation.INBOX) {
+                addInboxSessionID(requestJSON);
             }
             sendPostRequest(IterableConstants.ENDPOINT_TRACK_INAPP_CLICK, requestJSON);
         }
@@ -947,10 +948,9 @@ public class IterableApi {
             requestJSON.put(IterableConstants.ITERABLE_IN_APP_CLOSE_ACTION, closeAction.toString());
             requestJSON.put(IterableConstants.KEY_MESSAGE_CONTEXT, getInAppMessageContext(message, clickLocation));
             requestJSON.put(IterableConstants.KEY_DEVICE_INFO, getDeviceInfoJson());
-            if (IterableInboxSession.sessionId != null) {
-                requestJSON.put(IterableConstants.KEY_INBOX_SESSION_ID,IterableInboxSession.sessionId);
+            if (clickLocation == IterableInAppLocation.INBOX) {
+                addInboxSessionID(requestJSON);
             }
-
             sendPostRequest(IterableConstants.ENDPOINT_TRACK_INAPP_CLOSE, requestJSON);
         }
         catch (JSONException e) {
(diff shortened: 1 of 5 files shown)


iterableapi/src/main/java/com/iterable/iterableapi/IterableSession.java
```diff
@@ -0,0 +1,6 @@
+package com.iterable.iterableapi;
+
+public interface IterableSession {
```
vbabenkoru: This interface is not used anywhere, it is only implemented in one class and never called anywhere else.

iterableapi/src/main/java/com/iterable/iterableapi/IterableInboxSession.java
```diff
@@ -33,6 +39,9 @@ public IterableInboxSession() {
         this.endTotalMessageCount = 0;
         this.endUnreadMessageCount = 0;
         this.impressions = null;
+        if (this.sessionId == null) {
```
vbabenkoru: This is a constructor. `this.sessionId` is always null since it doesn't have a default value.

iterableapi/src/main/java/com/iterable/iterableapi/IterableInboxSession.java
```diff
@@ -15,6 +16,8 @@
     public final int endUnreadMessageCount;
     public final List<Impression> impressions;
 
+    public String sessionId;
```
vbabenkoru: Why isn't it final like the other fields?

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -835,14 +833,22 @@ public void trackInAppOpen(IterableInAppMessage message, IterableInAppLocation l
...
+                addSessionID(requestJSON);
+            }
             sendPostRequest(IterableConstants.ENDPOINT_TRACK_INAPP_OPEN, requestJSON);
         }
         catch (JSONException e) {
             e.printStackTrace();
         }
     }
 
+    void addSessionID(JSONObject requestJSON) throws JSONException{
```
vbabenkoru: `addInboxSessionID`?

iterableapi/src/main/java/com/iterable/iterableapi/IterableInboxSession.java
Ayyanchira: Oh... I had sessionId static before... forgot to remove this... Thanks

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -18,17 +18,14 @@
...
-import java.util.Date;
 import java.util.List;
-import java.util.TimeZone;
 import java.util.UUID;
 
 /**
  * Created by David Truong dt@iterable.com
  */
-public class IterableApi {
+public class IterableApi{
```
vbabenkoru: Why has this changed?

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -835,14 +833,22 @@ public void trackInAppOpen(IterableInAppMessage message, IterableInAppLocation l
...
+                addInboxSessionID(requestJSON);
+            }
             sendPostRequest(IterableConstants.ENDPOINT_TRACK_INAPP_OPEN, requestJSON);
         }
         catch (JSONException e) {
             e.printStackTrace();
         }
     }
 
+    void addInboxSessionID(JSONObject requestJSON) throws JSONException{
```
vbabenkoru: This should be private

vbabenkoru: Also, move this method to private functions at the end of the class.

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -1480,6 +1496,16 @@ private JSONObject getDeviceInfoJson() {
         return deviceInfo;
     }
 
+    @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+    public void setInboxSessionId (String inboxSessionId){
```
vbabenkoru: Please use autoformatting

iterableapi/src/test/java/com/iterable/iterableapi/IterableApiTest.java
```diff
@@ -364,6 +363,8 @@ public void testInAppOpenExtended() throws Exception {
 
         IterableApi.initialize(RuntimeEnvironment.application, "apiKey", new IterableConfig.Builder().setAutoPushRegistration(false).build());
         IterableApi.getInstance().setEmail("test@email.com");
+        IterableInboxSession inboxSession = new IterableInboxSession();
```
vbabenkoru: No need to create `IterableInboxSession` just to set the session id.. You can pass any string and then verify it.

iterableapi/src/test/java/com/iterable/iterableapi/IterableApiTest.java
```diff
@@ -509,6 +518,7 @@ public void testInboxSession() throws Exception {
         assertEquals(5, requestJson.getInt(IterableConstants.ITERABLE_INBOX_START_UNREAD_MESSAGE_COUNT));
         assertEquals(8, requestJson.getInt(IterableConstants.ITERABLE_INBOX_END_TOTAL_MESSAGE_COUNT));
         assertEquals(3, requestJson.getInt(IterableConstants.ITERABLE_INBOX_END_UNREAD_MESSAGE_COUNT));
+        assertNotNull(requestJson.getString(IterableConstants.KEY_INBOX_SESSION_ID));
```
vbabenkoru: Compare it with the value you passed to `setInboxSessionId`.

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
Ayyanchira: Seriously... no idea.. Why would I poke this thing 🤯

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
Ayyanchira: No comments 😅... Its installed. and I also use option command L frequently... Oh yes! Now I remember... When I did that.. It literally had thousands of lines changed. so I reverted back and tried to manually find spaces... But yes... Sorry for this again 😬

iterableapi/src/test/java/com/iterable/iterableapi/IterableApiTest.java
Ayyanchira: Wonderful!😄 How did I not think of it!

iterableapi/src/test/java/com/iterable/iterableapi/IterableApiTest.java
Ayyanchira: For other methods, yes. But at this line, I will assert equal with generated sessionID as InboxSession is internally called on the top.

PR #174, opened 2019-12-06
[MOB-914] Read Write in-app json on persistent file directory

Ayyanchira:
The in app load will now look for file directory. If it does not exist (which will occur during the first launch after new SDK migration), the inapp json will be loaded from cache directory and subsequent metadata changes will be saved in file directory.

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFileStorage.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFileStorage.java
@@ -99,6 +99,10 @@ public class IterableInAppFileStorage implements IterableInAppStorage, IterableI
     }
 
     private File getInAppStorageFile() {
+        return new File(getInAppContentFolder(), "itbl_inapp.json");
+    }
+
+    private File getInAppCacheStorageFile() {
         return new File(IterableUtil.getSdkCacheDir(context), "itbl_inapp.json");
     }
 
@@ -107,6 +111,9 @@ public class IterableInAppFileStorage implements IterableInAppStorage, IterableI
             File inAppStorageFile = getInAppStorageFile();
             if (inAppStorageFile.exists()) {
                 JSONObject jsonData = new JSONObject(IterableUtil.readFile(inAppStorageFile));
+                loadMessagesFromJson(jsonData);
+            } else if (getInAppCacheStorageFile().exists()) {
+                JSONObject jsonData = new JSONObject(IterableUtil.readFile(getInAppCacheStorageFile()));
                 loadMessagesFromJson(jsonData);
             }
         } catch (Exception e) {
--- a/iterableapi/src/test/java/com/iterable/iterableapi/IterableInAppFileStorageTest.java
+++ b/iterableapi/src/test/java/com/iterable/iterableapi/IterableInAppFileStorageTest.java
@@ -94,4 +94,40 @@ public class IterableInAppFileStorageTest {
         assertEquals(message.getContent().html,storage.getHTML(messageID1));
         assertNotNull(storage.getHTML(messageID2));
     }
+
+    @Test
+    public void loadMessagesWithNoJson() {
+        IterableInAppFileStorage storage = new IterableInAppFileStorage(RuntimeEnvironment.application);
+        assertEquals(0, storage.getMessages().size());
+    }
+
+    @Test
+    public void loadMessagesWithJsonInCache() throws Exception {
+        File folder = IterableUtil.getSdkCacheDir(RuntimeEnvironment.application);
+        File oldJsonStorageFile = new File(folder, "itbl_inapp.json");
+        Boolean fileWriteResult = IterableUtil.writeFile(oldJsonStorageFile, IterableTestUtils.getResourceString("inapp_payload_multiple.json"));
+        assertTrue(fileWriteResult);
+        IterableInAppFileStorage storage = new IterableInAppFileStorage(RuntimeEnvironment.application);
+        assertTrue(storage.getMessages().size() > 1);
+
+        // Simulate message update and check if new json file is created in SDK directory
+        storage.onInAppMessageChanged(storage.getMessages().get(0));
+        File sdkFilesDirectory = IterableUtil.getSDKFilesDirectory(RuntimeEnvironment.application);
+        File inAppDirectory = IterableUtil.getDirectory(sdkFilesDirectory, "IterableInAppFileStorage");
+        File inAppJsonFile = new File(inAppDirectory, "itbl_inapp.json");
+        assertTrue(inAppJsonFile.exists());
+    }
+
+    @Test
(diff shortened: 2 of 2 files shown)


iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFileStorage.java
```diff
@@ -99,16 +99,23 @@ private void loadMessagesFromJson(JSONObject jsonData) {
...
     private void load() {
         try {
             File inAppStorageFile = getInAppStorageFile();
+            JSONObject jsonData;
             if (inAppStorageFile.exists()) {
-                JSONObject jsonData = new JSONObject(IterableUtil.readFile(inAppStorageFile));
-                loadMessagesFromJson(jsonData);
+                jsonData = new JSONObject(IterableUtil.readFile(inAppStorageFile));
+            } else {
+                jsonData = new JSONObject(IterableUtil.readFile(getInAppCacheStorageFile()));
```
vbabenkoru: This will now result in an error on clean installs, since none of the files exist. It should not be an error since it's a perfectly normal state.

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFileStorage.java
```diff
@@ -108,6 +112,9 @@ private void load() {
             if (inAppStorageFile.exists()) {
                 JSONObject jsonData = new JSONObject(IterableUtil.readFile(inAppStorageFile));
                 loadMessagesFromJson(jsonData);
+            } else {
+                JSONObject jsonData = new JSONObject(IterableUtil.readFile(getInAppCacheStorageFile()));
+                loadMessagesFromJson(jsonData);
```
vbabenkoru: This will still throw an exception.

Ayyanchira: In that case, it will just log error statements and continue fetching data from server and writing new json file.

vbabenkoru: But it is not an error, this is a completely normal state. We should only log errors if something unexpected happens.

iterableapi/src/test/java/com/iterable/iterableapi/IterableInAppFileStorageTest.java
```diff
@@ -94,4 +94,32 @@ public void testHTMLForMigration() throws Exception {
...
+        File oldJsonStorageFile = new File(folder, "itbl_inapp.json");
+        assertTrue(!oldJsonStorageFile.exists());
+
+        IterableInAppFileStorage storage = new IterableInAppFileStorage(RuntimeEnvironment.application);
+
+        // Case 1: Json doesn't exist in cache or file directory.
+        assertEquals(0, storage.getMessages().size());
+
+        // Case 2: Json exists in cache directory only
+        Boolean fileWriteOperation = IterableUtil.writeFile(oldJsonStorageFile, IterableTestUtils.getResourceString("inapp_payload_multiple.json"));
```
vbabenkoru: `fileWriteOperation` is not an operation, it is a result

iterableapi/src/test/java/com/iterable/iterableapi/IterableInAppFileStorageTest.java
```diff
@@ -94,4 +94,32 @@ public void testHTMLForMigration() throws Exception {
         assertEquals(message.getContent().html,storage.getHTML(messageID1));
         assertNotNull(storage.getHTML(messageID2));
     }
+
+    @Test
+    public void testOldMetadataMigration() throws Exception {
+        File folder = IterableUtil.getSdkCacheDir(RuntimeEnvironment.application);
+        assertTrue(folder.isDirectory());
+        File oldJsonStorageFile = new File(folder, "itbl_inapp.json");
+        assertTrue(!oldJsonStorageFile.exists());
```
vbabenkoru: I think there should be no need to test this since it's basically testing the test environment.

iterableapi/src/test/java/com/iterable/iterableapi/IterableInAppFileStorageTest.java
```diff
@@ -94,4 +94,32 @@ public void testHTMLForMigration() throws Exception {
...
+        // Case 1: Json doesn't exist in cache or file directory.
+        assertEquals(0, storage.getMessages().size());
+
+        // Case 2: Json exists in cache directory only
+        Boolean fileWriteOperation = IterableUtil.writeFile(oldJsonStorageFile, IterableTestUtils.getResourceString("inapp_payload_multiple.json"));
+        assertTrue(fileWriteOperation);
+        storage = new IterableInAppFileStorage(RuntimeEnvironment.application);
+        assertTrue(storage.getMessages().size() > 1);
+
+        // Case 3: Json also exists in file folder.
```
vbabenkoru: Can you add another validation to check that if a file at the old location exists, `IterableInAppFileStorage` saves it to a new location?

iterableapi/src/test/java/com/iterable/iterableapi/IterableInAppFileStorageTest.java
```diff
@@ -94,4 +94,34 @@ public void testHTMLForMigration() throws Exception {
         assertEquals(message.getContent().html,storage.getHTML(messageID1));
         assertNotNull(storage.getHTML(messageID2));
     }
+
+    @Test
+    public void testOldMetadataMigration() throws Exception {
```
vbabenkoru: Since this test has multiple scenarios that are better executed separately, can you split each scenario into a separate test with its own name?

iterableapi/src/test/java/com/iterable/iterableapi/IterableInAppFileStorageTest.java
```diff
@@ -94,4 +94,34 @@ public void testHTMLForMigration() throws Exception {
...
+        assertTrue(fileWriteResult);
+        storage = new IterableInAppFileStorage(RuntimeEnvironment.application);
+        assertTrue(storage.getMessages().size() > 1);
+
+        // Simulate message update and check if new json file is created in SDK directory
+        storage.onInAppMessageChanged(storage.getMessages().get(0));
+        File sdkFilesDirectory = IterableUtil.getSDKFilesDirectory(RuntimeEnvironment.application);
+        File inAppDirectory = IterableUtil.getDirectory(sdkFilesDirectory, "IterableInAppFileStorage");
+        File inAppJsonFile = new File(inAppDirectory, "itbl_inapp.json");
+        assertTrue(inAppDirectory.exists());
```
vbabenkoru: Also check that inAppJsonFile exists

Ayyanchira: Nice catch...! It was actually supposed to be inAppJsonFile.

PR #178, opened 2019-12-17
Make In App Save Asynchronous

Ayyanchira:
Implementing fileoperation queue with delay of 0.1 sec to perform the operation and avoid repeated save calls

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFileStorage.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFileStorage.java
@@ -1,17 +1,22 @@
 package com.iterable.iterableapi;
 
 import android.content.Context;
+import android.os.Handler;
+import android.os.Looper;
 
 import org.json.JSONArray;
 import org.json.JSONException;
 import org.json.JSONObject;
 
 import java.io.File;
+import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
 
 public class IterableInAppFileStorage implements IterableInAppStorage, IterableInAppMessage.OnChangeListener {
     private static final String TAG = "IterableInAppFileStorage";
@@ -19,6 +24,12 @@ public class IterableInAppFileStorage implements IterableInAppStorage, IterableI
     private final Context context;
     private Map<String, IterableInAppMessage> messages =
             Collections.synchronizedMap(new LinkedHashMap<String, IterableInAppMessage>());
+    private ArrayDeque<FileOperation> fileOperationQueue = new ArrayDeque<>();
+    private ExecutorService singleThreadExecutor = Executors.newSingleThreadExecutor();
+
+    enum Operation {
+        SAVE, LOAD, OTHER
+    }
 
     IterableInAppFileStorage(Context context) {
         this.context = context;
@@ -39,7 +50,7 @@ public class IterableInAppFileStorage implements IterableInAppStorage, IterableI
     public synchronized void addMessage(IterableInAppMessage message) {
         messages.put(message.getMessageId(), message);
         message.setOnChangeListener(this);
-        saveMessages();
+        saveMessagesInBackground();
     }
 
     @Override
@@ -47,12 +58,12 @@ public class IterableInAppFileStorage implements IterableInAppStorage, IterableI
         message.setOnChangeListener(null);
         removeHTML(message.getMessageId());
         messages.remove(message.getMessageId());
-        saveMessages();
+        saveMessagesInBackground();
     }
 
     @Override
     public void onInAppMessageChanged(IterableInAppMessage message) {
-        saveMessages();
+        saveMessagesInBackground();
     }
 
     private synchronized void clearMessages() {
@@ -121,6 +132,27 @@ public class IterableInAppFileStorage implements IterableInAppStorage, IterableI
         }
     }
 
+    private void saveMessagesInBackground() {
+        //Add SaveRequest to Queue
+        FileOperation operation = null;
+        try {
+            operation = fileOperationQueue.getLast();
+        } catch (RuntimeException e) {
+            operation = null;
+        }
+        if (operation == null || operation.operationType != Operation.SAVE) {
+            FileOperation save = new FileOperation(Operation.SAVE, new Runnable() {
+                @Override
+                public void run() {
+                    saveMessages();
+                }
+            });
+            fileOperationQueue.add(save);
+        }
+        executeOperationQueueTasks();
+    }
+
+
     private synchronized void saveMessages() {
         saveHTMLContent();
         saveMetadata();
@@ -134,7 +166,6 @@ public class IterableInAppFileStorage implements IterableInAppStorage, IterableI
             }
         }
     }
-
 
     private synchronized void saveMetadata() {
         try {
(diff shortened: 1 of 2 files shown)


iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFileStorage.java
```diff
@@ -1,24 +1,35 @@
...
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
 
 public class IterableInAppFileStorage implements IterableInAppStorage, IterableInAppMessage.OnChangeListener {
     private static final String TAG = "IterableInAppFileStorage";
     private static final String FOLDER_PATH = "IterableInAppFileStorage";
     private final Context context;
     private Map<String, IterableInAppMessage> messages =
             Collections.synchronizedMap(new LinkedHashMap<String, IterableInAppMessage>());
+    private ArrayDeque<FileOperation> fileOperationQueue = new ArrayDeque<>();
```
vbabenkoru: If this is shared across threads, every operation with it must be atomic/synchronized.

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFileStorage.java
```diff
@@ -1,24 +1,35 @@
...
+    private ExecutorService singleThreadExecutor = Executors.newSingleThreadExecutor();
+
+    enum Operation {
+        SAVE, LOAD, OTHER
```
vbabenkoru: Will `OTHER` ever do anything?

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFileStorage.java
```diff
@@ -121,6 +132,27 @@ private void load() {
...
+    private void saveMessagesInBackground() {
+        //Add SaveRequest to Queue
+        FileOperation operation = null;
+        try {
+            operation = fileOperationQueue.getLast();
+        } catch (RuntimeException e) {
+            operation = null;
+        }
+        if (operation == null || operation.operationType != Operation.SAVE) {
+            FileOperation save = new FileOperation(Operation.SAVE, new Runnable() {
```
vbabenkoru: It's posting both message type _and_ code to run.. I think it should be one or the other - either post a message and let it execute its own code, or post code, like `Handler` does.

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFileStorage.java
```diff
@@ -121,6 +132,27 @@ private void load() {
...
+                @Override
+                public void run() {
+                    saveMessages();
+                }
+            });
+            fileOperationQueue.add(save);
+        }
+        executeOperationQueueTasks();
```
vbabenkoru: Ideally this should not be required to call this at the end. Otherwise it makes it fragile: if you forget to call this, nothing will happen.

iterableapi/src/test/java/com/iterable/iterableapi/IterableInAppFileStorageTest.java
```diff
@@ -130,4 +132,10 @@ public void loadMessagesWithJsonInFilesDirectory() throws Exception {
         assertEquals(1, storage.getMessages().size());
     }
 
+    private void flushThreads() throws InterruptedException {
```
vbabenkoru: We need to remove this somehow.. Maybe inject an executor in tests?

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFileStorage.java
```diff
@@ -121,6 +132,27 @@ private void load() {
         }
     }
 
+    private void saveMessagesInBackground() {
+        //Add SaveRequest to Queue
+        FileOperation operation = null;
+        try {
+            operation = fileOperationQueue.getLast();
+        } catch (RuntimeException e) {
```
vbabenkoru: Exceptions should not be part of a normal code path.. Is there any way to check it without throwing an exception?

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFileStorage.java
Ayyanchira: May be operations like clearing the cache folder. 🤷‍♂️

vbabenkoru: It'll probably be called `CLEAR_CACHE` then.. You probably will never use `OTHER` value itself.

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFileStorage.java
```diff
@@ -133,23 +130,18 @@ private void load() {
     }
 
     private void saveMessagesInBackground() {
-        //Add SaveRequest to Queue
-        FileOperation operation = null;
-        try {
-            operation = fileOperationQueue.getLast();
-        } catch (RuntimeException e) {
-            operation = null;
+        if (fileOperationThread.getLooper() == null) {
```
vbabenkoru: Since we're probably going to save anyway, I think it might be better to move this thread initialization code to the constructor.

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFileStorage.java
```diff
@@ -133,23 +130,18 @@ private void load() {
...
-            });
-            fileOperationQueue.add(save);
+        if (customHandler == null) {
+            customHandler = new CustomHandler(fileOperationThread.getLooper());
+        }
+
+        if (customHandler != null && !customHandler.isProcessing) {
+            Message msg = customHandler.obtainMessage();
+            msg.what = OPERATION_SAVE;
+            customHandler.sendMessage(msg);
```
vbabenkoru: I think you can replace this with just `customHandler.sendMessage(OPERATION_SAVE);`

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFileStorage.java
```diff
@@ -262,13 +243,25 @@ public void removeHTML(String messageID) {
...
+            super(threadLooper);
+        }
 
-        public FileOperation(Operation operationType, Runnable backgroundMethod) {
-            this.operationType = operationType;
-            this.backgroundMethod = backgroundMethod;
+        @Override
+        public void handleMessage(Message msg) {
+            if (msg.what == OPERATION_SAVE) {
+                customHandler.postDelayed(new Runnable() {
```
vbabenkoru: Post the message with a delay, no need to post a runnable within the handler.

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFileStorage.java
```diff
@@ -262,13 +243,25 @@ public void removeHTML(String messageID) {
...
+                    @Override
+                    public void run() {
+                        saveMessages();
+                        isProcessing = false;
```
vbabenkoru: You also probably won't need `isProcessing` if you post a message with a delay and use `hasMessage` to check if there is one in the queue before posting.

Ayyanchira: I thought so too.. implementation with hasMessage did not go well. It returned false even if the the process was delayed and not yet executed. 😕

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFileStorage.java
Ayyanchira: Handler was not receiving the callback unless I used obtainMessage. I'll check it again.

PR #181, opened 2019-12-23
[MOB 1026] Default sort by date is now in descending order

Ayyanchira:
Sort by date applied after filter when getting inbox messages.
Added date fields to test method json files.
Made necessary changes to test method to test ordering of inbox messages.

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppMessage.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppMessage.java
@@ -10,7 +10,7 @@ import org.json.JSONObject;
 
 import java.util.Date;
 
-public class IterableInAppMessage {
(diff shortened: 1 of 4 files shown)


iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppMessage.java
```diff
@@ -440,4 +440,12 @@ public static JSONObject getNextMessageFromPayload(String payload) {
         }
         return returnObject;
     }
+
+    @Override
+    public int compareTo(IterableInAppMessage message) {
```
vbabenkoru: This sets the sorting logic order for all in-app messages. It makes more sense to define the comparator in `IterableInAppManager`, since it's only used for inbox messages and will be customizable in the future.

PR #182, opened 2019-12-26
[MOB-1006] Add support for 'read' flag from getMessages API

Ayyanchira:
Added inbox parameter to constructor. Default read status will be false unless specified in json obtained from server.
One message in json file used for testing updated to have read status as true.
One Test method updated and one test method added to verify the functionality.

### DIFF
--- a/iterableapi/src/test/java/com/iterable/iterableapi/IterableInboxTest.java
+++ b/iterableapi/src/test/java/com/iterable/iterableapi/IterableInboxTest.java
@@ -92,17 +92,27 @@ public class IterableInboxTest extends BaseTest {
         IterableInAppManager inAppManager = IterableApi.getInstance().getInAppManager();
         inAppManager.syncInApp();
         List<IterableInAppMessage> inboxMessages = inAppManager.getInboxMessages();
-        assertEquals(2, inAppManager.getUnreadInboxMessagesCount());
+        assertEquals(1, inAppManager.getUnreadInboxMessagesCount());
         assertEquals(2, inboxMessages.size());
         assertFalse(inboxMessages.get(0).isRead());
-        assertFalse(inboxMessages.get(1).isRead());
-
-        inAppManager.setRead(inboxMessages.get(1), true);
-
-        assertEquals(1, inAppManager.getUnreadInboxMessagesCount());
-        assertEquals(2, inAppManager.getInboxMessages().size());
-        assertFalse(inboxMessages.get(0).isRead());
         assertTrue(inboxMessages.get(1).isRead());
+        inAppManager.setRead(inboxMessages.get(0), true);
+        assertEquals(0, inAppManager.getUnreadInboxMessagesCount());
+        assertEquals(2, inAppManager.getInboxMessages().size());
+        assertTrue(inboxMessages.get(0).isRead());
+        assertTrue(inboxMessages.get(1).isRead());
+    }
+
+    @Test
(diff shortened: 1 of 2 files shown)


iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppMessage.java
```diff
@@ -289,18 +290,18 @@ static IterableInAppMessage fromJSONObject(JSONObject messageJson, IterableInApp
         }
 
         Boolean saveToInbox = messageJson.has(IterableConstants.ITERABLE_IN_APP_SAVE_TO_INBOX) ? messageJson.optBoolean(IterableConstants.ITERABLE_IN_APP_SAVE_TO_INBOX) : null;
+        Boolean read = messageJson.has(IterableConstants.ITERABLE_IN_APP_READ) ? messageJson.optBoolean(IterableConstants.ITERABLE_IN_APP_READ) : false;
```
vbabenkoru: The code in this file is identical to the original in behavior. If I revert the changes in this file, the test still passes. Looks like it was supporting reading `read` from the server already.

iterableapi/src/test/java/com/iterable/iterableapi/IterableInboxTest.java
```diff
@@ -92,17 +92,27 @@ public void testSetRead() throws Exception {
...
+        assertTrue(inboxMessages.get(1).isRead());
+        inAppManager.setRead(inboxMessages.get(0), true);
+        assertEquals(0, inAppManager.getUnreadInboxMessagesCount());
+        assertEquals(2, inAppManager.getInboxMessages().size());
+        assertTrue(inboxMessages.get(0).isRead());
+        assertTrue(inboxMessages.get(1).isRead());
+    }
 
+    @Test
+    public void testMessageReadStatus() throws Exception {
```
vbabenkoru: I think this should indicate that it's testing read status from the server, right now it's not clear what the difference between the two test methods is.

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppMessage.java
Ayyanchira: Yep, we had all the provision already. I realized when I went to create constants! 😅
Thought of having the value assigned in constructor though. I didn't realize however that the test would pass without this change too!

iterableapi/src/test/java/com/iterable/iterableapi/IterableInboxTest.java
Ayyanchira: I felt this repeated too. But because the other method was more focused on `setRead` method, I thought of having it seperate. We can remove it altogether and just verify read status in already existing test method.

vbabenkoru: It's a good idea to have 2 separate tests for this, they just need to have names to tell what the difference between them is.

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppMessage.java
vbabenkoru: This change basically moved `read` from assignment to a constructor parameter, but the end result is the same.

Ayyanchira: Ok. I will revert this one.

PR #184, opened 2020-01-24
[MOB-1003] Inbox Customization

vbabenkoru:


iterableapi-ui/src/main/java/com/iterable/iterableapi/ui/inbox/IterableInboxAdapter.java
```diff
@@ -178,8 +193,16 @@ public int hashCode() {
     private List<InboxRow> inboxRowListFromInboxMessages(List<IterableInAppMessage> messages) {
         ArrayList<InboxRow> inboxRows = new ArrayList<>(messages.size());
         for (IterableInAppMessage message : messages) {
-            inboxRows.add(new InboxRow(message));
+            if (filter.filter(message)) {
```
Ayyanchira: filter.filter is very confusing.

iterableapi-ui/src/main/java/com/iterable/iterableapi/ui/inbox/IterableInboxAdapter.java
```diff
@@ -178,8 +193,16 @@ public int hashCode() {
...
+                inboxRows.add(new InboxRow(message));
+            }
         }
+        Collections.sort(inboxRows, new Comparator<InboxRow>() {
+            @Override
+            public int compare(InboxRow o1, InboxRow o2) {
+                return comparator.compare(o1.message, o2.message);
```
Ayyanchira: Just a thought : A filter applies per message while a comparator is something which needs to be set only once for entire set of messages. 
How would it look like if we then pass the comparator to IterableInboxFragment and not have interface approach there? may be like this.setComparator(comparator)?

Ayyanchira:
Overall, a good working prototype 👍
Would like to hear from @tapashmajumder too. Especially if it involved similar architectural changes or was it very different for iOS...

iterableapi-ui/src/main/java/com/iterable/iterableapi/ui/inbox/IterableInboxFilter.java
```diff
@@ -0,0 +1,9 @@
+package com.iterable.iterableapi.ui.inbox;
+
+import android.support.annotation.NonNull;
+
+import com.iterable.iterableapi.IterableInAppMessage;
+
+public interface IterableInboxFilter {
```
tapashmajumder: Does Android support `@FunctionalInterface`?

iterableapi-ui/src/main/java/com/iterable/iterableapi/ui/inbox/IterableInboxAdapter.java
```diff
@@ -178,8 +193,16 @@ public int hashCode() {
     private List<InboxRow> inboxRowListFromInboxMessages(List<IterableInAppMessage> messages) {
         ArrayList<InboxRow> inboxRows = new ArrayList<>(messages.size());
         for (IterableInAppMessage message : messages) {
-            inboxRows.add(new InboxRow(message));
+            if (filter.filter(message)) {
+                inboxRows.add(new InboxRow(message));
+            }
         }
+        Collections.sort(inboxRows, new Comparator<InboxRow>() {
```
tapashmajumder: Just curious. Not a comment. This seems to be too verbose. Does Android support lambda expressions.

Ayyanchira: It does support lambda expressions if using java > 8.0 or Kotlin.

vbabenkoru: It does or it does not, depending on the version of the build tools. Android runtime is still considered Java 6, so you have to use a desugaring compiler to transform Java 8 lambdas into Java 6 constructs. It'd be an easy 'yes' if this were a mobile app, but for a library I'm not exactly sure if we can use lambdas safely.

iterableapi-ui/src/main/java/com/iterable/iterableapi/ui/inbox/IterableInboxFilter.java
vbabenkoru: Beginning with API 24, and our minimum API level is 15...

iterableapi-ui/src/main/java/com/iterable/iterableapi/ui/inbox/IterableInboxAdapter.java
```diff
@@ -19,21 +19,28 @@
...
 
 public class IterableInboxAdapter extends RecyclerView.Adapter<IterableInboxAdapter.ViewHolder> {
 
     private static final String TAG = "IterableInboxAdapter";
-    private final @LayoutRes int itemLayoutId;
+
+    private final @NonNull OnListInteractionListener listener;
+    private final @NonNull IterableInboxAdapterExtension extension;
+    private final @NonNull IterableInboxComparator comparator;
+    private final @NonNull IterableInboxFilter filter;
```
Ayyanchira: Not sure but will there be scenario where adaptor values need to be changed without destroying it?

iterableapi-ui/src/main/java/com/iterable/iterableapi/ui/inbox/IterableInboxAdapter.java
```diff
@@ -44,28 +51,50 @@ public void onClick(View v) {
...
+
+        if (holder.unreadIndicator != null) {
+            if (inboxRow.isRead) {
+                holder.unreadIndicator.setVisibility(View.INVISIBLE);
+            } else {
+                holder.unreadIndicator.setVisibility(View.VISIBLE);
+            }
+        }
+
+        if (holder.date != null) {
```
Ayyanchira: Now I second on moving to Kotlin 😁

iterableapi-ui/src/main/java/com/iterable/iterableapi/ui/inbox/IterableInboxAdapter.java
```diff
@@ -44,28 +51,50 @@ public void onClick(View v) {
...
+            holder.date.setText(formatDate(inboxRow.createdAt));
+        }
```
Ayyanchira: Now I second on moving to Kotlin 😁

vbabenkoru: haha

iterableapi-ui/src/main/java/com/iterable/iterableapi/ui/inbox/IterableInboxAdapter.java
vbabenkoru: There shouldn't be. Since the Fragment can be destroyed and created again with just arguments, we should limit what can be modified on the Adapter or the Fragment I think..

iterableapi-ui/src/main/java/com/iterable/iterableapi/ui/inbox/IterableInboxFragment.java
```diff
@@ -148,35 +193,65 @@ public void onListItemTapped(IterableInAppMessage message) {
...
 
-    public void setInboxMode(InboxMode inboxMode) {
-        this.inboxMode = inboxMode;
+    /**
+     * Default implementation of the comparator: descending by creation date
+     */
+    private static class DefaultInboxComparator implements IterableInboxComparator {
+        @Override
+        public int compare(@NonNull IterableInAppMessage message1, @NonNull IterableInAppMessage message2) {
+            return -message1.getCreatedAt().compareTo(message2.getCreatedAt());
```
Ayyanchira: message2 compareTo message1 can help us get rid of the minus(-) prefix

vbabenkoru: This is intentional. I never remember whether comparator should compare o1 with o2 or vice versa, but the negative sign informs that we’re reversing the order.

iterableapi-ui/src/main/java/com/iterable/iterableapi/ui/inbox/IterableInboxAdapter.java
vbabenkoru: I’m actually thinking about making these setters protected so they can only be called from a subclass..

Ayyanchira: That sounds good 👍

PR #185, opened 2020-01-27
Make DialogFragment notification fullscreen

Ayyanchira:
Set Android's No Action Bar Style theme to notification.


iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFragmentHTMLNotification.java
```diff
@@ -84,7 +89,23 @@ public void onCreate(@Nullable Bundle savedInstanceState) {
         if (args != null) {
             htmlString = args.getString("html", null);
         }
-        setStyle(DialogFragment.STYLE_NO_FRAME, 0);
+        viewModel = ViewModelProviders.of(this).get(InAppNotificationViewModel.class);
+        loadAndRetrieveFromViewModel();
+    }
+
+    private void loadAndRetrieveFromViewModel() {
+        //Checking transient variables which go null after configuration changes.
```
Ayyanchira: Definitely not a right way to determining if configuration changes has occurred. I am looking for it.

iterableapi/src/main/java/com/iterable/iterableapi/model/InAppNotificationViewModel.java
```diff
@@ -0,0 +1,13 @@
...
+import android.arch.lifecycle.ViewModel;
+
+import com.iterable.iterableapi.IterableHelper;
+import com.iterable.iterableapi.IterableInAppLocation;
+
+public class InAppNotificationViewModel extends ViewModel {
+    public String htmlString;
+    public String messageID;
+    public IterableHelper.IterableUrlCallback clickCallback;
+    public IterableInAppLocation location;
```
Ayyanchira: May be we can add backgroundAlpha and padding here too. Not sure as they, along with htmlString and messageID stay intact even without model class.
Either we can store all variables in model class making it uniform or just store the transient data which tend to get destroyed on configuration changes

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFragmentHTMLNotification.java
```diff
@@ -84,7 +89,23 @@ public void onCreate(@Nullable Bundle savedInstanceState) {
...
+        if (clickCallback == null) {
+            htmlString = viewModel.htmlString;
+            messageId = viewModel.messageID;
+            clickCallback = viewModel.clickCallback;
+            location = viewModel.location;
+        } else {
+            viewModel.htmlString = htmlString;
+            viewModel.messageID = messageId;
+            viewModel.clickCallback = clickCallback;
+            viewModel.location = location;
```
Ayyanchira: Definitely not a right way to determining if configuration changes has occurred. I am looking for it.

iterableapi/build.gradle
```diff
@@ -56,6 +56,7 @@ dependencies {
     androidTestImplementation 'androidx.test.uiautomator:uiautomator:2.2.0'
     androidTestImplementation 'com.squareup.okhttp3:mockwebserver:4.2.2'
     androidTestImplementation 'org.mockito:mockito-android:3.0.0'
+    implementation 'android.arch.lifecycle:extensions:1.1.1'
```
vbabenkoru: This adds an extra dependency apps would have to pull when using our SDK.. As a library, we should limit our dependencies unless absolutely necessary.

Ayyanchira:
Resolved by adding the change to Branch [MOB-961]

PR #190, opened 2020-02-04
[MOB-1157] - Retain data during configuration change

Ayyanchira:
Globally saves urlCallBack and location.
Retains if the view is going to reload. If not, it will assign the stored variable to null.

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppDisplayer.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppDisplayer.java
@@ -95,13 +95,10 @@ class IterableInAppDisplayer {
                     return false;
                 }
 
-                IterableInAppFragmentHTMLNotification notification = IterableInAppFragmentHTMLNotification.createInstance(context, htmlString, callbackOnCancel);
+                IterableInAppFragmentHTMLNotification notification = IterableInAppFragmentHTMLNotification.createInstance(context, htmlString, callbackOnCancel, clickCallback, location);
                 notification.setTrackParams(messageId);
-                notification.setCallback(clickCallback);
                 notification.setBackgroundAlpha(backgroundAlpha);
                 notification.setPadding(padding);
-                notification.setLocation(location);
-
                 notification.show(currentActivity.getSupportFragmentManager(), "iterable_in_app");
                 return true;
             }
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFragmentHTMLNotification.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFragmentHTMLNotification.java
@@ -43,9 +43,8 @@ public class IterableInAppFragmentHTMLNotification extends DialogFragment implem
     String messageId;
     double backgroundAlpha;
     Rect insetPadding;
-    IterableHelper.IterableUrlCallback clickCallback;
-    IterableInAppLocation location;
-
+    static IterableHelper.IterableUrlCallback clickCallback;
+    static IterableInAppLocation location;
     private boolean callbackOnCancel = false;
 
     /**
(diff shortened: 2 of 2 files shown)


iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFragmentHTMLNotification.java
```diff
@@ -90,6 +91,25 @@ public void onCreate(@Nullable Bundle savedInstanceState) {
...
+        isReloading = false;
+        setCallback(IterableInAppDisplayer.urlCallBack);
+        setLocation(IterableInAppDisplayer.location);
+    }
+
+    @Override
+    public void onSaveInstanceState(Bundle outState) {
+        super.onSaveInstanceState(outState);
+        isReloading = true;
+        outState.putBoolean("isReloading", true);
```
vbabenkoru: `Fragment` itself does this
```
        FragmentActivity activity = this.getActivity();
        boolean isChangingConfigurations = activity != null && activity.isChangingConfigurations();
        if (this.mViewModelStore != null && !isChangingConfigurations) {
            this.mViewModelStore.clear();
        }
```
Looks like we can use a similar check for whether it's being destroyed because of a configuration change.

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppDisplayer.java
```diff
@@ -101,7 +105,16 @@ static boolean showIterableFragmentNotificationHTML(Context context, String html
                 notification.setBackgroundAlpha(backgroundAlpha);
                 notification.setPadding(padding);
                 notification.setLocation(location);
-
+                
+//                TODO: Callback on cancel for DialogFragment needs to be implemented
```
vbabenkoru: This is not necessary, `callbackOnCancel` is already handled in `IterableInAppFragmentHTMLNotification`.

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppDisplayer.java
```diff
@@ -87,6 +89,8 @@ public void onCancel(DialogInterface dialog) {
      * @param padding
      */
     static boolean showIterableFragmentNotificationHTML(Context context, String htmlString, String messageId, final IterableHelper.IterableUrlCallback clickCallback, double backgroundAlpha, Rect padding, boolean callbackOnCancel, IterableInAppLocation location) {
+        IterableInAppDisplayer.urlCallBack = clickCallback;
```
vbabenkoru: This assigns the static variables before the checks that follow. It is possible that the callback will be assigned, but no in-app message will get spawned. I think it might be better to move these variables to `IterableInAppFragmentHTMLNotification` class instead and pass them through `createInstance`...

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppDisplayer.java
```diff
@@ -22,7 +24,7 @@ boolean showMessage(IterableInAppMessage message, IterableInAppLocation location
         Activity currentActivity = activityMonitor.getCurrentActivity();
         // Prevent double display
         if (currentActivity != null) {
-            return IterableInAppDisplayer.showIterableNotificationHTML(currentActivity,
+            return IterableInAppDisplayer.showIterableFragmentNotificationHTML(currentActivity,
```
vbabenkoru: `isShowingInApp` is still referencing the old class, so this line may not be safe to switch without changing it.

vbabenkoru: Might also be more correct to move `notification = null;` to `onDestroy` (and only reset it if the dialog isn't being recreated because of a configuration change).

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppDisplayer.java
Ayyanchira: This should have been the very first approach. 👍

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppDisplayer.java
```diff
@@ -15,14 +15,14 @@
...
-        return IterableInAppHTMLNotification.getInstance() != null;
+        return IterableInAppFragmentHTMLNotification.getInstance() != null;
     }
 
     boolean showMessage(IterableInAppMessage message, IterableInAppLocation location, final IterableHelper.IterableUrlCallback clickCallback) {
         Activity currentActivity = activityMonitor.getCurrentActivity();
         // Prevent double display
         if (currentActivity != null) {
-            return IterableInAppDisplayer.showIterableNotificationHTML(currentActivity,
+            return IterableInAppDisplayer.showIterableFragmentNotificationHTML(currentActivity,
```
vbabenkoru: I think we should only switch to `IterableInAppFragmentHTMLNotification` when all known issues are resolved. Right now it doesn't retain some properties on rotation and uses a default DialogFragment animation.

PR #194, opened 2020-02-07
[MOB-1185] Migrate to AndroidX

vbabenkoru:
Only changing dependencies and imports, no functional changes.
The SDK will require AndroidX libraries now.


vbabenkoru:
I'll submit another PR for nullability annotations first, to minimize the potential diff conflicts.

PR #200, opened 2020-02-14
[MOB-888] Add nullability annotations for better Kotlin interoperability

vbabenkoru:


iterableapi/src/main/java/com/iterable/iterableapi/IterableAction.java
```diff
@@ -49,7 +54,8 @@ static IterableAction actionOpenUrl(String url) {
         return null;
     }
 
-    static IterableAction actionCustomAction(String customActionName) {
+    @Nullable
+    static IterableAction actionCustomAction(@Nullable String customActionName) {
```
Ayyanchira: name cannot be null right? If it is, then we have not handled the JSONException that might occur

iterableapi/src/main/java/com/iterable/iterableapi/IterableAttributionInfo.java
```diff
@@ -9,12 +12,13 @@
     public final int templateId;
     public final String messageId;
 
-    public IterableAttributionInfo(int campaignId, int templateId, String messageId) {
+    public IterableAttributionInfo(int campaignId, int templateId, @Nullable String messageId) {
```
Ayyanchira: Didnt got the context but can messageID be null here?

iterableapi/src/main/java/com/iterable/iterableapi/IterableAction.java
vbabenkoru: JSONException would not occur if `customActionName` is null, it will just do nothing. But good point, in the two places we're calling this, in the worst case the value is an empty string, but not null.

iterableapi/src/main/java/com/iterable/iterableapi/IterableAttributionInfo.java
vbabenkoru: Practically no, but since the code creating `IterableAttributionInfo` decodes this value from JSON via `optString` and does not explicitly check for null, I've marked this as nullable. At this point I'm just adding annotations trying to minimize code changes.

iterableapi/src/main/java/com/iterable/iterableapi/IterableAction.java
vbabenkoru: Updated

PR #202, opened 2020-02-24
[MOB-1222] Various cleanup

vbabenkoru:
Various cleanup and fixing lint warnings. See individual commits for specifics.


iterableapi/src/main/java/com/iterable/iterableapi/IterableRequest.java
```diff
@@ -298,6 +298,7 @@ public IterableApiRequest(String apiKey, String baseUrl, String resourcePath, JS
 
     public IterableApiRequest(String apiKey, String resourcePath, JSONObject json, String requestType, IterableHelper.SuccessHandler onSuccess, IterableHelper.FailureHandler onFailure) {
         this.apiKey = apiKey;
+        this.baseUrl = null;
```
Ayyanchira: baseUrl is final and we are assigning null here in constructor? When are we assigning the value to it then?

Ayyanchira: oh ok... there is another constructor for that.... Got it...

vbabenkoru: It is only ever set in one of the constructors (the one above this one), and never  changed, so it is safe to make it final and assign to null in all other constructors.

PR #207, opened 2020-02-25
[MOB-1193] - Maintain DialogFragment consistency on rotation

Ayyanchira:
static Notification updated on creation. Retained insetPadding.

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFragmentHTMLNotification.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFragmentHTMLNotification.java
@@ -13,9 +13,6 @@ import android.net.Uri;
 import android.os.Build;
 import android.os.Bundle;
 import android.os.Handler;
-import androidx.annotation.NonNull;
-import androidx.annotation.Nullable;
-import androidx.fragment.app.DialogFragment;
 import android.util.DisplayMetrics;
 import android.view.Display;
 import android.view.Gravity;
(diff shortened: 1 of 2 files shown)


iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFragmentHTMLNotification.java
```diff
@@ -82,16 +90,22 @@ public IterableInAppFragmentHTMLNotification() {
...
         Bundle args = getArguments();
         if (args != null) {
-            htmlString = args.getString("html", null);
-            callbackOnCancel = args.getBoolean("callbackOnCancel", false);
+            htmlString = args.getString(HTML_STRING, null);
+            callbackOnCancel = args.getBoolean(CALLBACK_ON_CANCEL, false);
+            messageId = args.getString(MESSAGE_ID);
+            backgroundAlpha = args.getDouble(BACKGROUND_ALPHA);
+            insetPadding = args.getParcelable(INSET_PADDING);
+            setStyle(DialogFragment.STYLE_NO_FRAME, R.style.Theme_AppCompat_NoActionBar);
```
vbabenkoru: Is there a reason to do this twice, in the constructor and in `onCreate`?

Ayyanchira: Thanks for finding that. Removed it from `onCreate`. Only in constructor

PR #211, opened 2020-03-02
[MOB-1192] Remove animation from Dialog Fragment startup

Ayyanchira:
Hiding and showing in-app before resizing to avoid animation effect.

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFragmentHTMLNotification.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFragmentHTMLNotification.java
@@ -46,7 +46,7 @@ public class IterableInAppFragmentHTMLNotification extends DialogFragment implem
     static IterableInAppLocation location;
 
     private IterableWebView webView;
-    private PageStatus webViewStatus;
+    private boolean loaded;
     private OrientationEventListener orientationListener;
     private boolean callbackOnCancel = false;
     private String htmlString;
@@ -86,7 +86,7 @@ public class IterableInAppFragmentHTMLNotification extends DialogFragment implem
      * HTML In-App Notification
      */
     public IterableInAppFragmentHTMLNotification() {
-        this.webViewStatus = PageStatus.NOT_INITIALIZED;
+        this.loaded = false;
         this.backgroundAlpha = 0;
         this.messageId = "";
         insetPadding = new Rect();
@@ -118,13 +118,13 @@ public class IterableInAppFragmentHTMLNotification extends DialogFragment implem
         location = null;
     }
 
-
     /**
      * Sets the loaded flag
-     * @param webViewStatus
+     *
+     * @param loaded
      */
-    public void setWebViewStatus(PageStatus webViewStatus) {
-        this.webViewStatus = webViewStatus;
+    public void setLoaded(boolean loaded) {
+        this.loaded = loaded;
     }
 
     @NonNull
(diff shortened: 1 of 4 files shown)


Ayyanchira:
I am assuming resize script is called every time for all inapps. Tried on full screen inapps and center aligned inapps

Ayyanchira:
> What happens if it takes time to load? Is it possible to interact with the app while the in-app is on the screen but invisible?

back button works and can still take the user to previous screen. Will have to check what if the app itself is fullscreen and there is no back button in the context.

Ayyanchira:
And also we are showing the window after the resize script is run. So before I check, I assume it is not dependent on the actual data getting loaded. Will check it with the image delay link that you used in one of the in-apps before...

vbabenkoru:
Resize is only called after the entire HTML content is loaded. If there is any asset that takes a long time to load, it will delay the `resize` call.

Ayyanchira:
Oh yes... Thats true...

Ayyanchira:
Oh... one observation:
Even if we use deelayme.com to try delaying the image resource inside the InApps html, onPageFinished is called even before the image is loaded. I think its because the html by itself is loaded from disk and not from web directly. May be thats why on page finished is getting called even before the resources inside aren't completely loaded?

Ayyanchira:
> What happens if it takes time to load? Is it possible to interact with the app while the in-app is on the screen but invisible?

Checked with setting the delay load images of 10 seconds in inapp.. 
Not only we can interact with app while the inapp is invisible, but it exceeded the expectation by showing the inapp 10 seconds later even if we hit back button and switch to different activity altogether (which again is not what we wanted) 😅

Ayyanchira:
Rebased with master

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFragmentHTMLNotification.java
```diff
@@ -170,16 +157,13 @@ public View onCreateView(@NonNull LayoutInflater inflater, @Nullable ViewGroup c
         if (orientationListener == null) {
             orientationListener = new OrientationEventListener(getContext(), SensorManager.SENSOR_DELAY_NORMAL) {
                 public void onOrientationChanged(int orientation) {
-                    // Resize the webview on device rotation
-                    if (loaded) {
```
vbabenkoru, line 174: Do we not need this anymore?

iterableapi/src/main/java/com/iterable/iterableapi/IterableWebViewClient.java
```diff
@@ -27,19 +28,22 @@ public boolean shouldOverrideUrlLoading(WebView view, String url) {
         return true;
     }
 
+    @Override
+    public void onPageStarted(WebView view, String url, Bitmap favicon) {
```
vbabenkoru, line 32: Looks like this is not doing anything.

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFragmentHTMLNotification.java
Ayyanchira: this was a boolean flag which was replaced with tri-state : not loaded, initialized and finished loading, which was then replaced by our new granular refresh.

iterableapi/src/main/java/com/iterable/iterableapi/IterableWebViewClient.java
Ayyanchira: Thanks...
Removed... 👍

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFragmentHTMLNotification.java
vbabenkoru: Hmm, but this was there before.. Was there a reason why we can't resize on rotation until it's loaded? Seems too specific.

PR #213, opened 2020-03-04
[MOB-1219] - Prevent redundant trackinapp calls

Ayyanchira:
Added a check inside oncreate to see if saved bundle contains information about trackInAppOpen.


iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFragmentHTMLNotification.java
```diff
@@ -189,10 +190,18 @@ public void run() {
         RelativeLayout.LayoutParams layoutParams = new RelativeLayout.LayoutParams(RelativeLayout.LayoutParams.MATCH_PARENT,RelativeLayout.LayoutParams.MATCH_PARENT);
         relativeLayout.addView(webView, layoutParams);
 
-        IterableApi.sharedInstance.trackInAppOpen(messageId, location);
+        if ((savedInstanceState == null) || !savedInstanceState.getBoolean(INAPP_OPEN_TRACKED, false)) {
```
vbabenkoru:
```suggestion
        if (savedInstanceState == null || !savedInstanceState.getBoolean(INAPP_OPEN_TRACKED, false)) {
```

PR #216, opened 2020-03-17
[MOB-1286] - support new parameters for update subscription api call

Ayyanchira:
A new method which now accepts subscribedMessageTypeIds, campaignId and templateId for updateSubscription call. Resolves feature parity on iOS and Android.

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
@@ -601,6 +601,10 @@ private static final String TAG = "IterableApi";
      * @param unsubscribedMessageTypeIds
      */
     public void updateSubscriptions(@Nullable Integer[] emailListIds, @Nullable Integer[] unsubscribedChannelIds, @Nullable Integer[] unsubscribedMessageTypeIds) {
+        updateSubscriptions(emailListIds, unsubscribedChannelIds, unsubscribedMessageTypeIds, null, null, null);
+    }
+
+    public void updateSubscriptions(@Nullable Integer[] emailListIds, @Nullable Integer[] unsubscribedChannelIds, @Nullable Integer[] unsubscribedMessageTypeIds, @Nullable Integer[] subcribedMessageTypeIDs, Integer campaignId, Integer templateId) {
         if (!checkSDKInitialization()) {
             return;
         }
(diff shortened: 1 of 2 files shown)


iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -601,6 +601,10 @@ public void disablePush() {
      * @param unsubscribedMessageTypeIds
      */
     public void updateSubscriptions(@Nullable Integer[] emailListIds, @Nullable Integer[] unsubscribedChannelIds, @Nullable Integer[] unsubscribedMessageTypeIds) {
+        updateSubscriptions(emailListIds, unsubscribedChannelIds, unsubscribedMessageTypeIds, null, null, null);
+    }
+
+    public void updateSubscriptions(@Nullable Integer[] emailListIds, @Nullable Integer[] unsubscribedChannelIds, @Nullable Integer[] unsubscribedMessageTypeIds, @Nullable Integer[] subcribedMessageTypeIDs, Integer campaignId, Integer templateId) {
```
vbabenkoru: Other places where `campaignId` and `templateId` are passed are using `int` types... I'm not sure in this case though...

davidtruong: Not sure the implications if you store the Integer object instead of a int in the request Json.
@Ayyanchira Can you make a sure you test the api calls with the backend to make sure this works as intended.

Alternately you could use campaignId.intValue() if you want to explicitly convert it to an int.

Ayyanchira:
```
{
      "email": "victor@iterable.com",
      "emailListIds": [
        382123
      ],
      "unsubscribedChannelIds": [
        232341,
        123452
      ],
      "unsubscribedMessageTypeIds": [
        12334
      ],
      "subscribedMessageTypeIds": [
        43231
      ],
      "campaignId": 2213,
      "templateId": 21243
    }
```
Tried with this above data. Checked on api.iterable.com and the SDK as well. got success response. 
And we can replace it with int. But was there a reason Integer was used?

davidtruong: @Ayyanchira Are you referring to nullable Integer[]? I think that can be nullable int[].

PR #223, opened 2020-04-19
Adds Campaign ID to IterableInAppMessage

nkotula:
This is a very simple edit which exposes the `campaignId` in the `IterableInAppMessage`.  It converts a missing `clientId` key or a null value to an empty string, just like the Swift SDK

I didn't touch anything regarding tests, but I see that there are quite a few test suites that use `IterableInAppMessage`.  If updating tests is something that should be included in this PR, let me know and I can get on it.  All tests are passing on my machine with this change included.


nkotula:
Interesting... That 7 constructor parameter check wouldn't even have been satisfied before I made any changes.  Was there a performance or compiler reason for that check?  Or is just to keep code cleaner?

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppMessage.java
```diff
@@ -7,7 +7,6 @@
 import androidx.annotation.RestrictTo;
 import androidx.core.util.ObjectsCompat;
 
-import org.json.JSONArray;
```
nkotula, line 10: Removed redundant import.

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppMessage.java
```diff
@@ -30,8 +30,18 @@
...
-    IterableInAppMessage(@NonNull String messageId, @NonNull Content content, @NonNull JSONObject customPayload, @NonNull Date createdAt, @NonNull Date expiresAt, @NonNull Trigger trigger, @Nullable Boolean saveToInbox, @Nullable InboxMetadata inboxMetadata) {
+    IterableInAppMessage(@NonNull String messageId,
+                         @NonNull String campaignId,
+                         @NonNull Content content,
+                         @NonNull JSONObject customPayload,
+                         @NonNull Date createdAt,
+                         @NonNull Date expiresAt,
+                         @NonNull Trigger trigger,
+                         @Nullable Boolean saveToInbox,
+                         @Nullable InboxMetadata inboxMetadata) {
```
nkotula, line 41: I reorganized the code here because I find this easier to read, but this is solely a code style change and I could revert this back to one line if that's preferred.

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppMessage.java
```diff
@@ -306,8 +322,17 @@ static IterableInAppMessage fromJSONObject(@NonNull JSONObject messageJson, @Nul
...
+        IterableInAppMessage message = new IterableInAppMessage(
+                messageId,
+                campaignId,
+                new Content(html, padding, backgroundAlpha),
+                customPayload,
+                createdAt,
+                expiresAt,
+                trigger,
+                saveToInbox,
+                inboxMetadata);
```
nkotula, line 334: Again, I find this style of constructor/method invocation where parameters are on individual lines much more readable, but I could very easily revert this as it is only code-style related.  The core change is to simply add the `campaignId` to the constructor invocation.

vbabenkoru:
@nkotula This looks great, thank you!
CampaignId is a number in the API, but since it is declared as `String` in the iOS SDK, it makes sense to maintain parity.
Which parameter check are you referring to?

vbabenkoru:
@nkotula Actually, it might be better to make this `Long` to match the API, and we'll update the iOS implementation to a number in a later major release.

nkotula:
Hi Victor.  For the check, I was referring to the Codeclimate check failure.  When I check the details of that failure, the error is:  "Constructor has 9 parameters, which is greater than 7 authorized." 🤷 

For making the ID a `long`, how should we handle invalid data?  When we parse a received message, it would be simple enough to make this property a Java `Long` so it could be nullable.   When trying to turn it back into JSON, what are the valid values for the `campaignId`?  I'm guessing that `null` and anything less than 1 would be invalid, so we don't write the key to the JSON.  Would that be the correct behavior?

vbabenkoru:
@nkotula Yes, that totally makes sense! I think 0 may also be a valid value.

nkotula:
Can you confirm that range for me?  I'd like to build a check that can't possibly encode an invalid value and send it out to the API.

vbabenkoru:
For a `Long` value, the check would be `!= null && >= 0`.

vbabenkoru:
And don't worry about codeclimate, it is there to help detect maintainability issues but is not required for PRs.

nkotula:
Ready for review again.

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppMessage.java
```diff
@@ -193,6 +203,11 @@ public String getMessageId() {
         return messageId;
     }
 
+    @Nullable
+    public Long getCampaignId() {
```
vbabenkoru, line 207: I'd make it return `long` and default to 0. `Long` is here only to serialize/deserialize properly, for a consumer of this API it shouldn't matter whether it is 0 or null.

nkotula: Okay, trying some push-back here...  I hope I don't overstep my bounds 😅 

As a consumer of any API, I actually prefer the distinction between `null` and a value being present.  `null` to me clearly states that a value is not present, and more importantly, it shows that a value _can_ be missing.  By changing this to a primitive `long`, a consumer of this API might assume the value is always present in every message, which could lead to unexpected behavior.  Also, they would not be able to tell the difference between a campaign ID that was missing from the message, and one where the campaign ID is actually `0`.

Another reason I don't like defaulting to `0` is that it has a kind of "magic number" feeling to it where I as the consumer would have to know what `0` means in the context of this API.  Some maintainer down the line could see a line like this: `if (message.getCampaignId() > 0)`, and think, "wait, is 0 a valid campaign ID?"  If instead they saw this:  `if (message.getCampaignId() != null)` they would immediately know that we're checking if the value is present or not.

Those are my thoughts on the issue.  After programming in Kotlin for so long, I've become accustomed to the ease-of-use surrounding nullables and leveraging them to alleviate the risks I described above.  I know that Java makes null-checking like this very verbose, so if primitives are still preferred, I will change it as you described 😃

vbabenkoru: It is actually present in all API calls per our API spec.

vbabenkoru: Because of this I'm not sure additional null-checking and validation is necessary here..

nkotula: Oh!  Lol that settles that!  I'll change it 👍

vbabenkoru: Sorry for the confusion..

vbabenkoru: And `campaignId` is also declared as plain `int` (should make it `long` at some point) elsewhere, in `IterableAttributionInfo`. I would make it `long` in this case, the only reason to keep it nullable is to make the tests pass until we add `campaignId` to all test JSON payloads (how did I even miss that?).

nkotula: Oh!  Okay, so do the check for `null` and convert `null` to `0` in the getter?

nkotula:
Ready for review again.

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppMessage.java
```diff
@@ -281,6 +299,7 @@ static IterableInAppMessage fromJSONObject(@NonNull JSONObject messageJson, @Nul
         }
 
         String messageId = messageJson.optString(IterableConstants.KEY_MESSAGE_ID);
+        final Long campaignId = IterableUtil.retrieveValidCampaignIdOrNull(messageJson, IterableConstants.KEY_CAMPAIGN_ID);
```
davidtruong, line 302: Why not just use an inline ternary to rather than creating separate retrieval/validation methods?

final Long campaignId = (messageJson.isNull(IterableConstants.KEY_CAMPAIGN_ID)) ? null :
     jsonObject.optLong(IterableConstants.KEY_CAMPAIGN_ID);

davidtruong: I see from the comments that you are checking for a valid non negative campaignId. Not sure if that check is necessary.

nkotula: My thought was to centralize validation logic to be used elsewhere.  That, and my thought was that it shouldn't be the responsibility of `IterableInAppMessage` to know how to validate the ID should that validation ever change in the future.  That may not be relevant here, but that's what I was considering when breaking this out into a Util method.

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppMessage.java
nkotula: So... @vbabenkoru I just did a check with an SSL proxy, and I see that when I send proofs to my device, the `campaignId` is indeed _not_ present in the raw message JSON with calls to `getMessages?email=my.email.here@example.com`.  Are we _sure_ we shouldn't keep this nullable?

vbabenkoru: @nkotula Just went to check again.. You're right, sorry for the confusion. It totally makes sense to make it a nullable `Long`.

nkotula: Okay, cool beans.  I'll go ahead and revert that change.

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppMessage.java
vbabenkoru: @nkotula Do you prefer to validate all data from API responses? My thought was that if there is an API spec, we shouldn't have to do client-side validation (unless it is something that may cause crashes).

nkotula: I do prefer to sanity-check data integrity on the client-side, but that is totally a personal preference.  To me, having both client-side and server-side validation of the same data makes the system as a whole more resilient because they can provide checks against each other, but it comes at the cost of a larger burden on the maintainers.  I will defer to your team as to whether this should be moved to an inline ternary operation as David mentioned.

vbabenkoru: I think it's ok to keep it. Thanks!

vbabenkoru (approved):
Thank you!

PR #224, opened 2020-04-23
Close streams after reading file

mictab:
Hi!

While using the SDK, we noticed when using StrictMode that these streams weren't being closed. So I thought that we should patch it.

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableUtil.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableUtil.java
@@ -8,7 +8,7 @@ import android.content.pm.PackageManager;
 import androidx.annotation.Nullable;
 import androidx.annotation.VisibleForTesting;
 
-import com.google.android.gms.common.util.IOUtils;
+import com.iterable.iterableapi.util.IOUtils;
 
 import org.json.JSONObject;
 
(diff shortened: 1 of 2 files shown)


iterableapi/src/main/java/com/iterable/iterableapi/IterableUtil.java
```diff
@@ -8,6 +8,8 @@
 import androidx.annotation.Nullable;
 import androidx.annotation.VisibleForTesting;
 
+import com.google.android.gms.common.util.IOUtils;
```
vbabenkoru, line 11: Is this public API? Is it possible that this might change?

vbabenkoru: Looks like it is not documented here:
https://developers.google.com/android/reference/com/google/android/gms/common/package-summary
I'd be cautious using this in a library since an update to `play-services-basement` might break our artifact at runtime.

mictab: Basically what that one does is to wrap the close in a try-catch and fail silently if for whatever reason the close throws. Would you prefer a manual implementation of that instead?

vbabenkoru: Yeah, if it's a few lines of code, I'd prefer to move it to one of our util classes.

mictab:
Thanks for the review @vbabenkoru. Feel free to have another look :)

vbabenkoru:
@mictab Looks good, thank you for the PR!

PR #229, opened 2020-05-07
[MOB-1424] InAppFragment improvements, delayed display

vbabenkoru:
* Hide the webView instead of DecorView for compatibility with API 29+
* Show the dialog when the WebView is fully loaded, or after 500 ms, whichever happens earlier

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFragmentHTMLNotification.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFragmentHTMLNotification.java
@@ -41,6 +41,8 @@ public class IterableInAppFragmentHTMLNotification extends DialogFragment implem
     private static final String MESSAGE_ID = "MessageId";
     private static final String INAPP_OPEN_TRACKED = "InAppOpenTracked";
 
+    private static final int DELAY_THRESHOLD_MS = 500;
+
     static IterableInAppFragmentHTMLNotification notification;
     static IterableHelper.IterableUrlCallback clickCallback;
     static IterableInAppLocation location;
@@ -125,6 +127,7 @@ public class IterableInAppFragmentHTMLNotification extends DialogFragment implem
      */
     public void setLoaded(boolean loaded) {
         this.loaded = loaded;
+        showDialogView();
     }
 
     @NonNull
@@ -148,7 +151,6 @@ public class IterableInAppFragmentHTMLNotification extends DialogFragment implem
         if (insetPadding.bottom == 0 && insetPadding.top == 0) {
             dialog.getWindow().setFlags(WindowManager.LayoutParams.FLAG_FULLSCREEN, WindowManager.LayoutParams.FLAG_FULLSCREEN);
         }
-        hideDialogView(dialog);
         return dialog;
     }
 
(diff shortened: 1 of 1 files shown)


iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFragmentHTMLNotification.java
```diff
@@ -190,20 +192,27 @@ public void run() {
...
+    private void hideDialogView() {
         try {
-            dialog.getWindow().getDecorView().setVisibility(View.INVISIBLE);
+            webView.setVisibility(View.INVISIBLE);
+            webView.postDelayed(new Runnable() {
+                @Override
+                public void run() {
+                    showDialogView();
+                }
+            }, DELAY_THRESHOLD_MS);
```
Ayyanchira, line 207: 👌

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFragmentHTMLNotification.java
```diff
@@ -190,20 +192,27 @@ public void run() {
...
         }
+        hideDialogView();
         return relativeLayout;
     }
 
-    private void hideDialogView(Dialog dialog) {
+    private void hideDialogView() {
         try {
-            dialog.getWindow().getDecorView().setVisibility(View.INVISIBLE);
+            webView.setVisibility(View.INVISIBLE);
```
Ayyanchira, line 201: Just a suggestion. With this approach the status bar on API 29 device seems to flash before hiding and showing.

getDialog().getWindow().getDecorView().setAlpha(0);
can be a good alternative and more closer to what we had achieved previously.

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFragmentHTMLNotification.java
```diff
@@ -190,20 +192,27 @@ public void run() {
...
         } catch (NullPointerException e) {
             IterableLogger.e(TAG, "View not present. Failed to hide before resizing inapp");
         }
     }
 
-    private void showDialogView(Dialog dialog) {
+    private void showDialogView() {
         try {
-            dialog.getWindow().getDecorView().setVisibility(View.VISIBLE);
+            webView.setVisibility(View.VISIBLE);
```
Ayyanchira, line 215: getDialog().getWindow().getDecorView().setAlpha(1);

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFragmentHTMLNotification.java
vbabenkoru: If Android suspends everything when `DecorView` is invisible, I'm wondering if they might also do that for opacity 0 at some point..

Ayyanchira: I dont think so... 
As opacity is more of cosmetic property (which can be used to animate). It also means that the space will be acquired by the view even if it's not visible because it is present.
While invisibility (with reference to iOS framework) actually has potential to remove the object. (Like for e.g A Button when hidden is actually removed from view hierarchy)

vbabenkoru: `View.VISIBLE` is supposed to be like that: it makes it invisible without removing it from the layout (unlike what `View.GONE` does).

Ayyanchira: Thats true too 🤷‍♂️

vbabenkoru: Yeah, alpha looks better, changed to that.

PR #236, opened 2020-05-27
[MOB-1482] Add Device Attributes methods

Ayyanchira:
Adding methods to add custom device attributes, remove device attributes. Integrated it with existing push registration and test methods.

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
@@ -22,6 +22,7 @@ import org.json.JSONException;
 import org.json.JSONObject;
 
 import java.util.Arrays;
+import java.util.HashMap;
 import java.util.List;
 import java.util.UUID;
 
@@ -52,6 +53,7 @@ private static final String TAG = "IterableApi";
 
     private IterableInAppManager inAppManager;
     private String inboxSessionId;
+    private HashMap<String, String> deviceAttributes = new HashMap<>();
 
 //---------------------------------------------------------------------------------------
 //endregion
@@ -200,6 +202,18 @@ private static final String TAG = "IterableApi";
         if (data != null) {
             setAttributionInfo(new IterableAttributionInfo(data.getCampaignId(), data.getTemplateId(), data.getMessageId()));
         }
+    }
+
+    HashMap getDeviceAttributes() {
+        return deviceAttributes;
+    }
+
+    public void setDeviceAttribute(String key, String value) {
+        deviceAttributes.put(key, value);
+    }
+
+    public void removeFromDeviceAttribute(String key) {
+        deviceAttributes.remove(key);
     }
 //---------------------------------------------------------------------------------------
 //endregion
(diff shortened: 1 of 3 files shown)


iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -52,6 +55,7 @@
 
     private IterableInAppManager inAppManager;
     private String inboxSessionId;
+    private HashMap<String, String> deviceAttributes;
```
vbabenkoru, line 58: Why not initialize it here?

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -201,6 +205,23 @@ void setNotificationData(IterableNotificationData data) {
             setAttributionInfo(new IterableAttributionInfo(data.getCampaignId(), data.getTemplateId(), data.getMessageId()));
         }
     }
+
+    HashMap getDeviceAttributes() {
+        return deviceAttributes;
+    }
+
+    public void setDeviceAttribute(String key, String value) {
+        if (deviceAttributes == null) {
```
vbabenkoru, line 214: You can remove these checks if you initialize it above.

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -1114,6 +1135,15 @@ protected void registerDeviceToken(@Nullable String email, @Nullable String user
             if (dataFields == null) {
                 dataFields = new JSONObject();
             }
+
+            if (deviceAttributes != null) {
+                Iterator<HashMap.Entry<String, String>> deviceAttributesIterator = deviceAttributes.entrySet().iterator();
```
vbabenkoru, line 1140: You can just do `for(HashMap.Entry<String, String> entry : deviceAttributes.entrySet())`

iterableapi/src/main/java/com/iterable/iterableapi/IterablePushRegistration.java
```diff
@@ -33,7 +34,8 @@ protected Void doInBackground(IterablePushRegistrationData... params) {
                             iterablePushRegistrationData.email,
                             iterablePushRegistrationData.userId,
                             iterablePushRegistrationData.pushIntegrationName,
-                            pushRegistrationObject.token);
+                            pushRegistrationObject.token,
+                            IterableApi.getInstance().getDeviceAttributes());
```
vbabenkoru, line 38: Hmmm I'm wondering if we should put device attributes in `iterablePushRegistrationData` since we store everything else related to the push registration there.. I don't think this will make a significant difference, but could be better for consistency. Your call.

Ayyanchira: Sticking with current one. I had thought about it too. DeviceAttributes is something developer can set on later stage too apart from during a push registration state. PushRegistration felt more related to Push and Firebase side of functionality and didn't felt like adding it there. This looks fine to me :)

vbabenkoru: The reason these properties are packed within the object is getting the token takes some time to get the token, so we capture the email, userId, push integration name at the time push registration was requested to make sure it is consistent.

vbabenkoru: Oh, actually we should keep it this way to make sure device attributes can be set after initialization. :) Disregard this.

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -201,6 +203,18 @@ void setNotificationData(IterableNotificationData data) {
...
+
+    HashMap getDeviceAttributes() {
+        return deviceAttributes;
+    }
+
+    public void setDeviceAttribute(String key, String value) {
+        deviceAttributes.put(key, value);
+    }
+
+    public void removeFromDeviceAttribute(String key) {
```
vbabenkoru, line 215:
```suggestion
    public void removeDeviceAttribute(String key) {
```

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -1114,6 +1128,13 @@ protected void registerDeviceToken(@Nullable String email, @Nullable String user
             if (dataFields == null) {
                 dataFields = new JSONObject();
             }
+
+            if (deviceAttributes != null) {
```
vbabenkoru, line 1132: It is never null

PR #240, opened 2020-06-23
[MOB-1754] - Disable Device for all users

Ayyanchira:
Email and UserID can be null. In this situation, all the users using the device token will be disabled according to the backend docks. Also ReactNative will be having seperate methods to disableAllUsers and disableSpecific users. This will enable it too.

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
@@ -1093,9 +1093,9 @@ private static final String TAG = "IterableApi";
 
     /**
      * Internal api call made from IterablePushRegistration after a registrationToken is obtained.
-     * It disables device under all users with this device by default. If a emailID or userID is provided, it will disable the device for specific user.
-     * @param email Specific email this device belongs to disable
-     * @param userId Specific userId this device belongs to disable
+     * It disables the device for all users with this device by default. If `email` or `userId` is provided, it will disable the device for the specific user.
+     * @param email User email for whom to disable the device.
+     * @param userId User ID for whom to disable the device.
      * @param token The device token
      */
     protected void disableToken(@Nullable String email, @Nullable String userId, @NonNull String token, @Nullable IterableHelper.SuccessHandler onSuccess, @Nullable IterableHelper.FailureHandler onFailure) {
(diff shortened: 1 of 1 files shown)


iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -1093,15 +1093,18 @@ protected void disableToken(@Nullable String email, @Nullable String userId, @No
 
     /**
      * Internal api call made from IterablePushRegistration after a registrationToken is obtained.
-     * @param token
+     * It disables device under all users with this device by default. If a emailID or userID is provided, it will disable the device for specific user.
+     * @param email Specific email this device belongs to disable
+     * @param userId Specific userId this device belongs to disable
+     * @param token The device token
      */
     protected void disableToken(@Nullable String email, @Nullable String userId, @NonNull String token, @Nullable IterableHelper.SuccessHandler onSuccess, @Nullable IterableHelper.FailureHandler onFailure) {
```
vbabenkoru, line 1101: Hmm, this method is protected.. When would it be called with email and userId = null?

Ayyanchira: This will be called from ReactNativeJS SDK. I dont know but the Doc says that if its not specified, it will disable the device for all users.

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -1093,15 +1093,18 @@ protected void disableToken(@Nullable String email, @Nullable String userId, @No
 
     /**
      * Internal api call made from IterablePushRegistration after a registrationToken is obtained.
-     * @param token
+     * It disables device under all users with this device by default. If a emailID or userID is provided, it will disable the device for specific user.
```
vbabenkoru, line 1096:
```suggestion
     * It disables the device for all users with this device by default. If `email` or `userId` is provided, it will disable the device for the specific user.
```

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -1093,15 +1093,18 @@ protected void disableToken(@Nullable String email, @Nullable String userId, @No
...
+     * @param email Specific email this device belongs to disable
```
vbabenkoru, line 1097:
```suggestion
     * @param email User email for whom to disable the device.
```

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -1093,15 +1093,18 @@ protected void disableToken(@Nullable String email, @Nullable String userId, @No
...
+     * @param userId Specific userId this device belongs to disable
```
vbabenkoru, line 1098:
```suggestion
     * @param userId User ID for whom to disable the device.
```

PR #244, opened 2020-07-02
[MOB-1788] Clear in-app messages on logout

vbabenkoru:
Make sure we clear the in-app queue immediately on logout.

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
@@ -1328,11 +1328,7 @@ private static final String TAG = "IterableApi";
     }
 
     private void onLogOut() {
-        if (!isInitialized()) {
-            return;
-        }
-
-        if (config.autoPushRegistration) {
+        if (config.autoPushRegistration && isInitialized()) {
             disablePush();
         }
         getInAppManager().reset();
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
@@ -1328,13 +1328,22 @@ private static final String TAG = "IterableApi";
     }
 
     private void onLogOut() {
-        if (config.autoPushRegistration && isInitialized()) {
+        if (!isInitialized()) {
+            return;
+        }
+
+        if (config.autoPushRegistration) {
             disablePush();
         }
+        getInAppManager().reset();
     }
 
     private void onLogIn() {
-        if (config.autoPushRegistration && isInitialized()) {
+        if (!isInitialized()) {
+            return;
+        }
+
+        if (config.autoPushRegistration) {
             registerForPush();
         }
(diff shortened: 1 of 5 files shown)


iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -1328,13 +1328,22 @@ private void retrieveEmailAndUserId() {
...
+        if (config.autoPushRegistration) {
             disablePush();
         }
+        getInAppManager().reset();
     }
 
     private void onLogIn() {
-        if (config.autoPushRegistration && isInitialized()) {
+        if (!isInitialized()) {
+            return;
```
Ayyanchira, line 1343: Even after login if we haven't initialized, shouldn’t we attempt to initialize??

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppManager.java
```diff
@@ -167,6 +167,20 @@ public void execute(String payload) {
         });
     }
 
+    /**
+     * Clear all in-app messages.
+     * Should be called on user logout.
+     */
```
Ayyanchira, line 173: Feels like reset name should be for something which is resetting everything...
May be more specific name like clearAllMessagesFromLocalStorage()

iterableapi/src/test/java/com/iterable/iterableapi/IterableInAppManagerTest.java
```diff
@@ -150,6 +150,20 @@ public void testSyncInApp() throws Exception {
...
+        dispatcher.enqueueResponse("/inApp/getMessages", new MockResponse().setBody(IterableTestUtils.getResourceString("inapp_payload_single.json")));
+        IterableInAppManager inAppManager = IterableApi.getInstance().getInAppManager();
+        assertEquals(0, inAppManager.getMessages().size());
+
+        inAppManager.syncInApp();
+        Robolectric.flushBackgroundThreadScheduler();
+        Robolectric.flushForegroundThreadScheduler();
+        assertEquals(1, inAppManager.getMessages().size());
+        inAppManager.reset();
+        assertEquals(0, inAppManager.getMessages().size());
```
Ayyanchira, line 164: 👍

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
vbabenkoru: Here is the code for `isInitialized()`:
```
    private boolean isInitialized() {
        return _apiKey != null && (_email != null || _userId != null);
    }
```
Basically, if there is no email or userId, there is no user to operate with.

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppManager.java
vbabenkoru: I'm mirroring the name in iOS for consistency.. Technically it does reset everything in the in-app manager.

PR #253, opened 2020-07-21
[MOB-1896] allow token to be set with same email or userId

roninopf:


iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -288,20 +288,20 @@ public void setEmail(@Nullable String email) {
...
+            storeAuthData();
 
-        onLogOut();
-        _email = email;
-        _userId = null;
-        _authToken = authToken;
-        storeAuthData();
-        onLogIn();
+            onLogIn();
+        } else if (email != null && (_authToken == null || _authToken != authToken)) {
```
Ayyanchira, line 301: I think it should be 
```
else if (email == _email && (_authToken == null || _authToken != authToken)) {
//update the authToken
}
```
Only if the email passed in the function is the one already been set in Iterable, we should update the token. 
If the email is new, we should perform log out, log in and then store the auth.

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -324,20 +324,20 @@ public void setUserId(@Nullable String userId) {
...
+            storeAuthData();
 
-        onLogOut();
-        _email = null;
-        _userId = userId;
-        _authToken = authToken;
-        storeAuthData();
-        onLogIn();
+            onLogIn();
+        } else if (userId != null & (_authToken == null || _authToken != authToken)) {
```
Ayyanchira, line 337: Just like with email, 

we might have to check if userId == !userId && (_authToken == null || _authToken != authToken)) {
//update authToken
}

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -288,20 +288,20 @@ public void setEmail(@Nullable String email) {
      * @param authToken Authorization token
      */
     public void setEmail(@Nullable String email, @Nullable String authToken) {
-        if (_email != null && _email.equals(email)) {
-            return;
-        }
+        if (_email != null && !_email.equals(email)) {
```
Ayyanchira, line 291: Looks like this conditions will fail if the emailId is already null and we want to set a email? 

This will mostly occur during initialization where email is not set yet.

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -288,20 +288,20 @@ public void setEmail(@Nullable String email) {
...
 
-        onLogOut();
-        _email = email;
-        _userId = null;
-        _authToken = authToken;
-        storeAuthData();
-        onLogIn();
+            onLogIn();
+        } else if (email != null && email.equals(_email) && (_authToken == null || _authToken != authToken)) {
+            _authToken = authToken;
```
Ayyanchira, line 302: we will need to set _email to new email

roninopf: It's already the same by that point

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
vbabenkoru, line 291: This will skip execution if `_email` (current email) is null. The tests caught that.

roninopf: Ah, because it should empty out authentication info if it's `null`.

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
Ayyanchira: Not when _email(current email) is null as stated in above comments. If null case is handled on top, `else if` part can be considered as just updating the authToken

roninopf: This got really confusing to follow the flow for something so important, so I simplified it. The handling of `null` and setting info accurately should be right, now, if I haven't messed anything else up!

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -288,20 +288,29 @@ public void setEmail(@Nullable String email) {
      * @param authToken Authorization token
      */
     public void setEmail(@Nullable String email, @Nullable String authToken) {
-        if (_email != null && _email.equals(email)) {
-            return;
-        }
+        if (email == null) {
```
vbabenkoru, line 291: I think this still doesn't do what we want it to do.. We don't want to run logout/login if current email & userId are null, and the new email is also null.

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -288,20 +288,29 @@ public void setEmail(@Nullable String email) {
...
 
-        if (_email == null && _userId == null && email == null) {
-            return;
-        }
+            _email = null;
+            _userId = null;
+            _authToken = null;
+
+            storeAuthData();
+        } else if (_email.equals(email) && _authToken != authToken) {
```
vbabenkoru, line 299: Since we don't check for nullity of `_email` here, if we pass an non-null `email` when the current `_email` is `null`, it will throw an exception.

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -289,6 +289,17 @@ public void setEmail(@Nullable String email) {
      */
     public void setEmail(@Nullable String email, @Nullable String authToken) {
         if (_email != null && _email.equals(email)) {
+            if (_authToken == null && authToken == null) {
+                return;
```
Ayyanchira, line 293: if we `return` here, 
we won't perform setEmail for where developer doesnt want to implement authToken at all. `authToken` and `_authToken` being null should still allow setEmail operation

roninopf: Hmm, this is only for when the previous and new emails are the same. Nothing should happen, right? That was the previous behavior too. It just used to `return`.

Ayyanchira: oh yes.. looks good here.

Ayyanchira: I strongly feel we should have multiple tests written for it first. Kind of if those pass, all good

roninopf: I'll see if I can fit tests into this PR.

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -1223,6 +1245,10 @@ protected void registerDeviceToken(@Nullable String email, @Nullable String user
         }
     }
 
+    protected String getAuthToken() {
```
vbabenkoru, line 1248: What do we need this for?

iterableapi/src/test/java/com/iterable/iterableapi/IterableApiTest.java
```diff
@@ -630,4 +630,194 @@ public void testTrackInAppDeleteWithNullParameters() throws Exception {
         assertNull(requestJson.optString(IterableConstants.KEY_INBOX_SESSION_ID, null));
     }
 
+    @Test
+    public void testSetEmailWithTokenPersistence() throws Exception {
```
vbabenkoru, line 634: Since these tests aren't testing persistence, I think these shouldn't have `*Persistence` suffixes.
The tests checking persistence create a new instance of `IterableApi` to make sure it doesn't have values in memory and loads them from disk.

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
Ayyanchira: Oh! I suggested that in case we need to check authToken within our IterableAPI classes internally. Could be useful in test. And possibly in RN too (later).

roninopf: Yeah, it was only to let the tests verify the authentication token directly. Otherwise the tests had no (easy) way to check them.

iterableapi/src/test/java/com/iterable/iterableapi/IterableApiTest.java
roninopf: Will change the suffix

roninopf: Renamed to `*Status`. There might be something better? But at least now it's not implying disk data.

iterableapi/src/test/java/com/iterable/iterableapi/IterableApiAuthTests.java
```diff
@@ -0,0 +1,232 @@
...
+    }
+
+    private void reInitIterableApi() {
+        IterableApi.sharedInstance = spy(new IterableApi());
+        IterableInAppManager inAppManagerMock = mock(IterableInAppManager.class);
+        doReturn(inAppManagerMock).when(IterableApi.sharedInstance).getInAppManager();
+    }
+
+    @Test
+    public void testSetEmailWithTokenStatus() throws Exception {
```
vbabenkoru, line 43: What does `*Status` prefix mean?

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
vbabenkoru: Hmm, interesting, I didn't know `protected` was accessible from tests.. I've always used package-private (no modifier) for this.

vbabenkoru: "Protected data member and method are only accessible _by the classes of the same package_ and the subclasses present in any package."
TIL! I thought they were only accessible by the subclasses.

Ayyanchira: Oh!!! When I read `package-private` above, I realized thats I wanted to tell before!😅 But yeah strange that protected worked too

vbabenkoru: However in this case we don't have subclasses of `IterableApi`, so it would also work with package-private. Let me take a look at how `getEmail`/`getUserId` are declared..

iterableapi/src/test/java/com/iterable/iterableapi/IterableApiAuthTests.java
roninopf: It's supposed to represent the status of the authentication data; I couldn't think of a good suffix that didn't make it more confusing, but I wanted to at least evade `*Persistence`.

Ayyanchira: Just plain 
`testSetEmailWithToken`
`testSetEmailEmptyToken`
`testSetEmailWithNullToken`
`testSetNullEmailWithToken`
kind of names would also work

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
vbabenkoru: They're package-private.. I think next to them would also be a better place for it. @roninopf Could you remove `protected` keyword and move this method right below `getUserId()` in "Package-Protected Functions" group?

iterableapi/src/test/java/com/iterable/iterableapi/IterableApiAuthTests.java
vbabenkoru: I think it won't lose anything without the suffix.. `testSetEmailWithTokenStatus` makes sense.

roninopf: Okay! I'll just drop it. I think you're right.

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
roninopf: Done

Ayyanchira: image
Is it protected causing the failure in travis?

vbabenkoru: @Ayyanchira Oh, right. Let's make it package-private to make it visible to tests.

PR #259, opened 2020-08-06
[MOB-1910] Resolve channel name as a string resource without logging warnings

### LINKED ISSUE #250: Resolve String / Resource ID For Channel Name
Iterable currently does not resolve the channel name correctly from the Manifest file. It expects a direct string when the value could be a resource ID. 

The following error is thrown.

```
Bundle: Key iterable_notification_channel_name expected String but value was a java.lang.Integer.  The default value <null> was returned.
Bundle: Attempt to cast generated internal exception:
    java.lang.ClassCastException: java.lang.Integer cannot be cast to java.lang.String
        at android.os.BaseBundle.getString(BaseBundle.java:1168)
        at com.iterable.iterableapi.IterableNotificationHelper$IterableNotificationHelperImpl.getChannelName(IterableNotificationHelper.java:279)
        at com.iterable.iterableapi.IterableNotificationHelper$IterableNotificationHelperImpl.createNotification(IterableNotificationHelper.java:101)
        at com.iterable.iterableapi.IterableNotificationHelper.createNotification(IterableNotificationHelper.java:36)
        at com.iterable.iterableapi.IterableFirebaseMessagingService.handleMessageReceived(IterableFirebaseMessagingService.java:60)
        at com.treatmyocd.nocd.service.NocdFirebaseMessagingService.onMessageReceived(NocdFirebaseMessagingService.kt:72)
        at com.google.firebase.messaging.FirebaseMessagingService.zzc(com.google.firebase:firebase-messaging@@20.2.1:80)
        at com.google.firebase.messaging.zzh.run(com.google.firebase:firebase-messaging@@20.2.1:2)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1167)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:641)
        at com.google.android.gms.common.util.concurrent.zza.run(Unknown Source:6)
        at java.lang.Thread.run(Thread.java:919)

```
The error is in the `getChannelName` method in `IterableNotificationHelper`. 

MaxthonSnap20200717124105

Thankfully, the exception is caught in the try/catch.

vbabenkoru:
Fixes #250.
`getString` was logging warnings when a string resource was provided via `android:resource`. This changes the code a bit so we don't call `getString`/`getInt`. The logic is still the same.

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableNotificationHelper.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableNotificationHelper.java
@@ -275,13 +275,16 @@ class IterableNotificationHelper {
             try {
                 ApplicationInfo info = context.getPackageManager().getApplicationInfo(context.getPackageName(), PackageManager.GET_META_DATA);
                 if (info.metaData != null) {
-                    // Literal string value
-                    channelName = info.metaData.getString(IterableConstants.NOTIFICATION_CHANNEL_NAME);
-
-                    // Try to read from a string resource
-                    int stringId = info.metaData.getInt(IterableConstants.NOTIFICATION_CHANNEL_NAME);
-                    if (channelName == null && stringId != 0) {
-                        channelName = context.getString(stringId);
+                    Object channelNameMetaData = info.metaData.get(IterableConstants.NOTIFICATION_CHANNEL_NAME);
+                    if (channelNameMetaData instanceof String) {
+                        // Literal string value
+                        channelName = (String) channelNameMetaData;
+                    } else if (channelNameMetaData instanceof Integer) {
+                        // Try to read from a string resource
+                        int stringId = (Integer) channelNameMetaData;
+                        if (stringId != 0) {
+                            channelName = context.getString(stringId);
+                        }
                     }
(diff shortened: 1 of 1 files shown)


iterableapi/src/main/java/com/iterable/iterableapi/IterableNotificationHelper.java
```diff
@@ -275,13 +275,16 @@ private String getChannelName(Context context) {
...
+                    Object channelNameMetaData = info.metaData.get(IterableConstants.NOTIFICATION_CHANNEL_NAME);
+                    if (channelNameMetaData instanceof String) {
+                        // Literal string value
+                        channelName = (String) channelNameMetaData;
+                    } else if (channelNameMetaData instanceof Integer) {
+                        // Try to read from a string resource
+                        int stringId = (Integer) channelNameMetaData;
+                        if (stringId != 0) {
+                            channelName = context.getString(stringId);
+                        }
```
Ayyanchira, line 287: I hope after all this, channelName never turns out null 😅

vbabenkoru: It will be `null` if nothing is specified in the manifest. We then use the default value.

PR #260, opened 2020-08-14
[MOB-1920] - Include javadoc in travis

Ayyanchira:


.travis.yml
```diff
@@ -69,6 +69,7 @@ script:
   - ./gradlew :iterableapi:jacocoTestDebugUnitTestReport :app:jacocoDebugTestReport :iterableapi:connectedCheck :app:jacocoDebugAndroidTestReport
 
 after_script:
+  - ./gradlew :iterableapi:javadoc
```
vbabenkoru, line 72: Does this actually fail if you have invalid javadoc? Have you tested it?

Ayyanchira: I saw it getting run in the logs and not failing as the docs were intact.. Will try to push some wrong documentation in this branch and see if its failing or not.

Ayyanchira: Force pushed with wrong docs.. Lets see 🤞

Ayyanchira:
Checked... it does not fail even with wrong docs 😦.. Trying to run that gradlew command separately in regular script section

PR #262, opened 2020-08-24
[MOB-1876] - Consume a message after click/back

Ayyanchira:
If a message is consume type, remove the message only after click events on in-app OR if pressed BACK_BUTTON.

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFragmentHTMLNotification.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFragmentHTMLNotification.java
@@ -243,6 +243,7 @@ public class IterableInAppFragmentHTMLNotification extends DialogFragment implem
         if (clickCallback != null) {
             clickCallback.execute(Uri.parse(url));
         }
+        processMessageRemoval();
         dismiss();
     }
 
@@ -252,6 +253,14 @@ public class IterableInAppFragmentHTMLNotification extends DialogFragment implem
     public void onBackPressed() {
         IterableApi.sharedInstance.trackInAppClick(messageId, BACK_BUTTON);
         IterableApi.sharedInstance.trackInAppClose(messageId, BACK_BUTTON, IterableInAppCloseAction.BACK, location);
+        processMessageRemoval();
+    }
+
+    private void processMessageRemoval() {
+        IterableInAppMessage message = IterableApi.sharedInstance.getInAppManager().getMessageById(messageId);
+        if (message.isMarkedForDeletion() && !message.isConsumed()) {
+            IterableApi.sharedInstance.getInAppManager().removeMessage(message, IterableInAppDeleteActionType.OTHER, location);
+        }
     }
 
     /**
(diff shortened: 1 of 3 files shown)


iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFragmentHTMLNotification.java
```diff
@@ -252,6 +252,11 @@ public void onUrlClicked(String url) {
     public void onBackPressed() {
         IterableApi.sharedInstance.trackInAppClick(messageId, BACK_BUTTON);
         IterableApi.sharedInstance.trackInAppClose(messageId, BACK_BUTTON, IterableInAppCloseAction.BACK, location);
+        IterableInAppMessage message = IterableApi.sharedInstance.getInAppManager().getMessageById(messageId);
```
vbabenkoru, line 255: Hmm at this point I think we should have a common method for handling both normal clicks and back click, since the handling for both is the same.

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppManager.java
```diff
@@ -235,8 +238,7 @@ public void execute(Uri url) {
         })) {
             setRead(message, true);
             if (consume) {
-                // Remove the message without tracking
-                removeMessage(message);
+                message.setConsumed(true);
```
vbabenkoru, line 241: I think doing `setConsumed` changes the semantics of what 'consumed' means. Previously it was an indicator that we've sent a `consume` API call to the server. Now it's not very clear what it means.

Ayyanchira: Should I introduce new variable - markForRemoval? in inAppMessage?

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFragmentHTMLNotification.java
```diff
@@ -252,6 +253,14 @@ public void onUrlClicked(String url) {
     public void onBackPressed() {
         IterableApi.sharedInstance.trackInAppClick(messageId, BACK_BUTTON);
         IterableApi.sharedInstance.trackInAppClose(messageId, BACK_BUTTON, IterableInAppCloseAction.BACK, location);
+        processMessageRemoval();
+    }
+
+    private void processMessageRemoval() {
+        IterableInAppMessage message = IterableApi.sharedInstance.getInAppManager().getMessageById(messageId);
+        if (message.isMarkedForDeletion() && !message.isConsumed()) {
+            IterableApi.sharedInstance.getInAppManager().removeMessage(message, IterableInAppDeleteActionType.OTHER, location);
```
vbabenkoru, line 262: I think this will track a delete event, unlike the previous code that was calling `removeMessage` with just one parameter. Is there a reason to use another version of `removeMessage` here?

Ayyanchira: I think we introduced DeleteActionType.Other for this purposes. Also thought it would be better to pass location if we know the context of where the inapp was being displayed.

May be we can discuss this in future. Reverting back to using removeMessage with single parameter

vbabenkoru (approved):

PR #264, opened 2020-09-03
[MOB-2075] - Handle 401 jwt error code

Ayyanchira:
Code is still 401.
Just an additional check if the message returned is "JWT Authorization header is not set", as currently sent by our backend.
If there are different types of error messages for jwt, I would need to update that here. Or have some different approach.
For e.g, "invalid authKey", "expired authKey" etc


Ayyanchira:
Changes has been accommodated by DT in the main MOB 2006 branch. Hence closing this PR

PR #265, opened 2020-09-15
Feature/mob 2006 jwt mobile auth management

davidtruong:
READY 

Jwt auth management: https://iterable.quip.com/ZtIfAqGabyXT/Mobile-Auth-Management-Updates


vbabenkoru (changes requested):
@davidtruong It seems like it will request a new token multiple times in parallel if multiple network calls fail in sequence. I think it should only request once and wait until it completes before spawning new requests.

iterableapi/src/main/java/com/iterable/iterableapi/IterablePushRegistration.java
```diff
@@ -41,9 +41,13 @@ protected Void doInBackground(IterablePushRegistrationData... params) {
                     IterableApi.sharedInstance.disableToken(
                             iterablePushRegistrationData.email,
                             iterablePushRegistrationData.userId,
-                            pushRegistrationObject.token);
+                            pushRegistrationObject.token,
+                            null,
+                            null,
+                            iterablePushRegistrationData.priorAuthToken
```
vbabenkoru, line 47: I think we should just store `authToken` with email and userId everywhere in this class, to keep it more consistent. Basically, if we're being passed `email` and `userId`, `authToken` should also be there, because it is tightly coupled with email and userId.

iterableapi/src/main/java/com/iterable/iterableapi/IterablePushRegistration.java
```diff
@@ -97,7 +101,7 @@ private void disableOldDeviceIfNeeded() {
                             public void onSuccess(@NonNull JSONObject data) {
                                 sharedPref.edit().putBoolean(IterableConstants.SHARED_PREFS_FCM_MIGRATION_DONE_KEY, true).apply();
                             }
-                        }, null);
+                        }, null, priorAuthToken);
```
vbabenkoru, line 104: We're using `iterablePushRegistrationData` to get the email/userId. Passing `priorAuthToken` should not be necessary since it's also available in `iterablePushRegistrationData`.

vbabenkoru (changes requested):

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -276,54 +296,21 @@ public static void initialize(@NonNull Context context, @NonNull String apiKey,
...
-    /**
-     * Set user email used for API calls
-     * Calling this or {@link #setUserId(String)} is required before making any API calls.
-     *
-     * Note: This clears userId and persists the user email so you only need to call this once when the user logs in.
-     * @param email User email
-     * @param authToken Authorization token
-     */
-    void setEmail(@Nullable String email, @Nullable String authToken) {
-        if (_email != null && _email.equals(email)) {
```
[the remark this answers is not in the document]
davidtruong: I removed it so that we would require a single place to fetch the auth token for maintainability. However thinking about it more, having this in there should be fine since it achieves the same behavior and we call will call the same setAuthToken function which should handle all of the expirations.

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -205,6 +219,12 @@ void setNotificationData(IterableNotificationData data) {
         }
     }
 
+    void onSetAuthToken(String authToken) {
+        _authToken = authToken;
```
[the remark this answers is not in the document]
davidtruong: We should be able to do a check for authToken and not set it or call storeAuthData. However we will still need to call onLogIn.

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -332,34 +319,23 @@ public void setUserId(@Nullable String userId) {
...
         _userId = userId;
-        _authToken = authToken;
         storeAuthData();
-        onLogIn();
+
+        if (userId != null) {
+            getAuthManager().requestNewAuthToken(false, new IterableHelper.SuccessAuthHandler() {
+                @Override
+                public void onSuccess(@NonNull String authToken) {
+                    onSetAuthToken(authToken);
```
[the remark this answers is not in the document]
davidtruong: Makes sense, this is unnecessary now since the implementation of requestNewAuthToken is now asynchronous.

iterableapi/src/main/java/com/iterable/iterableapi/IterablePushRegistration.java
davidtruong: Ah yes, I missed that, Good catch.

davidtruong:
> @davidtruong It seems like it will request a new token multiple times in parallel if multiple network calls fail in sequence. I think it should only request once and wait until it completes before spawning new requests.

Actually It won't request multiple tokens in parallel. The logic in IterableAuthManager.requestNewAuthToken takes in a hasFailedPriorAuth flag. If this is set from an earlier failed Iterable api call w/ an auth error, then it won't attempt to make a second request to onAuthTokenRequested. 

EX: 2 api requests to the backend in parallel (A, B)
A returns a 401 auth
A triggers requestNewAuthToken with hasFailedPriorAuth = true
A triggers onAuthTokenRequested
B returns a 401 auth
B triggers requestNewAuthToken with hasFailedPriorAuth = true
At this point since the AuthManager's hasFailedPriorAuth is also true, B will not trigger a new onAuthTokenRequested

LMK if you see any issues with this or if you were thinking about a different use case.

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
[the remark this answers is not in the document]
davidtruong: What do you think would be the benefit of having this call signature? When a new token requested from our SDK via `onAuthTokenRequested`, they would need to implement that handler as well.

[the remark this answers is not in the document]
vbabenkoru: Oh, I meant the checks that'd skip if the email is the same as it is, etc. Without those, it triggers token refresh and login code every time `setEmail` is called, even if the email didn't change.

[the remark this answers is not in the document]
vbabenkoru: The current version calls `onLogOut` even if the email/userId hasn't changed, triggering `disableDevice` call.

iterableapi/src/main/java/com/iterable/iterableapi/IterableAuthManager.java
```diff
@@ -0,0 +1,111 @@
...
+    private boolean hasFailedPriorAuth;
+
+    IterableAuthManager(IterableApi api, IterableAuthHandler authHandler, long authRefreshPeriod) {
+        timer = new Timer(true);
+        this.api = api;
+        this.authHandler = authHandler;
+        this.authRefreshPeriod = authRefreshPeriod;
+    }
+
+    public void requestNewAuthToken(boolean hasFailedPriorAuth) {
```
vbabenkoru, line 35: @davidtruong It seems like it will request a new token multiple times in parallel if multiple network calls fail in sequence. I think it should only request once and wait until it completes before spawning new requests.

iterableapi/src/main/java/com/iterable/iterableapi/IterableRequest.java
```diff
@@ -187,6 +192,8 @@ protected String doInBackground(IterableApiRequest... params) {
                         handleFailure("No data received", jsonResponse);
                     } else if (error != null) {
                         handleFailure(error, null);
+                    } else {
+                        IterableApi.getInstance().getAuthManager().resetFailedAuth();
```
vbabenkoru, line 196: What is this line for? Looks like it's only going to be executed if `error == null` and `requestResult.length < 0`, which will never be `true`.

iterableapi/src/main/java/com/iterable/iterableapi/IterableAuthManager.java
davidtruong: On receiving a 401 from the backend. the network calls should all call the same authManager which should have the same reference to the hasFailedPriorAuth flag, so it should skip additional calls to onAuthTokenRequested. LMK if I'm not understanding your concern correctly.

iterableapi/src/main/java/com/iterable/iterableapi/IterableRequest.java
davidtruong: Ah, missed the case statement above for the success for error ==null and equestResult.length() > 0.

iterableapi/src/main/java/com/iterable/iterableapi/IterableAuthManager.java
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    private void scheduleAuthTokenRefresh(long timeDuration) {
+        timer.cancel();
+        timer = new Timer(true);
+        try {
+            timer.schedule(new TimerTask() {
+                @Override
+                public void run() {
+                    api.getAuthManager().requestNewAuthToken(false);
```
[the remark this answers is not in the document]
davidtruong: This should be fine since a majority the functionality in `requestNewAuthToken` runs in a background thread as well.

iterableapi/src/main/java/com/iterable/iterableapi/IterableAuthManager.java
vbabenkoru: Hm, yeah, you're right, but I've seen two parallel calls to `requestNewAuthToken`. Refreshes from requests have a limit of 1 at a time, but anything that calls with `hasFailedPriorAuth = false` will result in multiple parallel requests. Since those are asynchronous, there is no guarantee of ordering. A call that was triggered first might return last and vice versa.

davidtruong: Gotcha, that makes sense for the non-401 case for standard initialization: setEmail + registerForPush + getMessages. I'll add back in the pending auth flag to skip to restrict simultaneous the calls.

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -276,30 +298,8 @@ public static void initialize(@NonNull Context context, @NonNull String apiKey,
...
-            }
-
-            if (_authToken != null && _authToken.equals(authToken)) {
-                return;
-            }
-
-            _authToken = authToken;
-            storeAuthData();
-
+            getAuthManager().requestNewAuthToken(false);
```
vbabenkoru, line 302: Do we need to request a new token if the email stays the same?

vbabenkoru: There are very likely some apps that call `setEmail` on every app launch. This will cause it to re-request the token on every launch.

davidtruong: I'm fine with removing this call if the email is the same as the stored _email. However are there any alternatives for re-requesting a new auth token manually from the app without changing the email?

vbabenkoru:
@davidtruong Looks like one of the tests is failing.

davidtruong:
@vbabenkoru Looks like a flaky test. I'll remove it for when we finish fixing the unit tests.

iterableapi/src/main/java/com/iterable/iterableapi/IterableAuthManager.java
```diff
@@ -0,0 +1,132 @@
...
+                        public String call() throws Exception {
+                            return authHandler.onAuthTokenRequested();
+                        }
+                    }).onSuccess(new Future.SuccessCallback<String>() {
+                        @Override
+                        public void onSuccess(String authToken) {
+                            if (authToken != null) {
+                                queueExpirationRefresh(authToken);
+                            }
+                            IterableApi.getInstance().setAuthToken(authToken);
```
Ayyanchira, line 54: Thinking if checking email/userId exists can help before assigning the authToken can avoid situation where 
1. Valid setEmail was called with email xyz.
2. setEmail (null) got called before getting response to previous one.

By the time we get response, email could be set to null and we would have received a auth token.

Saw the test failure and thought could this be a case?

vbabenkoru (approved):

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -205,6 +219,16 @@ void setNotificationData(IterableNotificationData data) {
         }
     }
 
+    void setAuthToken(String authToken) {
+        if (isInitialized()) {
+            if (_authToken != authToken) {
+                _authToken = authToken;
+                storeAuthData();
+            }
+            onLogIn();
```
vbabenkoru, line 228: We should not be calling `onLogIn()` if it's just a token refresh and the email/userId did not change.

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -205,6 +219,16 @@ void setNotificationData(IterableNotificationData data) {
         }
     }
 
+    void setAuthToken(String authToken) {
+        if (isInitialized()) {
+            if (_authToken != authToken) {
```
vbabenkoru, line 224: This is doing object comparison instead of string comparison, this will be `false` for identical strings if they're not the _same object_.

PR #268, opened 2020-10-09
Feature/mob 2054 in app animation poc

Ayyanchira:


iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFragmentHTMLNotification.java
```diff
@@ -40,6 +40,9 @@
     private static final String CALLBACK_ON_CANCEL = "CallbackOnCancel";
     private static final String MESSAGE_ID = "MessageId";
     private static final String INAPP_OPEN_TRACKED = "InAppOpenTracked";
+    private static final String INAPP_BGALPHA = "InAppBgAlpha";
+    private static final String INAPP_BGCOLOR = "InAppBgColor";
+    private static final String INAPP_SHOULD_ANIMATE = "shouldAnimate";
```
vbabenkoru, line 45: Formatting is different in this value (starts with a lowercase letter).

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppMessage.java
```diff
@@ -12,6 +12,11 @@
 
 import java.util.Date;
 
+import static com.iterable.iterableapi.IterableConstants.ITERABLE_IN_APP_BGCOLOR;
```
vbabenkoru, line 15: Use `IterableConstants.*` instead of importing every constant statically.

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppMessage.java
```diff
@@ -142,6 +149,26 @@ public int hashCode() {
         }
     }
 
+    public static class InAppDisplaySetting {
```
vbabenkoru, line 152: The field is called `inAppDisplaySettings` (plural).

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppMessage.java
```diff
@@ -303,11 +330,22 @@ static IterableInAppMessage fromJSONObject(@NonNull JSONObject messageJson, @Nul
...
-
-        JSONObject paddingOptions = contentJson.optJSONObject(IterableConstants.ITERABLE_IN_APP_DISPLAY_SETTINGS);
-        Rect padding = getPaddingFromPayload(paddingOptions);
+        JSONObject inAppDisplaySettingsJson = contentJson.optJSONObject(IterableConstants.ITERABLE_IN_APP_DISPLAY_SETTINGS);
+        Rect padding = getPaddingFromPayload(inAppDisplaySettingsJson);
         double backgroundAlpha = contentJson.optDouble(IterableConstants.ITERABLE_IN_APP_BACKGROUND_ALPHA, 0);
+        boolean shouldAnimate = inAppDisplaySettingsJson.optBoolean(ITERABLE_IN_APP_DEFAULT_ANIMATION, true);
+        JSONObject bgColorSetting = inAppDisplaySettingsJson.optJSONObject(ITERABLE_IN_APP_BGCOLOR);
+
+//        String bgColorInHex = "#00D3D3D3";
```
vbabenkoru, line 339: Looks like something temporary.

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppMessage.java
```diff
@@ -303,11 +330,22 @@ static IterableInAppMessage fromJSONObject(@NonNull JSONObject messageJson, @Nul
...
 
         String html = contentJson.optString(IterableConstants.ITERABLE_IN_APP_HTML, null);
-
-        JSONObject paddingOptions = contentJson.optJSONObject(IterableConstants.ITERABLE_IN_APP_DISPLAY_SETTINGS);
-        Rect padding = getPaddingFromPayload(paddingOptions);
+        JSONObject inAppDisplaySettingsJson = contentJson.optJSONObject(IterableConstants.ITERABLE_IN_APP_DISPLAY_SETTINGS);
+        Rect padding = getPaddingFromPayload(inAppDisplaySettingsJson);
         double backgroundAlpha = contentJson.optDouble(IterableConstants.ITERABLE_IN_APP_BACKGROUND_ALPHA, 0);
+        boolean shouldAnimate = inAppDisplaySettingsJson.optBoolean(ITERABLE_IN_APP_DEFAULT_ANIMATION, true);
+        JSONObject bgColorSetting = inAppDisplaySettingsJson.optJSONObject(ITERABLE_IN_APP_BGCOLOR);
```
vbabenkoru, line 337: To match the convention used in this method, I'd name it `bgColorJson`.

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppMessage.java
```diff
@@ -303,11 +330,22 @@ static IterableInAppMessage fromJSONObject(@NonNull JSONObject messageJson, @Nul
...
+//        String bgColorInHex = "#00D3D3D3";
+//        double bgAlpha = 0.3f;
+        String bgColorInHex = null;
+        double bgAlpha = 0.0f;
+        if (bgColorSetting != null) {
+            bgColorInHex = bgColorSetting.optString(ITERABLE_IN_APP_BGCOLOR_HEX);
+            bgAlpha = bgColorSetting.optDouble(ITERABLE_IN_APP_BGCOLOR_ALPHA);
+        }
 
+        InAppDisplaySetting inAppDisplaySetting = new InAppDisplaySetting(shouldAnimate, new InAppBgColor(bgColorInHex, bgAlpha));
```
vbabenkoru, line 348: If there is no `bgColor` in the payload, do we want to create an `InAppBgColor` with null values? It might be better to not create one and just pass `null`.

iterableapi/src/main/java/com/iterable/iterableapi/IterableWebViewClient.java
```diff
@@ -34,12 +45,87 @@ public boolean shouldOverrideUrlLoading(WebView view, String url) {
      */
     @Override
     public void onPageFinished(WebView view, String url) {
+        htmlView = view;
         inAppHTMLNotification.setLoaded(true);
         view.loadUrl(RESIZE_SCRIPT);
     }
 
+    void showWebView() {
```
vbabenkoru, line 53: A `WebViewClient` ideally should not know anything about the view hierarchy.. As I mentioned in an earlier comment, it is probably better to move it up to the `*HTMLNotification` class.

iterableapi/src/main/java/com/iterable/iterableapi/IterableWebViewClient.java
```diff
@@ -34,12 +45,87 @@ public boolean shouldOverrideUrlLoading(WebView view, String url) {
...
+
     interface HTMLNotificationCallbacks {
         void onUrlClicked(String url);
+
         void setLoaded(boolean loaded);
     }
+
 }
+
+enum InAppLayout {
```
vbabenkoru, line 126: 👍
I'd also move this up to the in-app html notification class since that is where we manage the layout.

vbabenkoru (changes requested):
Nice work! Left some comments.
Please add the new payload to `inapp_payload_multiple.json` to make sure we're testing serialization/deserialization of those new fields.

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppMessage.java
```diff
@@ -142,6 +149,26 @@ public int hashCode() {
...
+        InAppBgColor inAppBgColor;
+
+        public InAppDisplaySetting(boolean shouldAnimate, InAppBgColor inAppBgColor) {
+            this.shouldAnimate = shouldAnimate;
+            this.inAppBgColor = inAppBgColor;
+        }
+    }
+
+    public static class InAppBgColor {
+        String bgHexColor;
```
vbabenkoru, line 163: InAppBgColor isn't actually used yet for background color, right?

iterableapi/src/main/res/anim/bottom_exit.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
+<translate xmlns:android="http://schemas.android.com/apk/res/android"
+    android:fromYDelta="0%p"
+    android:toYDelta="100%p"
+    android:duration="500" />
```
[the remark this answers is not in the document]
Ayyanchira: Agreed

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppMessage.java
Ayyanchira: Yes, it is used.. In createInstance of IterableInAppFragmentHTMLNotification.

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppMessage.java
Ayyanchira: Yes.. kept their for test convenience... Easy to uncomment and see in app animation working..
Will remove it too

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFragmentHTMLNotification.java
```diff
@@ -297,27 +333,20 @@ public void run() {
...
-                        //Set the window properties
-                        WindowManager.LayoutParams wlp = window.getAttributes();
-                        wlp.x = offset;
-                        wlp.gravity = getVerticalLocation(insetPadding);
-                        wlp.dimAmount = (float) notification.backgroundAlpha;
-                        wlp.flags = WindowManager.LayoutParams.FLAG_DIM_BEHIND;
-                        window.setAttributes(wlp);
+                        //Set background window fading if animation and backgroundAlpha/opacity configured
+                        if (shouldAnimate && (inAppBackgroundColor != null)) {
+                            window.setBackgroundDrawable(new ColorDrawable(Color.parseColor(inAppBackgroundColor)));
```
vbabenkoru, line 340: We'll get a `IllegalArgumentException` if the color string is invalid here.. Can it possibly be invalid?

Ayyanchira: We would only get hex strings which should be passed from UI.. probably from a color selection..Shouldnt go wrong.. However, we can default it to transparent though if at all error occurs

app/src/androidTest/java/com/iterable/iterableapi/MainActivityTest.java
```diff
@@ -143,6 +143,7 @@ public void checkIfMessageShownAsPopUpForImproperIntent() throws Exception {
         rule.launchActivity(intent);
 
         onView(withText("Tips and tricks 2")).perform(click());
+        waitFor(500);
```
vbabenkoru, line 146: Hmm wondering if it's possible to replace these delays with something else..

iterableapi/src/main/java/com/iterable/iterableapi/IterableConstants.java
```diff
@@ -204,6 +207,8 @@
     public static final String ITERABLE_INBOX_IMPRESSIONS                = "impressions";
     public static final String ITERABLE_INBOX_IMP_DISPLAY_COUNT          = "displayCount";
     public static final String ITERABLE_INBOX_IMP_DISPLAY_DURATION       = "displayDuration";
+    public static final String ITERABLE_IN_APP_SHOULD_ANIMATE            = "shouldAnimate";
+    public static final int ITERABLE_IN_APP_ANIMATION_DURATION           = 500;
```
vbabenkoru, line 211: Do you want to keep the duration here or move to xml? I think either option is valid.

Ayyanchira: Yeah.. lets keep it here..

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFragmentHTMLNotification.java
```diff
@@ -53,8 +61,13 @@
     private boolean callbackOnCancel = false;
     private String htmlString;
     private String messageId;
+
+    //TODO: To delete this variable in future iterations
```
vbabenkoru, line 65: Is anything preventing us from removing it now? (Not necessarily in this PR)

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFragmentHTMLNotification.java
```diff
@@ -151,6 +155,8 @@ public void onBackPressed() {
         dialog.requestWindowFeature(Window.FEATURE_NO_TITLE);
         if (insetPadding.bottom == 0 && insetPadding.top == 0) {
             dialog.getWindow().setFlags(WindowManager.LayoutParams.FLAG_FULLSCREEN, WindowManager.LayoutParams.FLAG_FULLSCREEN);
+        } else if (getInAppLayout(insetPadding) != InAppLayout.TOP) {
```
vbabenkoru, line 158: If we're converting padding to `InAppLayout` here anyway, I think it's better to also replace the condition 2 lines above that is still using `insetPadding`.

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFragmentHTMLNotification.java
```diff
@@ -151,6 +155,8 @@ public void onBackPressed() {
...
+            dialog.getWindow().setFlags(WindowManager.LayoutParams.FLAG_TRANSLUCENT_STATUS, WindowManager.LayoutParams.FLAG_TRANSLUCENT_STATUS);
```
vbabenkoru, line 159: Why are we only setting this if it's centered or bottom?

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFragmentHTMLNotification.java
Ayyanchira: We were skeptical if backend would ever send that value. But now that we know nothing had been implemented on backend side to send any old backgroundAlpha values, we can delete it right away.. But just to be safe not this time.

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFragmentHTMLNotification.java
```diff
@@ -254,6 +261,108 @@ public void onBackPressed() {
...
+                }
+            }, DELAY_THRESHOLD_MS);
+            loadBackground();
+        } catch (NullPointerException e) {
+            IterableLogger.e(TAG, "View not present. Failed to hide before resizing inapp");
+        }
+    }
+
+    private void loadBackground() {
+        if (shouldAnimate && (inAppBackgroundColor != null)) {
```
vbabenkoru, line 280: Background color is dependent on `shouldAnimate` being enabled. Is that what we want?

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFragmentHTMLNotification.java
```diff
@@ -254,6 +261,108 @@ public void onBackPressed() {
...
+        }
+
+        Runnable dismissWebviewRunnable = new Runnable() {
+            @Override
+            public void run() {
+                IterableInAppFragmentHTMLNotification.super.dismiss();
+            }
+        };
+
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN) {
```
vbabenkoru, line 358: Does this mean we don't animate on Android 4.2 and lower?

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFragmentHTMLNotification.java
Ayyanchira: If we do translucent status for TOP, there is a small overlap of in app content and status bar which we discussed earlier. So to avoid that, TOP in apps will have opaque status bar with content starting right after it

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFragmentHTMLNotification.java
Ayyanchira: I did mentioned it on Slab.. will followup on it urgently

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppMessage.java
```diff
@@ -142,6 +144,26 @@ public int hashCode() {
         }
     }
 
+    public static class InAppDisplaySettings {
```
vbabenkoru, line 147: Hmm, if we store padding values within `inAppDisplaySettings`, maybe it makes sense to move the padding values here as well? (Not necessarily within this PR)

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFragmentHTMLNotification.java
```diff
@@ -254,6 +261,108 @@ public void onBackPressed() {
...
+            webView.postOnAnimationDelayed(dismissWebviewRunnable, 400);
+        } else {
+            dismissWebviewRunnable.run();
```
vbabenkoru, line 361:
```suggestion
            webView.postDelayed(dismissWebviewRunnable, 400);
```

iterableapi/src/test/java/com/iterable/iterableapi/IterableInAppMessageTest.java
```diff
@@ -37,11 +37,19 @@ public void testInAppMessageDeserialization() throws Exception {
                 //Stripping out HTML content from the copy
                 JSONObject messageJsonHTMLStripped = new JSONObject(messageJson.toString());
                 JSONObject content = (JSONObject) messageJsonHTMLStripped.get("content");
+                JSONObject inAppDisplaySettings = content.getJSONObject("inAppDisplaySettings");
+                assertEquals((Boolean) inAppDisplaySettings.get("shouldAnimate"), true);
```
vbabenkoru, line 41: `assertTrue`?

vbabenkoru: Actually, you don't need any new code in this file.

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFragmentHTMLNotification.java
vbabenkoru: I think it'd be helpful to add a comment here explaining this, so that next time someone looks at it, there is context for why it's done this way.

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFragmentHTMLNotification.java
vbabenkoru: Sounds good!

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppMessage.java
Ayyanchira: I think html and padding can go together at similar level and we can reserve inAppDisplaySettings for higher functionalities.. (May be custom animation parameters in future).

app/src/androidTest/java/com/iterable/iterableapi/MainActivityTest.java
Ayyanchira: Other best alternative I found is waitFor(300) 

😁
Jokes apart.. 
This looked like a different case than our regular threading issues where we could flush background Threads.. 

May be I can looking optimizing it in tomorrow's test day

iterableapi/src/main/java/com/iterable/iterableapi/IterableWebViewClient.java
```diff
@@ -24,6 +25,7 @@
     @Override
     public boolean shouldOverrideUrlLoading(WebView view, String url) {
         inAppHTMLNotification.onUrlClicked(url);
+        view.setVisibility(View.GONE);
```
vbabenkoru, line 28: Why do we set it to `GONE` here?

Ayyanchira: Wow.. thank you so much! I dont think we need that code here. Totally forgot to remove this..

iterableapi/src/main/java/com/iterable/iterableapi/IterableWebView.java
```diff
@@ -24,7 +25,7 @@
      * @param html
      */
     void createWithHtml(IterableWebViewClient.HTMLNotificationCallbacks notificationDialog, String html) {
-        IterableWebViewClient webViewClient = new IterableWebViewClient(notificationDialog);
+        webViewClient = new IterableWebViewClient(notificationDialog);
```
vbabenkoru, line 28: You can also remove this, since we don't use `webViewClient` anymore.

Ayyanchira: Missed out this one too... Thanks again

vbabenkoru (approved):

Ayyanchira:
🤩

PR #270, opened 2020-10-13
Fixed the check for string equality by DT

Ayyanchira:
Doesn't call login logic if the auth token has not changed

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
@@ -220,10 +220,16 @@ private static final String TAG = "IterableApi";
     }
 
     void setAuthToken(String authToken) {
+        setAuthToken(authToken, false);
+    }
+
+    void setAuthToken(String authToken, boolean bypassAuth) {
         if (isInitialized()) {
             if ((authToken != null && !authToken.equalsIgnoreCase(_authToken)) || (_authToken != null && !_authToken.equalsIgnoreCase(authToken))) {
                 _authToken = authToken;
                 storeAuthData();
+                onLogIn();
+            } else if (bypassAuth) {
                 onLogIn();
             }
         }
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
@@ -221,11 +221,11 @@ private static final String TAG = "IterableApi";
 
     void setAuthToken(String authToken) {
         if (isInitialized()) {
-            if (_authToken != authToken) {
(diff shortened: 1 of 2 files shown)


iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -221,11 +221,11 @@ void setNotificationData(IterableNotificationData data) {
 
     void setAuthToken(String authToken) {
         if (isInitialized()) {
-            if (_authToken != authToken) {
+            if ((authToken != null && !authToken.equalsIgnoreCase(_authToken)) || (_authToken != null && !_authToken.equalsIgnoreCase(authToken))) {
```
vbabenkoru, line 224: I think we can just leave the first part of the check (`authToken != null && !authToken.equalsIgnoreCase(_authToken)`).
It will return `true` if the new token is not null and is different from the current one. The second condition enables one case: when the new token is `null` and the old token is not null. I'm not sure when we could encounter that and not sure if we want to run login in that case.

vbabenkoru (commented):
Left a comment on the comparison logic.
I think one thing that we still need to think about is whether it's ok that we run the login sequence every time an auth token changes.

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
davidtruong: Don't we need the second case for the logout case?
When setting the email to null, shouldn't we also be removing the authToken.

vbabenkoru: If email is null, `isInitialized` will be false and the inner conditional won't be executed.

vbabenkoru (approved):
Approved for the comparison fix.
We've decided to defer the login sequence on token change to the next release.

Ayyanchira:
Yeah...

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
Ayyanchira: Does this flow allow us to set authToken to null?

Ayyanchira: It does...

vbabenkoru: @Ayyanchira Hmm you're right, it also prevents us from setting it to `null`.

vbabenkoru: Oh yeah, you're right. If current token is not null, we can set it to null. If it's already null, we don't do anything.

Ayyanchira: But then we will never be able to set it to null...

PR #276, opened 2020-10-30
[MOB-2123] Handle extras == null in IterablePushActionReceiver

vbabenkoru:
This doesn't really happen on real devices, only when bots try to hack the apps with the SDK. But it creates noise in crash reports, so it's worth fixing.


davidtruong (approved):
LGTM

PR #278, opened 2020-11-04
[MOB-2270] Support delayed initialization

vbabenkoru:
Currently, we require that `IterableApi.initialize` is always called in `Application#onCreate`. However, there is no easy way to do that in React Native. This adds support for late initialization of the SDK with two changes:
* Push actions: If SDK is not initialized by the time we receive a push action, it is stored in a static variable and executed after SDK init
* Silent push messages are skipped if (we're going to refresh the in-app queue on launch anyway, so skipping these should be fine)


Ayyanchira (approved):
👍

PR #279, opened 2020-11-05
[MOB-2186] Extract ApiClient

vbabenkoru:


iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -62,13 +59,20 @@
 
 //region Constructor
 //---------------------------------------------------------------------------------------
-    IterableApi() {
+IterableApi() {
+    config = new IterableConfig.Builder().build();
+}
```
Ayyanchira, line 64: indentation

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -474,23 +481,7 @@ public void track(@NonNull String eventName, int campaignId, int templateId, @Nu
             return;
```
Ayyanchira, line 481: Can `checkSDKInitialization` be also refactored somehow?
Its been called at start of many Api calls...

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -296,7 +301,9 @@ public static void initialize(@NonNull Context context, @NonNull String apiKey,
         sharedInstance.checkForDeferredDeeplink();
         IterableActivityMonitor.getInstance().registerLifecycleCallbacks(context);
         IterableActivityMonitor.getInstance().addCallback(sharedInstance.activityMonitorListener);
-        sharedInstance.inAppManager = new IterableInAppManager(sharedInstance, sharedInstance.config.inAppHandler, sharedInstance.config.inAppDisplayInterval);
+        if (sharedInstance.inAppManager == null) {
```
Ayyanchira, line 304: Wondering what would be a good case where `IterableApi` itself is initializing and `IterableInAppManager` wont be null? 🤔

Ayyanchira:
Thats 400 lines gone from `IterableAPI` 👌👌👌

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
vbabenkoru: There is a special constructor for tests where you can inject an `inAppManager` instance.

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
vbabenkoru: Hmm, I don't think so... short of doing OOP shenanigans with an interface and 'initialized' and 'non-initialized' implementations.

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
vbabenkoru: Huh.. looks like an IntelliJ/Android Studio formatter bug.. It forced this indentation until I added an empty line above.

Ayyanchira (approved):

PR #280, opened 2020-11-06
[MOB-2328] Update to Robolectric 4.4

vbabenkoru:


Ayyanchira:
Looks like we need to call idling methods at more places where tests are waiting for methods to get invoked as reported in travis.

vbabenkoru:
@Ayyanchira Yeah.. It ran successfully on my machine, but I guess Travis has different concurrency and it picked up things that were missed in local test runs.

vbabenkoru:
@Ayyanchira Fixed!

Ayyanchira:
Wow.. 👌👌👌
Looks like this will also help in jwt auth testing...

Ayyanchira (approved):

PR #284, opened 2020-11-12
[MOB-2046] - Setting up database

Ayyanchira:
1. Created Datamanager
2. Created DatabaseManager
3. Created IterableTask class acting as dao for offline events


iterableapi/src/main/java/com/iterable/iterableapi/IterableDataManager.java
```diff
@@ -0,0 +1,176 @@
...
+            sharedInstance = new IterableDataManager();
+        }
+        return sharedInstance;
+    }
+
+    /*
+    * Create and insert a new task in the database
+    *
+    * */
+    void createTask(String name) {
```
vbabenkoru, line 67: I think this class has various levels of abstraction. It manages/create a database (low level), but also manages tasks. Tasks seem to be a higher level abstraction that may be better placed in its own class.

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -35,6 +38,7 @@
      * {@link IterableApi} singleton instance
      */
     static volatile IterableApi sharedInstance = new IterableApi();
+    private static IterableTaskManager dataManager;
```
vbabenkoru, line 41: Remove

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -307,6 +311,24 @@ public static void initialize(@NonNull Context context, @NonNull String apiKey,
                     sharedInstance.config.inAppDisplayInterval);
         }
         IterablePushActionReceiver.processPendingAction(context);
+
+        //TODO: Temporary code to test. Datamanger would eventually either go to IterableRequest or apiInternal as architecture concretes.
```
vbabenkoru, line 315: Remove

iterableapi/src/main/java/com/iterable/iterableapi/IterableTask.java
```diff
@@ -0,0 +1,92 @@
...
+        this.id = UUID.randomUUID().toString();
+        this.name = name;
+        this.createdAt = new Date();
+        this.scheduledAt = new Date();
+        this.requestedAt = new Date();
+        this.taskType = taskType;
+
+    }
+
+    IterableTask updateTask(int attempts, Date lastAttemptedAt, Boolean processing, Date scheduledAt, Object data, Object taskFailureData) {
```
vbabenkoru, line 80: How/when is this going to be used?

iterableapi/src/main/java/com/iterable/iterableapi/IterableTaskDatabaseManager.java
```diff
@@ -0,0 +1,25 @@
+package com.iterable.iterableapi;
+
+import android.content.Context;
+import android.database.sqlite.SQLiteDatabase;
+import android.database.sqlite.SQLiteOpenHelper;
+
+class IterableTaskDatabaseManager extends SQLiteOpenHelper {
```
vbabenkoru, line 7: Is this designed just for the tasks? If we need to store something else, are we going to create a separate db or reuse this one?

iterableapi/src/main/java/com/iterable/iterableapi/IterableTaskDatabaseManager.java
```diff
@@ -0,0 +1,25 @@
...
+    private static final String DATABASE_NAME = "iterable_sdk.db";
```
vbabenkoru, line 8: Depending on the answer to above, we probably need to scope the name to tasks only, or rename the class itself.

iterableapi/src/main/java/com/iterable/iterableapi/IterableTaskManager.java
```diff
@@ -0,0 +1,335 @@
...
+    Boolean deleteTask(String id) {
+        if (! precheck()) return false;
+        int numberOfEntriesDeleted = database.delete(ITERABLE_TASK_TABLE_NAME, IterableTask.TASK_ID + " =?", new String[]{id});
+        IterableLogger.v(TAG, "Deleted entry - " + numberOfEntriesDeleted);
+        return true;
+    }
+
+    Boolean updateModifiedAt(String id, Date date) {
+        if (!precheck()) return false;
+        contentValues.put(IterableTask.MODIFIED_AT, date.toString());
```
vbabenkoru, line 239: Why is it using a class field vs. local variable for `contentValues`?

iterableapi/src/main/java/com/iterable/iterableapi/IterableTaskManager.java
```diff
@@ -0,0 +1,335 @@
...
+        if (result > 0) return true;
+        else return false;
+    }
+
+    private boolean precheck() {
+        if (database == null) {
+            IterableLogger.e(TAG, "Database not initialized");
+            return false;
+        }
+        if (getTaskCount() == 0) {
```
vbabenkoru, line 302: Do we need to check for task count before running queries? I think in that case it'll just return nothing, which should be a valid return value.

Ayyanchira: Removing.. unnecessary overhead

iterableapi/src/main/java/com/iterable/iterableapi/IterableTaskManager.java
Ayyanchira: had it referenced from a website previously.. having it local is safe and prevent possible data leak. Thanks. Made them local

iterableapi/src/main/java/com/iterable/iterableapi/IterableTaskDatabaseManager.java
Ayyanchira: Made it generic.. Creation of all the tables can go here in its onCreate

iterableapi/src/main/java/com/iterable/iterableapi/IterableTaskDatabaseManager.java
Ayyanchira: Not changing the Iterable_sdk. Will keep it like this assuming we will need only one db and tables could be added in this class as feature set requiring table grows

iterableapi/src/main/java/com/iterable/iterableapi/IterableTask.java
Ayyanchira: Not gonna be used. Deleting.

iterableapi/src/main/java/com/iterable/iterableapi/IterableTaskManager.java
```diff
@@ -0,0 +1,409 @@
...
+            IterableTask.TYPE + " TEXT," +
+            IterableTask.ATTEMPTS + " INTEGER" + ")";
+    private static final String KEY_ROWID = "rowid";
+
+    private SQLiteDatabase database;
+    private IterableDatabaseManager databaseManager;
+
+    IterableTaskManager() {
+        try {
+            Context context = IterableApi.getInstance().getMainActivityContext();
```
vbabenkoru, line 44: Ideally this should not have an implicit dependency on `IterableApi`.. One alternative is to just pass `Context` around in the constructor and in `sharedInstance`.

iterableapi/src/main/java/com/iterable/iterableapi/IterableTaskManager.java
```diff
@@ -0,0 +1,409 @@
...
+    private static final String TAG = "IterableTaskManager";
+
+    static final String ITERABLE_TASK_TABLE_NAME = "OfflineTask";
+    private static final String REPLACING_STRING = "*#*#*#*";
+    private static final String QUERY_GET_TASK_BY_ID = "select * from OfflineTask where task_id = '" + REPLACING_STRING + "'";
+
+    static final String OFFLINE_TASK_COLUMN_DATA = " (" + IterableTask.TASK_ID + " TEXT PRIMARY KEY," +
+            IterableTask.NAME + " TEXT," +
+            IterableTask.VERSION + " INTEGER," +
+            IterableTask.CREATED_AT + " TEXT," +
```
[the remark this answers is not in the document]
Ayyanchira: No it seems.
https://sqlite.org/datatype3.html#date_and_time_datatype

iterableapi/src/main/java/com/iterable/iterableapi/IterableTaskManager.java
```diff
@@ -0,0 +1,409 @@
...
+        IterableLogger.v(TAG, "Found " + cursor.getColumnCount() + " columns");
+        return taskIds;
+    }
+
+    /**
+     * Gets number of rows in the OfflineTask table.
+     *
+     * @return Number of rows in the event table.
+     */
+    long getTaskCount() {
```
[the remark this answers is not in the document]
Ayyanchira: Just a left over method I thought could be useful. Will remove

iterableapi/src/main/java/com/iterable/iterableapi/IterableTaskManager.java
```diff
@@ -0,0 +1,409 @@
...
+     * @return Whether or not the task was updated
+     */
+    Boolean updateData(String id, String data) {
+        if (!precheck()) return false;
+        ContentValues contentValues = new ContentValues();
+        contentValues.put(IterableTask.DATA, data);
+        return (0 > database.update(ITERABLE_TASK_TABLE_NAME, contentValues, IterableTask.TASK_ID + "=?", new String[]{id}));
+    }
+
+    private boolean precheck() {
```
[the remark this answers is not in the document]
Ayyanchira: `isOpen` and `inTransaction()` can be a good check to have 👍

iterableapi/src/main/java/com/iterable/iterableapi/IterableTaskManager.java
```diff
@@ -0,0 +1,402 @@
...
+        IterableLogger.v(TAG, "Found " + cursor.getColumnCount() + "columns");
+        return task;
+    }
+
+    /**
+     * Gets ids of all the tasks in OfflineTask table
+     *
+     * @return {@link ArrayList} of {@link String} ids for all the tasks in OfflineTask table
+     */
+    ArrayList<String> getAllTaskIds() {
```
vbabenkoru, line 205: `@NonNull`

iterableapi/src/main/java/com/iterable/iterableapi/IterableTaskManager.java
```diff
@@ -0,0 +1,402 @@
...
+        return;
+    }
+
+    /**
+     * Deletes a task from OfflineTask table
+     *
+     * @param id for the task
+     * @return Whether or not the task was deleted
+     */
+    Boolean deleteTask(String id) {
```
vbabenkoru, line 238: Why `Boolean` and not `boolean`?

iterableapi/src/main/java/com/iterable/iterableapi/IterableTaskManager.java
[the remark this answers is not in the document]
vbabenkoru: Looks like it still has the same name. `inTransaction` I think is not a good name as it would imply it's in an SQL transaction, and this code doesn't use transactions.

iterableapi/src/main/java/com/iterable/iterableapi/IterableTaskManager.java
```diff
@@ -0,0 +1,402 @@
...
+        IterableTaskType type = null;
+        int version = 1;
+        int attempts = 0;
+        Date dateCreated = null;
+        Date dateModified = null, datelastAttempted = null, datescheduled = null, dateRequested = null;
+        Boolean processing = false, failed = false, blocking = false;
+        String data = null, error = null;
+
+
+        if (database == null) {
```
vbabenkoru, line 143: Why is it calling a method (`precheck`) elsewhere and using a direct check here?

iterableapi/src/main/java/com/iterable/iterableapi/IterableTaskManager.java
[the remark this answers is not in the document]
vbabenkoru: Basically I don't know what `precheck` does without looking at its code. Ideally the name should tell what it checks for (in high level terms).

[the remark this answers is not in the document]
Ayyanchira: oh I misread the comment made. I thought you were suggesting of having additional checks inside the method `precheck` 🤦‍♂️ .. my bad

iterableapi/src/main/java/com/iterable/iterableapi/IterableTaskManager.java
Ayyanchira: No particular reason to choose Boolean. Changing it to primitive to keep it simple and consistent.

iterableapi/src/main/java/com/iterable/iterableapi/IterableTaskManager.java
```diff
@@ -0,0 +1,394 @@
...
+            blocking = cursor.getInt(cursor.getColumnIndex(IterableTask.BLOCKING)) > 0;
+        }
+        if (!cursor.isNull(cursor.getColumnIndex(IterableTask.DATA))) {
+            data = cursor.getString(cursor.getColumnIndex(IterableTask.DATA));
+        }
+        if (!cursor.isNull(cursor.getColumnIndex(IterableTask.ERROR))) {
+            error = cursor.getString(cursor.getColumnIndex(IterableTask.ERROR));
+        }
+        if (!cursor.isNull(cursor.getColumnIndex(IterableTask.TYPE))) {
+            type = IterableTaskType.valueOf(cursor.getString(cursor.getColumnIndex(IterableTask.TYPE)));
```
vbabenkoru, line 176: `type` is not used anywhere. Is it needed?

iterableapi/src/main/java/com/iterable/iterableapi/IterableTaskManager.java
```diff
@@ -0,0 +1,394 @@
...
+        try {
+            if (context == null) {
+                return;
+            }
+
+            if (databaseManager == null) {
+                databaseManager = new IterableDatabaseManager(IterableApi.getInstance().getMainActivityContext());
+            }
+            database = databaseManager.getWritableDatabase();
+        } catch (Throwable t) {
```
vbabenkoru, line 54: Catch Exception instead of Throwable
See CodeClimate: https://codeclimate.com/github/Iterable/iterable-android-sdk/pull/284#

vbabenkoru (approved):

PR #287, opened 2020-11-25
[MOB-2002] - Request Processor - Online

Ayyanchira:
1. Introduced RequestProcessor interface
2. Implemented OnlineRequestProcessor which now creates IterableRequest and executes it
3. emails, userIds and authToken is now mainly fetched from authProvider from apiClient
4. changed/removed method signature for internal methods which had authToken or passed emailId or userId around
5  Changed IterablePushRegistration classes to accomodate authHandler changes
6. Changes in test method to make it pass.

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
@@ -429,14 +429,14 @@ private static final String TAG = "IterableApi";
      * @param deviceToken Push token obtained from GCM or FCM
      */
     public void registerDeviceToken(@NonNull String deviceToken) {
-        registerDeviceToken(getPushIntegrationName(), deviceToken, deviceAttributes);
+        registerDeviceToken(_email, _userId, _authToken, getPushIntegrationName(), deviceToken, deviceAttributes);
     }
 
-    protected void registerDeviceToken(final @NonNull String applicationName, final @NonNull String deviceToken, final HashMap<String, String> deviceAttributes) {
+    protected void registerDeviceToken(final @Nullable String email, final @Nullable String userId, final @Nullable String authToken, final @NonNull String applicationName, final @NonNull String deviceToken, final HashMap<String, String> deviceAttributes) {
         if (deviceToken != null) {
             final Thread registrationThread = new Thread(new Runnable() {
                 public void run() {
-                    registerDeviceToken(applicationName, deviceToken, null, deviceAttributes);
+                    registerDeviceToken(email, userId, authToken, applicationName, deviceToken, null, deviceAttributes);
                 }
             });
             registrationThread.start();
@@ -592,7 +592,7 @@ private static final String TAG = "IterableApi";
             return;
         }
 
-        IterablePushRegistrationData data = new IterablePushRegistrationData(getPushIntegrationName(), IterablePushRegistrationData.PushRegistrationAction.ENABLE);
+        IterablePushRegistrationData data = new IterablePushRegistrationData(_email, _userId, _authToken, getPushIntegrationName(), IterablePushRegistrationData.PushRegistrationAction.ENABLE);
         IterablePushRegistration.executePushRegistrationTask(data);
     }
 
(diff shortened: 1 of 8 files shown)


iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -429,14 +429,14 @@ public boolean isIterableIntent(@Nullable Intent intent) {
      * @param deviceToken Push token obtained from GCM or FCM
      */
     public void registerDeviceToken(@NonNull String deviceToken) {
-        registerDeviceToken(_email, _userId, _authToken, getPushIntegrationName(), deviceToken, deviceAttributes);
+        registerDeviceToken(getPushIntegrationName(), deviceToken, deviceAttributes);
     }
 
-    protected void registerDeviceToken(final @Nullable String email, final @Nullable String userId, final @Nullable String authToken, final @NonNull String applicationName, final @NonNull String deviceToken, final HashMap<String, String> deviceAttributes) {
```
vbabenkoru, line 435: Why did you remove email/userId/authToken?

iterableapi/src/main/java/com/iterable/iterableapi/IterableApiClient.java
```diff
@@ -473,20 +473,12 @@ void tryAddArrayToJSON(JSONObject requestJSON, String key, Object[] value) {
...
     }
 
     void sendPostRequest(@NonNull String resourcePath, @NonNull JSONObject json, @Nullable IterableHelper.SuccessHandler onSuccess, @Nullable IterableHelper.FailureHandler onFailure) {
-        sendPostRequest(resourcePath, json, authProvider.getAuthToken(), onSuccess, onFailure);
-    }
-
-    void sendPostRequest(@NonNull String resourcePath, @NonNull JSONObject json, @Nullable String authToken, @Nullable IterableHelper.SuccessHandler onSuccess, @Nullable IterableHelper.FailureHandler onFailure) {
-        IterableApiRequest request = new IterableApiRequest(authProvider.getApiKey(), resourcePath, json, IterableApiRequest.POST, authToken, onSuccess, onFailure);
-        new IterableRequest().execute(request);
+        RequestProcessor requestProcessor = new OnlineRequestProcessor();
```
vbabenkoru, line 480: Do we need to create the request processor every time?

Ayyanchira: For now... Once we take parameters from config whether its online or offline, apiclient can create corresponding global request processor.

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
Ayyanchira: I am thinking, now that we have authProvider, why not make only requestProcessor take email, userID and auth from that interface.
Also noticed that
1. public method only takes token.
2. we already have email, userId set before calling registerForPush

Although not sure why the line 435 method is protected.

vbabenkoru: Hmm.. we should have regression tests for these. Basically, the reason we're passing email, userId and authToken here is because getting a token is asynchronous. By the time we get the token, the user identifiers might've changed.
Example:
* You set email to `null`
* It calls `disableDevice` to disable the token for the logged out user
* If we pass the current email/userId/authToken and then set email to `null`, we'll be able to call `disableDevice` with old auth even though the current one is all null
* If we don't store the auth with the token request, it will get the _current_ auth info from `authProvider`, which is already null, and will fail to disable the device.

vbabenkoru: Added an integration/regression test here: #288.

iterableapi/src/test/java/com/iterable/iterableapi/IterableApiIntegrationTest.java
```diff
@@ -0,0 +1,79 @@
...
+
+import static android.os.Looper.getMainLooper;
+import static junit.framework.Assert.assertEquals;
+import static junit.framework.Assert.assertNotNull;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+import static org.robolectric.Shadows.shadowOf;
+
+public class IterableApiIntegrationTest extends BaseTest {
```
vbabenkoru, line 26: Rebase on latest master and maybe drop the merge commit to drop this change from the PR

iterableapi/src/test/java/com/iterable/iterableapi/IterableApiTest.java
```diff
@@ -50,6 +50,7 @@
     private MockWebServer server;
     private IterableApiClient originalApiClient;
     private IterableApiClient mockApiClient;
+    private IterablePushRegistration.IterablePushRegistrationImpl originalPushRegistrationImpl;
```
vbabenkoru, line 53: same, this should already be in master

vbabenkoru (changes requested):
Looks like this needs cleanup to remove the already merged PR from the diff.

vbabenkoru (approved):

PR #288, opened 2020-11-26
[MOB-2380] Add an integration test for disabling token

vbabenkoru:
There is an asynchronous gotcha in the way we disable tokens: getting a token is an asynchronous operation, and by the time we retrieve the token, email/userId might've changed. So we pass the current email/userId/authToken directly to the token registration/disable action so we use the correct user identifier even if the current one is different.
We didn't have integration tests covering this, and this adds one.


Ayyanchira (approved):

Ayyanchira:
Ahh sorry.. forgot to approve..

PR #295, opened 2020-12-12
[MOB-2372][1] Refactor network requests and task storage

vbabenkoru:
## 🔹 Jira Ticket(s) if any

* [MOB-2372](https://iterable.atlassian.net/browse/MOB-2372)

## ✏️ Description

Part 1 of MOB-2372. Multiple refactors to `IterableRequest` and `IterableTaskManager`. See individual commits for the steps and individual changes.

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableRequestTask.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableRequestTask.java
@@ -23,7 +23,7 @@ import java.util.Iterator;
  * Async task to handle sending data to the Iterable server
  * Created by David Truong dt@iterable.com
  */
-class IterableRequest extends AsyncTask<IterableApiRequest, Void, IterableApiResponse> {
+class IterableRequestTask extends AsyncTask<IterableApiRequest, Void, IterableApiResponse> {
     static final String TAG = "IterableRequest";
     static final String ITERABLE_BASE_URL = "https://api.iterable.com/api/";
 
@@ -250,8 +250,8 @@ class IterableRequest extends AsyncTask<IterableApiRequest, Void, IterableApiRes
         boolean retryRequest = !response.success && response.responseCode >= 500;
 
         if (retryRequest && retryCount <= MAX_RETRY_COUNT) {
-            final IterableRequest request = new IterableRequest();
-            request.setRetryCount(retryCount + 1);
+            final IterableRequestTask requestTask = new IterableRequestTask();
+            requestTask.setRetryCount(retryCount + 1);
 
             long delay = 0;
             if (retryCount > 2) {
@@ -262,7 +262,7 @@ class IterableRequest extends AsyncTask<IterableApiRequest, Void, IterableApiRes
             handler.postDelayed(new Runnable() {
                 @Override
                 public void run() {
-                    request.execute(iterableApiRequest);
+                    requestTask.execute(iterableApiRequest);
                 }
             }, delay);
             return;
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableTaskStorage.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableTaskStorage.java
@@ -230,10 +230,11 @@ class IterableTaskStorage {
 
     /**
      * Returns the next scheduled task for processing
+     *
      * @return next scheduled {@link IterableTask}
      */
     @Nullable
-    IterableTask nextTask() {
+    IterableTask getNextScheduledTask() {
         Cursor cursor = database.rawQuery("select * from OfflineTask order by scheduled limit 1", null);
         IterableTask task = null;
         if (cursor.moveToFirst()) {
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableTaskStorage.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableTaskStorage.java
@@ -12,11 +12,11 @@ import androidx.annotation.Nullable;
 import java.util.ArrayList;
 import java.util.Date;
 
-class IterableTaskManager {
+class IterableTaskStorage {
 
-    private static IterableTaskManager sharedInstance;
+    private static IterableTaskStorage sharedInstance;
 
-    private static final String TAG = "IterableTaskManager";
+    private static final String TAG = "IterableTaskStorage";
 
     static final String ITERABLE_TASK_TABLE_NAME = "OfflineTask";
 
@@ -58,7 +58,7 @@ class IterableTaskManager {
     private SQLiteDatabase database;
     private IterableDatabaseManager databaseManager;
 
-    private IterableTaskManager(Context context) {
+    private IterableTaskStorage(Context context) {
         try {
             if (context == null) {
                 return;
@@ -73,9 +73,9 @@ class IterableTaskManager {
         }
     }
 
-    static IterableTaskManager sharedInstance(Context context) {
+    static IterableTaskStorage sharedInstance(Context context) {
         if (sharedInstance == null) {
-            sharedInstance = new IterableTaskManager(context);
+            sharedInstance = new IterableTaskStorage(context);
         }
         return sharedInstance;
     }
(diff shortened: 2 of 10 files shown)


iterableapi/src/main/java/com/iterable/iterableapi/IterableRequestTask.java
```diff
@@ -264,23 +262,27 @@ protected void onPostExecute(String s) {
...
+        } else if (response.success) {
+            IterableApi.getInstance().getAuthManager().resetFailedAuth();
             if (iterableApiRequest.successCallback != null) {
-                iterableApiRequest.successCallback.onSuccess(requestResultJson);
+                iterableApiRequest.successCallback.onSuccess(response.responseJson);
             }
         } else {
+            if (matchesErrorCode(response.responseJson, ERROR_CODE_INVALID_JWT_PAYLOAD)) {
+                IterableApi.getInstance().getAuthManager().requestNewAuthToken(true);
+            }
```
Ayyanchira, line 277: Hmm not sure if there will be any behavorial change from moving this from `doInBackground` to `onPostExecute`

iterableapi/src/main/java/com/iterable/iterableapi/IterableTaskStorage.java
```diff
@@ -202,11 +228,28 @@ IterableTask getTask(String id) {
...
+    @Nullable
+    IterableTask nextTask() {
+        Cursor cursor = database.rawQuery("select * from OfflineTask order by scheduled limit 1", null);
+        IterableTask task = null;
+        if (cursor.moveToFirst()) {
+            task = createTaskFromCursor(cursor);
+        }
+        cursor.close();
+        return task;
+    }
```
Ayyanchira, line 244: On the first glance, I felt nextTask should have been relative to some task. Or we can document that it returns a task from the top of the taskStorage table.

I felt,
`nextTask` implies that the task it returns is next to something. But it actually is the first task. It assumes that the previous ones should be deleted. May be it could be nextTaskTo(Task task1) which could delete the existing task and then pass the next one as Next Task. 
Either the next task should delete the current one in context and then pass the top one. Or the name should be something different. Like GetOldestTask, GetFirstTask, GetTask. Or we can just change the description of method on top.

iterableapi/src/main/java/com/iterable/iterableapi/OfflineRequestProcessor.java
```diff
@@ -52,13 +52,13 @@ void scheduleTask(IterableApiRequest request, @Nullable IterableHelper.SuccessHa
...
+        IterableTaskStorage taskStorage = IterableTaskStorage.sharedInstance(IterableApi.getInstance().getMainActivityContext());
+        ArrayList<String> taskIds = taskStorage.getAllTaskIds();
         for (String id : taskIds) {
             try {
-                IterableApiRequest request = makeRequestFromTask(taskManager.getTask(id));
-                new IterableRequest().execute(request);
-                taskManager.deleteTask(id);
+                IterableApiRequest request = makeRequestFromTask(taskStorage.getTask(id));
+                new IterableRequestTask().execute(request);
+                taskStorage.deleteTask(id);
```
Ayyanchira, line 61: I think the best place to deleteTask will be when we get task completion callback. This will also mean that when we ask for `nextTask`, we might still get the top one which is not deleted yet. Hence the point of nextTask being relative to another task. 
We can totally defer this for next version if applicable as there is no retries to be done in initial version.

iterableapi/src/main/java/com/iterable/iterableapi/IterableRequestTask.java
```diff
@@ -175,58 +176,64 @@ protected String doInBackground(IterableApiRequest... params) {
...
+    }
+
+    private static void logError(IterableApiRequest iterableApiRequest, String baseUrl, Exception e) {
         IterableLogger.e(TAG, "<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<\n" +
                 "Exception occurred for : " + baseUrl + iterableApiRequest.resourcePath);
         IterableLogger.e(TAG, e.getMessage(), e);
     }
 
-    private String buildHeaderString(HttpURLConnection urlConnection) {
+    private static String buildHeaderString(HttpURLConnection urlConnection) {
```
Ayyanchira, line 236: Not sure but it's safe to have this function static right? 
What happens when multiple requests are fired together? How does the static method behaves in that scenario? Does a static function wait for other tasks to execute?
Will look it up though... 👍

iterableapi/src/main/java/com/iterable/iterableapi/IterableRequestTask.java
```diff
@@ -368,3 +370,27 @@ static IterableApiRequest fromJSON(JSONObject jsonData, @Nullable IterableHelper
         return null;
     }
 }
+
+class IterableApiResponse {
```
Ayyanchira, line 374: 👌

iterableapi/src/main/java/com/iterable/iterableapi/IterableTaskStorage.java
```diff
@@ -119,76 +139,82 @@ IterableTask getTask(String id) {
...
+        }
+
+        IterableTask task = createTaskFromCursor(cursor);
+
+        IterableLogger.v(TAG, "Found " + cursor.getColumnCount() + "columns");
+        cursor.close();
+        return task;
+    }
+
+    private IterableTask createTaskFromCursor(Cursor cursor) {
```
Ayyanchira, line 156: 👌

Ayyanchira (commented):
Left some comments..

iterableapi/src/main/java/com/iterable/iterableapi/IterableRequestTask.java
vbabenkoru: The only functional difference I can think of is that `requestNewAuthToken` will be called on the main thread now. It does call the handler on a background anyway, so it shouldn't matter.
The intent of this change is to separate pure network calls from other logic. The method that makes the request is already long and complicated on its own, and I wanted to extract that logic to use in an offline mode processor, without using the async code provided by `IterableRequestTask`.

iterableapi/src/main/java/com/iterable/iterableapi/IterableTaskStorage.java
vbabenkoru: I was mirroring iOS naming scheme in this case. `nextTask` in this case means 'next task to be processed`, not adjacen or being in some kind of proximity. How exactly it is determined which task should be processed next should ideally be contained in a single class, in this case `IterableTaskStorage`.
Since tasks are not physically located within any kind of space, I think `nextTaskTo` is even less clear.
I can think of 2 solutions:
* Add description to this method
* Call it `getNextScheduledTask()` or something similar

iterableapi/src/main/java/com/iterable/iterableapi/OfflineRequestProcessor.java
vbabenkoru: Yeah, this is addressed in follow-up PRs.

iterableapi/src/main/java/com/iterable/iterableapi/IterableRequestTask.java
vbabenkoru: As long as the function doesn't use any global shared state, it should be fine. I think this method doesn't use any global variables.

vbabenkoru: Basically I wanted to extract all pure networking code from async handling and callbacks here. I guess Ideally it should be extracted to a separate class, but I kept it here for now.

iterableapi/src/main/java/com/iterable/iterableapi/IterableTaskStorage.java
Ayyanchira: Adding description should be good 👍

vbabenkoru: Hm, wait, it already has a description.

vbabenkoru: Renamed to `getNextScheduledTask`.

Ayyanchira (approved):

PR #296, opened 2020-12-12
[MOB-2372][2] Implement IterableTaskRunner

vbabenkoru:
## 🔹 Jira Ticket(s) if any

* [MOB-2372](https://iterable.atlassian.net/browse/MOB-2372)

## ✏️ Description

Part 2 of MOB-2372. Implements `IterableTaskRunner` that doesn't detect network yet, but runs in a background thread and executes all requests directly.
For next steps, needs: 
* callback handling (unless we only send requests that don't need callbacks)
* network detection

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableTaskRunner.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableTaskRunner.java
@@ -1,0 +1,81 @@
+package com.iterable.iterableapi;
+
+import android.os.Handler;
+import android.os.HandlerThread;
+import android.os.Message;
+
+import androidx.annotation.NonNull;
+import androidx.annotation.WorkerThread;
+
+import org.json.JSONObject;
+
+class IterableTaskRunner implements IterableTaskStorage.TaskCreatedListener, Handler.Callback {
+    private static final String TAG = "IterableTaskRunner";
+    private IterableTaskStorage taskStorage;
+    private IterableActivityMonitor activityMonitor;
+
+    private static final int OPERATION_PROCESS_TASKS = 100;
+
+    private final HandlerThread networkThread = new HandlerThread("NetworkThread");
+    private Handler handler;
+
+    IterableTaskRunner(IterableTaskStorage taskStorage, IterableActivityMonitor activityMonitor) {
+        this.taskStorage = taskStorage;
+        this.activityMonitor = activityMonitor;
+
+        networkThread.start();
+        handler = new Handler(networkThread.getLooper(), this);
+        taskStorage.addTaskCreatedListener(this);
+    }
+
+    @Override
+    public void onTaskCreated(IterableTask iterableTask) {
+        runNow();
+    }
+
+    private void runNow() {
+        handler.removeCallbacksAndMessages(this);
+        handler.sendEmptyMessage(OPERATION_PROCESS_TASKS);
+    }
+
+    @WorkerThread
+    @Override
+    public boolean handleMessage(@NonNull Message msg) {
+        if (msg.what == OPERATION_PROCESS_TASKS) {
+            processTasks();
+            return true;
+        }
+        return false;
+    }
+
+    @WorkerThread
+    private void processTasks() {
+        while (true) {
+            boolean proceed = processNextTask();
+            if (!proceed) {
+                break;
+            }
+        }
+    }
+
+    @WorkerThread
+    private boolean processNextTask() {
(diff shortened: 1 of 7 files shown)


iterableapi/src/main/java/com/iterable/iterableapi/IterableTaskRunner.java
```diff
@@ -0,0 +1,81 @@
...
+        }
+
+        if (task.taskType == IterableTaskType.API) {
+            try {
+                IterableApiRequest request = IterableApiRequest.fromJSON(new JSONObject(task.data), null, null);
+                IterableApiResponse response = IterableRequestTask.executeApiRequest(request);
+            } catch (Exception e) {
+                IterableLogger.e(TAG, "Error while processing request task", e);
+            }
+            taskStorage.deleteTask(task.id);
```
Ayyanchira, line 76: Yeah.. this is the delete I was initially talking about. 
Should the delete operation happen here? Or should we move it to taskCompletion handler?

iterableapi/src/main/java/com/iterable/iterableapi/OfflineRequestProcessor.java
```diff
@@ -1,38 +1,50 @@
...
+        taskScheduler = new TaskScheduler(taskStorage);
+        taskRunner = new IterableTaskRunner(taskStorage, IterableActivityMonitor.getInstance());
+    }
 
     @Override
     public void processGetRequest(@Nullable String apiKey, @NonNull String resourcePath, @NonNull JSONObject json, String authToken, @Nullable IterableHelper.IterableActionHandler onCallback) {
-        //Invoke taskScheduler and add task to database
+        // Call GET requests directly, without using the queue
+        IterableApiRequest request = new IterableApiRequest(apiKey, resourcePath, json, IterableApiRequest.GET, authToken, onCallback);
+        new IterableRequestTask().execute(request);
```
Ayyanchira, line 27: 👍

Ayyanchira (commented):
Really amazing revamp! 👌

iterableapi/src/main/java/com/iterable/iterableapi/IterableTaskRunner.java
vbabenkoru: Since the Runner is getting the next task immediately afterwards, if we move deletion farther away from it, it'd be heavily dependent on behavior of another class (if we remove or break `deleteTask` in another class, it'll make `IterableTaskRunner` get stuck in an infinite loop.

iterableapi/src/main/java/com/iterable/iterableapi/IterableTaskRunner.java
```diff
@@ -0,0 +1,81 @@
...
+        handler = new Handler(networkThread.getLooper(), this);
+        taskStorage.addTaskCreatedListener(this);
+    }
+
+    @Override
+    public void onTaskCreated(IterableTask iterableTask) {
+        runNow();
+    }
+
+    private void runNow() {
```
Ayyanchira, line 36: May be we can also think of some name for this method too...
notifyTaskListener/ notifyHandler/ handleCallBackAndMessage

iterableapi/src/main/java/com/iterable/iterableapi/IterableTaskRunner.java
```diff
@@ -0,0 +1,81 @@
...
+        if (msg.what == OPERATION_PROCESS_TASKS) {
+            processTasks();
+            return true;
+        }
+        return false;
+    }
+
+    @WorkerThread
+    private void processTasks() {
+        while (true) {
```
Ayyanchira, line 53: Thinking of a condition here... If in some case, our getNextScheduledTask fails to create a Task (because of improper data) and returns null, other entries in DB wont get a chance to run at all and we might not delete the one causing problem.

Ayyanchira (approved):

iterableapi/src/main/java/com/iterable/iterableapi/IterableTaskRunner.java
vbabenkoru: This is not what it's doing though

iterableapi/src/main/java/com/iterable/iterableapi/IterableTaskRunner.java
vbabenkoru: Good idea. I think I'll handle this in a follow-up PR, since we'll have to deal with this when implementing network error handling anyway.

PR #298, opened 2020-12-16
[MOB-2411] - Offline flag in config

Ayyanchira:
## 🔹 Jira Ticket(s) if any

https://iterable.atlassian.net/browse/MOB-2411

## ✏️ Description

1.  Added offlineProcessing boolean flag in config defaulted to false.
2. Added setRequestProcessor method in apiClient for IterableApi to be able to set offline mode after/during initializaiton.
3. In IterableAPI, if the flag is true, setRequestProcessor to offline.

Consequence: Had to make apiClient static and its required authProvider static as well.

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
@@ -306,6 +306,7 @@ private static final String TAG = "IterableApi";
             sharedInstance.inAppManager = new IterableInAppManager(sharedInstance, sharedInstance.config.inAppHandler,
                     sharedInstance.config.inAppDisplayInterval);
         }
+        sharedInstance.apiClient.setOfflineProcessingEnabled(sharedInstance.config.offlineProcessing);
         IterablePushActionReceiver.processPendingAction(context);
     }
 
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableApiClient.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableApiClient.java
@@ -46,6 +46,14 @@ class IterableApiClient {
         return requestProcessor;
     }
 
+    void setOfflineProcessingEnabled(boolean offlineMode) {
+        if (offlineMode) {
+            this.requestProcessor = new OfflineRequestProcessor(authProvider.getContext());
+        } else {
+            this.requestProcessor = new OnlineRequestProcessor();
+        }
+    }
+
     public void track(@NonNull String eventName, int campaignId, int templateId, @Nullable JSONObject dataFields) {
         JSONObject requestJSON = new JSONObject();
         try {
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableConfig.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableConfig.java
@@ -66,6 +66,11 @@ public class IterableConfig {
     final IterableAuthHandler authHandler;
 
     /**
+     * When set to true, the SDK will keep track of events when the device and trigger them once the network establishes
+     */
+    boolean offlineProcessing;
+
+    /**
      * Duration prior to an auth expiration that a new auth token should be requested.
      */
     final long expiringAuthTokenRefreshPeriod;
@@ -82,6 +87,7 @@ public class IterableConfig {
         inAppDisplayInterval = builder.inAppDisplayInterval;
         authHandler = builder.authHandler;
         expiringAuthTokenRefreshPeriod = builder.expiringAuthTokenRefreshPeriod;
+        offlineProcessing = builder.offlineProcessing;
     }
 
     public static class Builder {
@@ -96,7 +102,7 @@ public class IterableConfig {
         private double inAppDisplayInterval = 30.0;
         private IterableAuthHandler authHandler;
         private long expiringAuthTokenRefreshPeriod = 60000L;
-
+        private boolean offlineProcessing = false;
         public Builder() {}
 
         /**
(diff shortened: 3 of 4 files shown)


Ayyanchira:
@vbabenkoru , the test `testSdkInitializedWithoutEmailOrUserId` fails now.
When run standalone, it succeeds.. 
Not understanding how the `apiclient` mock is getting invoked/interacted if email is set to null at the beginning.. Logically, it checks if sdkinitialrized -> email is null -> return false -> and nothing should have happened. But somehow, I dont know how, making apiClient static might have caused the problem. 🤔

vbabenkoru:
@Ayyanchira I think we should not make it static if there is any other way.

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -306,6 +306,9 @@ public static void initialize(@NonNull Context context, @NonNull String apiKey,
             sharedInstance.inAppManager = new IterableInAppManager(sharedInstance, sharedInstance.config.inAppHandler,
                     sharedInstance.config.inAppDisplayInterval);
         }
+        if(sharedInstance.config.offlineProcessing) {
+            apiClient.setRequestProcessor(true, context);
```
vbabenkoru, line 310: I think this is confusing.. `setRequestProcessor(true, context)` - for a setter, I'd expect it to take some kind of a `RequestProcessor` class.. What `true` means in this case isn't clear.

iterableapi/src/main/java/com/iterable/iterableapi/IterableConfig.java
```diff
@@ -218,6 +224,16 @@ public Builder setExpiringAuthTokenRefreshPeriod(@NonNull Long period) {
             return this;
         }
 
+        /**
+         * When set to true, the SDK will capture events when the device goes offline.
+         * @param offlineProcessing Auth handler provided by the app
+         */
+        @NonNull
+        public Builder setOfflineProcessing(@NonNull boolean offlineProcessing) {
```
vbabenkoru, line 232: This should be package-private for now, since it's not ready for production use yet.

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -48,7 +48,7 @@
     private String _deviceId;
     private boolean _firstForegroundHandled;
 
-    IterableApiClient apiClient = new IterableApiClient(new IterableApiAuthProvider());
+    static IterableApiClient apiClient = new IterableApiClient(new IterableApiAuthProvider());
```
vbabenkoru, line 51: Do we need to make this static?

Ayyanchira: Couldn't figure out a way around! 😕
Because `initialize` was a good place to setRequestProcessor; and because initialize was static, it was asking to make apiClient as static as well.

iterableapi/src/main/java/com/iterable/iterableapi/IterableApiClient.java
```diff
@@ -46,6 +46,13 @@ private RequestProcessor getRequestProcessor() {
         return requestProcessor;
     }
 
+    void enableOfflineProcessing(boolean offlineMode) {
+        if (offlineMode) {
+            this.requestProcessor = new OfflineRequestProcessor(IterableApi.getInstance().getMainActivityContext());
```
vbabenkoru, line 51:
```suggestion
            this.requestProcessor = new OfflineRequestProcessor(authProvider.getContext());
```

iterableapi/src/main/java/com/iterable/iterableapi/IterableApiClient.java
```diff
@@ -46,6 +46,13 @@ private RequestProcessor getRequestProcessor() {
...
+        }
+        //TODO: if api has to reinitialize, say after logout -> login, should apiClient also reset? In that case, there should be an else here. Or ApiClient should be reset on logout or some place
```
vbabenkoru, line 53: We don't need to reset apiClient on logout

iterableapi/src/main/java/com/iterable/iterableapi/IterableApiClient.java
```diff
@@ -46,6 +46,13 @@ private RequestProcessor getRequestProcessor() {
         return requestProcessor;
     }
 
+    void enableOfflineProcessing(boolean offlineMode) {
```
vbabenkoru, line 49: `setOfflineProcessingEnabled`?

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -306,6 +306,9 @@ public static void initialize(@NonNull Context context, @NonNull String apiKey,
             sharedInstance.inAppManager = new IterableInAppManager(sharedInstance, sharedInstance.config.inAppHandler,
                     sharedInstance.config.inAppDisplayInterval);
         }
+        if (sharedInstance.config.offlineProcessing) {
```
vbabenkoru, line 309: The condition check is both here and inside `enableOfflineProcessing`. I think either the method should not have parameters, or we don't need the check here.

vbabenkoru (changes requested):

iterableapi/src/test/java/com/iterable/iterableapi/IterableApiTest.java
```diff
@@ -81,9 +82,10 @@ private void reInitIterableApi() {
 
     @Test
     public void testSdkInitializedWithoutEmailOrUserId() throws Exception {
-        IterableApi.initialize(getContext(), "apiKey");
-        IterableApi.getInstance().setEmail(null);
+        IterableApi.initialize(getContext(), "testApiKey");
+        verify(mockApiClient, times(1)).setOfflineProcessingEnabled(anyBoolean());
```
vbabenkoru, line 86: The test now checks for two things at the same time: that `setOfflineProcessingEnabled` was called and that no API requests are being made. I think ideally it shouldn't be explicitly checking for `setOfflineProcessingEnabled`.

Ayyanchira: That way we can be sure that it was only the setOfflineProcessingEnabled that was called. And not give free pass to other methods that could potentially get called if anything has to go wrong.

Also verify(mockApiClient).neededAMethod

Ayyanchira: Using `clearInvocation` 👍

vbabenkoru (approved):

PR #300, opened 2020-12-18
[MOB-2463] call JWT request on same user being set to SDK

roninopf:
## 🔹 Jira Ticket(s) if any

* [MOB-2463](https://iterable.atlassian.net/browse/MOB-2463)

## ✏️ Description

Makes a request to get a JWT if it's the "same" user being set to the SDK. That way, if a user that is already set to the SDK becomes JWT enabled, or updates to a new app version with JWT enabled, it will proceed to get the JWT, and avoid failing requests.

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
@@ -318,6 +318,7 @@ private static final String TAG = "IterableApi";
      */
     public void setEmail(@Nullable String email) {
         if (_email != null && _email.equals(email)) {
+            getAuthManager().requestNewAuthToken(false);
             return;
         }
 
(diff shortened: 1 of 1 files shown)


vbabenkoru:
I checked the original bug report, and it seems like this change doesn't fully address it: we don't require apps to call `setEmail/setUserId` on every app launch, and refreshing the token on every launch might be unnecessary. If an app doesn't call `setEmail`/`setUserId` on launch, this code won't prevent it from triggering the invalid state (assuming the bug exists on Android). Looks like the original issue was that the SDK wasn't trying to get a new token when requests were failing due to missing JWT. If that was working, it would get a new token and proceed, regardless of whether `setEmail`/`setUserId` is called or not.

vbabenkoru (changes requested):
Comment above

PR #303, opened 2020-12-22
[MOB-2285] - Add OS version to getMessagesAPI

Ayyanchira:
## 🔹 Jira Ticket(s) if any

https://iterable.atlassian.net/browse/MOB-2285

## ✏️ Description

1. Added systemVersion to getMessageApi json
2. Added test assertion to check if systemVersion is passed


iterableapi/src/main/java/com/iterable/iterableapi/IterableApiClient.java
```diff
@@ -164,6 +164,7 @@ public void getInAppMessages(int count, @NonNull IterableHelper.IterableActionHa
             requestJSON.put(IterableConstants.ITERABLE_IN_APP_COUNT, count);
             requestJSON.put(IterableConstants.KEY_PLATFORM, IterableConstants.ITBL_PLATFORM_ANDROID);
             requestJSON.put(IterableConstants.ITBL_KEY_SDK_VERSION, IterableConstants.ITBL_KEY_SDK_VERSION_NUMBER);
+            requestJSON.put(IterableConstants.DEVICE_SYSTEM_VERSION, Build.VERSION.RELEASE);
```
vbabenkoru, line 167: Even though `DEVICE_SYSTEM_VERSION` has the same value, `systemVersion`, it is prefixed with `DEVICE_`, which means it is a property of a device record sent to `registerDeviceToken`. I think in this case it may be better to add another constant.

vbabenkoru (approved):

PR #307, opened 2021-01-14
[MOB - 2417] - CreatedAt and SentAt in requests

Ayyanchira:
## 🔹 Jira Ticket(s) if any

https://iterable.atlassian.net/browse/MOB-2417

## ✏️ Description

1. Changed date type from Date to long to store unix epoch time in milliseconds
2. Added sentAt dates in headers
3. Added check in test method to see if sentAt is sent in headers for all the requests
4. Added method in taskRunner to add createdAt in json data when making request back from database.

Pending -> Test method to see if createdAt is present in offline requests.

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableRequestTask.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableRequestTask.java
@@ -18,6 +18,7 @@ import java.io.OutputStream;
 import java.io.OutputStreamWriter;
 import java.net.HttpURLConnection;
 import java.net.URL;
+import java.util.Date;
 import java.util.Iterator;
 
 /**
@@ -87,6 +88,7 @@ class IterableRequestTask extends AsyncTask<IterableApiRequest, Void, IterableAp
                     urlConnection.setRequestProperty(IterableConstants.HEADER_API_KEY, iterableApiRequest.apiKey);
                     urlConnection.setRequestProperty(IterableConstants.HEADER_SDK_PLATFORM, "Android");
                     urlConnection.setRequestProperty(IterableConstants.HEADER_SDK_VERSION, IterableConstants.ITBL_KEY_SDK_VERSION_NUMBER);
+                    urlConnection.setRequestProperty(IterableConstants.KEY_SENT_AT, String.valueOf(new Date().getTime()));
 
                     if (iterableApiRequest.authToken != null) {
                         urlConnection.setRequestProperty(IterableConstants.HEADER_SDK_AUTHORIZATION, IterableConstants.HEADER_SDK_AUTH_FORMAT + iterableApiRequest.authToken);
@@ -108,6 +110,7 @@ class IterableRequestTask extends AsyncTask<IterableApiRequest, Void, IterableAp
                     urlConnection.setRequestProperty(IterableConstants.HEADER_API_KEY, iterableApiRequest.apiKey);
                     urlConnection.setRequestProperty(IterableConstants.HEADER_SDK_PLATFORM, "Android");
                     urlConnection.setRequestProperty(IterableConstants.HEADER_SDK_VERSION, IterableConstants.ITBL_KEY_SDK_VERSION_NUMBER);
+                    urlConnection.setRequestProperty(IterableConstants.KEY_SENT_AT, String.valueOf(new Date().getTime()));
 
                     if (iterableApiRequest.authToken != null) {
                         urlConnection.setRequestProperty(IterableConstants.HEADER_SDK_AUTHORIZATION, IterableConstants.HEADER_SDK_AUTH_FORMAT + iterableApiRequest.authToken);
@@ -365,6 +368,9 @@ class IterableApiRequest {
                 authToken = jsonData.getString("authToken");
             }
             JSONObject json = jsonData.getJSONObject("data");
+            if (jsonData.has(IterableConstants.KEY_CREATED_AT)) {
+                json.put(IterableConstants.KEY_CREATED_AT, jsonData.getString(IterableConstants.KEY_CREATED_AT));
+            }
             return new IterableApiRequest(apikey, resourcePath, json, requestType, authToken, onSuccess, onFailure);
         } catch (JSONException e) {
             IterableLogger.e(TAG, "Failed to create Iterable request from JSON");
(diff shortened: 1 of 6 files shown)


iterableapi/src/main/java/com/iterable/iterableapi/IterableTask.java
```diff
@@ -12,11 +12,11 @@
     String id;
     String name;
     int version;
-    Date createdAt;
-    Date modifiedAt;
-    Date lastAttemptedAt;
-    Date scheduledAt;
-    Date requestedAt;
+    long createdAt;
```
vbabenkoru, line 15: If we change column types, I think we also should change the `CREATE TABLE` statement to specify numeric types for those columns in SQLite.

iterableapi/src/main/java/com/iterable/iterableapi/IterableTaskRunner.java
```diff
@@ -104,7 +105,8 @@ private boolean processNextTask() {
             IterableApiResponse response = null;
             TaskResult result = TaskResult.FAILURE;
             try {
-                IterableApiRequest request = IterableApiRequest.fromJSON(new JSONObject(task.data), null, null);
+                IterableApiRequest request = IterableApiRequest.fromJSON(getTaskDataWithDate(task), null, null);
+                //TODO: Any chances of request getting null?
```
vbabenkoru, line 109: Yes, but it will get caught by the statement just below this.

iterableapi/src/main/java/com/iterable/iterableapi/IterableTaskStorage.java
```diff
@@ -112,18 +112,19 @@ String createTask(String name, IterableTaskType type, String data) {
         contentValues.put(TASK_ID, iterableTask.id);
         contentValues.put(NAME, iterableTask.name);
         contentValues.put(VERSION, iterableTask.version);
-        contentValues.put(CREATED_AT, iterableTask.createdAt.toString());
-        if (iterableTask.modifiedAt != null) {
-            contentValues.put(MODIFIED_AT, iterableTask.modifiedAt.toString());
+        //TODO: you might have to change the created to epoch literally while storing
```
vbabenkoru, line 115: What does this mean?

iterableapi/src/main/java/com/iterable/iterableapi/IterableTaskRunner.java
```diff
@@ -119,6 +121,17 @@ private boolean processNextTask() {
         return false;
     }
 
+    JSONObject getTaskDataWithDate(IterableTask task) {
+        try {
+            JSONObject jsonData = new JSONObject(task.data);
+            jsonData.put(IterableConstants.KEY_CREATED_AT, task.createdAt);
```
vbabenkoru, line 127: Convert to seconds

vbabenkoru (changes requested):

iterableapi/src/main/java/com/iterable/iterableapi/IterableTaskStorage.java
Ayyanchira: 😬 😅 That was a note to myself to change the type from string to long in database 😅. Forgot to make that change.

iterableapi/src/main/java/com/iterable/iterableapi/IterableTaskStorage.java
```diff
@@ -42,11 +42,11 @@
     static final String OFFLINE_TASK_COLUMN_DATA = " (" + TASK_ID + " TEXT PRIMARY KEY," +
             NAME + " TEXT," +
             VERSION + " INTEGER," +
-            CREATED_AT + " TEXT," +
-            MODIFIED_AT + " TEXT," +
-            LAST_ATTEMPTED_AT + " TEXT," +
-            SCHEDULED_AT + " TEXT," +
-            REQUESTED_AT + " TEXT," +
+            CREATED_AT + " LONG," +
```
vbabenkoru, line 45: Is that a valid column type in SQLite? Have you tried reinstalling the app or deleting the database and running this table creation query?

Ayyanchira: Yes. Tried. It says Numeric on datatype when looked into the table

Ayyanchira: Although, I will just have them as Numeric... why risk..

vbabenkoru: I was just reading the SQLite documentation on types, and the most appropriate seems to be `INTEGER` (or any of its aliases): https://www.sqlite.org/datatype3.html#affinity_name_examples

vbabenkoru: Didn't see `LONG` there so I'm wondering if it's something that may not be supported in current or past versions of SQLite on various devices.

iterableapi/src/main/java/com/iterable/iterableapi/IterableRequestTask.java
```diff
@@ -365,6 +368,9 @@ static IterableApiRequest fromJSON(JSONObject jsonData, @Nullable IterableHelper
                 authToken = jsonData.getString("authToken");
             }
             JSONObject json = jsonData.getJSONObject("data");
+            if (jsonData.has(IterableConstants.KEY_CREATED_AT)) {
```
vbabenkoru, line 371: Why not just add it to `data` in `getTaskDataWithDate`? That way we won't need extra code here.

iterableapi/src/main/java/com/iterable/iterableapi/IterableTaskStorage.java
```diff
@@ -42,11 +42,11 @@
     static final String OFFLINE_TASK_COLUMN_DATA = " (" + TASK_ID + " TEXT PRIMARY KEY," +
             NAME + " TEXT," +
             VERSION + " INTEGER," +
-            CREATED_AT + " TEXT," +
-            MODIFIED_AT + " TEXT," +
-            LAST_ATTEMPTED_AT + " TEXT," +
-            SCHEDULED_AT + " TEXT," +
-            REQUESTED_AT + " TEXT," +
+            CREATED_AT + " NUMERIC," +
```
vbabenkoru, line 45: See this comment: https://github.com/Iterable/iterable-android-sdk/pull/307#discussion_r562898987
I think this should be `INTEGER` or `BIGINT`.

vbabenkoru (approved):

PR #308, opened 2021-01-19
[MOB-2435] Offline mode: handle network errors

vbabenkoru:
## 🔹 Jira Ticket(s) if any

* [MOB-2435](https://iterable.atlassian.net/browse/MOB-2435)

## ✏️ Description

Add logic to handle retriable errors and reschedule processing with a set interval (60 seconds) whenever they happen.

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableTaskRunner.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableTaskRunner.java
@@ -13,11 +13,13 @@ import org.json.JSONObject;
 
 import java.util.ArrayList;
 
-class IterableTaskRunner implements IterableTaskStorage.TaskCreatedListener, Handler.Callback, IterableNetworkConnectivityManager.IterableNetworkMonitorListener {
+class IterableTaskRunner implements IterableTaskStorage.TaskCreatedListener, Handler.Callback, IterableNetworkConnectivityManager.IterableNetworkMonitorListener, IterableActivityMonitor.AppStateCallback {
     private static final String TAG = "IterableTaskRunner";
     private IterableTaskStorage taskStorage;
     private IterableActivityMonitor activityMonitor;
     private IterableNetworkConnectivityManager networkConnectivityManager;
+
+    private static final int RETRY_INTERVAL_SECONDS = 60;
 
     private static final int OPERATION_PROCESS_TASKS = 100;
 
@@ -43,6 +45,7 @@ class IterableTaskRunner implements IterableTaskStorage.TaskCreatedListener, Han
         handler = new Handler(networkThread.getLooper(), this);
         taskStorage.addTaskCreatedListener(this);
         networkConnectivityManager.addNetworkListener(this);
+        activityMonitor.addCallback(this);
     }
 
     void addTaskCompletedListener(TaskCompletedListener listener) {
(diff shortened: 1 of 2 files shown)


iterableapi/src/main/java/com/iterable/iterableapi/IterableTaskRunner.java
```diff
@@ -85,21 +93,22 @@ public boolean handleMessage(@NonNull Message msg) {
...
+            IterableTask task = taskStorage.getNextScheduledTask();
+
+            if (task == null) {
+                return;
+            }
+
+            boolean proceed = processTask(task);
             if (!proceed) {
-                break;
+                scheduleRetry();
```
Ayyanchira, line 104: Should we set any limits onto how many times we will retry?

vbabenkoru: This is mirroring the iOS logic; I think we do need some kind of limits and/or exponential backoff. We'll need to discuss whether this is something we'd want to do before beta.

iterableapi/src/main/java/com/iterable/iterableapi/IterableTaskRunner.java
```diff
@@ -67,11 +70,27 @@ public void onNetworkConnected() {
     public void onNetworkDisconnected() {
 
     }
+
+    @Override
+    public void onSwitchToForeground() {
+        runNow();
```
Ayyanchira, line 76: 👍

iterableapi/src/test/java/com/iterable/iterableapi/IterableTaskRunnerTest.java
```diff
@@ -90,17 +96,43 @@ public void testNoRequestsWhenOffline() throws Exception {
...
+        when(mockActivityMonitor.isInForeground()).thenReturn(false);
+        when(mockNetworkConnectivityManager.isConnected()).thenReturn(true);
+        IterableTaskRunner.TaskCompletedListener taskCompletedListener = mock(IterableTaskRunner.TaskCompletedListener.class);
+        taskRunner.addTaskCompletedListener(taskCompletedListener);
+        server.enqueue(new MockResponse().setResponseCode(200).setBody("{}"));
+
+        taskRunner.onTaskCreated(null);
+        runHandlerTasks(taskRunner);
+
+        verify(mockActivityMonitor, times(1)).isInForeground();
```
Ayyanchira, line 132: How did we test if no request was made when in background as the method name suggests?

Ayyanchira (approved):

iterableapi/src/main/java/com/iterable/iterableapi/IterableTaskRunner.java
Ayyanchira: Just came in my mind... 
We are retrying the requests which says failed to connect. I think its assuming either Iterable server is down or network is down. We can skip retying here as our network monitor will be switching to offline and online back again when network comes back. and then it will call processTask again.
We wont need another manual scheduleRetry. We can just stop processing and not delete that particular task.

vbabenkoru: Network monitor only reports as 'offline' when the device is completely off the grid: there is no Wi-Fi connection or cellular connection at all.
This code handles the situation when the connection exists, but is not working (100% packet loss).

iterableapi/src/main/java/com/iterable/iterableapi/IterableTaskRunner.java
```diff
@@ -84,22 +103,28 @@ public boolean handleMessage(@NonNull Message msg) {
 
     @WorkerThread
     private void processTasks() {
+        if (!activityMonitor.isInForeground()) {
+            IterableLogger.d(TAG, "App not in foreground, skipping processing tasks");
+            return;
+        }
```
Ayyanchira, line 109: We can totally process the tasks when in background.

vbabenkoru: Yeah, but out of caution, we don't. Otherwise there is a risk that the app will continue to make network requests every minute, draining the battery. I think we may want to revisit that logic and have another kind of safeguard and allow background processing.

vbabenkoru: I also copied this logic from iOS, to mirror the behavior.

Ayyanchira: Cool 👍. Will update the test case written in slab doc accordingly then...

PR #309, opened 2021-01-20
[MOB-2437] - Get Remote Configuration

Ayyanchira:
## 🔹 Jira Ticket(s) if any

https://iterable.atlassian.net/browse/MOB-2437

## ✏️ Description

Get remote configuration for offline mode.
1. Call get remote configuration when app comes to foreground. Should ideally call it during initialization But existing tests are failing.
2. On Success -> call the setOfflineProcessingEnabled
3. A type check to NOT reassign and reinitialize similar request processor.

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
@@ -311,19 +311,15 @@ private static final String TAG = "IterableApi";
     }
 
     void fetchRemoteConfiguration() {
-        apiClient.getRemoteConfiguration(new IterableHelper.SuccessHandler() {
+        apiClient.getRemoteConfiguration(new IterableHelper.IterableActionHandler() {
             @Override
-            public void onSuccess(@NonNull JSONObject data) {
+            public void execute(@Nullable String data) {
                 try {
-                    sharedInstance.apiClient.setOfflineProcessingEnabled(data.getBoolean("offlineModeBeta"));
+                    JSONObject jsonData = new JSONObject(data);
+                    sharedInstance.apiClient.setOfflineProcessingEnabled(jsonData.getBoolean("offlineModeBeta"));
                 } catch (JSONException e) {
-                    IterableLogger.e(TAG, "OfflineMode parameter not found");
+                    IterableLogger.e(TAG, "Failed to read remote configuration");
                 }
-            }
-        }, new IterableHelper.FailureHandler() {
-            @Override
-            public void onFailure(@NonNull String reason, @Nullable JSONObject data) {
-                IterableLogger.e(TAG, "Failed to fetch remote configuration. Default values will be used");
             }
         });
     }
@@ -984,8 +980,8 @@ private static final String TAG = "IterableApi";
                 IterableLogger.d(TAG, "Performing automatic push registration");
                 sharedInstance.registerForPush();
             }
+            fetchRemoteConfiguration();
         }
-        fetchRemoteConfiguration();
     }
 
     private boolean isInitialized() {
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
@@ -310,6 +310,24 @@ private static final String TAG = "IterableApi";
         IterablePushActionReceiver.processPendingAction(context);
     }
 
+    void fetchRemoteConfiguration() {
+        apiClient.getRemoteConfiguration(new IterableHelper.SuccessHandler() {
+            @Override
+            public void onSuccess(@NonNull JSONObject data) {
+                try {
+                    sharedInstance.apiClient.setOfflineProcessingEnabled(data.getBoolean("offlineModeBeta"));
+                } catch (JSONException e) {
+                    IterableLogger.e(TAG, "OfflineMode parameter not found");
+                }
+            }
+        }, new IterableHelper.FailureHandler() {
+            @Override
+            public void onFailure(@NonNull String reason, @Nullable JSONObject data) {
+                IterableLogger.e(TAG, "Failed to fetch remote configuration. Default values will be used");
+            }
+        });
+    }
+
     /**
      * Set user email used for API calls
      * Calling this or {@link #setUserId(String)} is required before making any API calls.
@@ -967,6 +985,7 @@ private static final String TAG = "IterableApi";
                 sharedInstance.registerForPush();
             }
         }
+        fetchRemoteConfiguration();
     }
 
     private boolean isInitialized() {
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableApiClient.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableApiClient.java
@@ -60,15 +60,14 @@ class IterableApiClient {
         }
     }
 
-    void getRemoteConfiguration(IterableHelper.SuccessHandler successHandler, IterableHelper.FailureHandler failureHandler) {
+    void getRemoteConfiguration(IterableHelper.IterableActionHandler actionHandler) {
         JSONObject requestJSON = new JSONObject();
         try {
-            //TODO: Add data and make request
             requestJSON.putOpt(IterableConstants.KEY_PLATFORM, IterableConstants.ITBL_PLATFORM_ANDROID);
             requestJSON.putOpt(IterableConstants.DEVICE_APP_PACKAGE_NAME, authProvider.getContext().getPackageName());
             requestJSON.put(IterableConstants.ITBL_KEY_SDK_VERSION, IterableConstants.ITBL_KEY_SDK_VERSION_NUMBER);
             requestJSON.put(IterableConstants.ITBL_SYSTEM_VERSION, Build.VERSION.RELEASE);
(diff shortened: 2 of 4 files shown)


iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -306,10 +306,29 @@ public static void initialize(@NonNull Context context, @NonNull String apiKey,
             sharedInstance.inAppManager = new IterableInAppManager(sharedInstance, sharedInstance.config.inAppHandler,
                     sharedInstance.config.inAppDisplayInterval);
         }
+//        sharedInstance.getRemoteConfiguration();
         sharedInstance.apiClient.setOfflineProcessingEnabled(sharedInstance.config.offlineProcessing);
         IterablePushActionReceiver.processPendingAction(context);
     }
 
+    void getRemoteConfiguration() {
```
vbabenkoru, line 314: `get` usually implies a getter that returns something to the caller. In this case, it doesn't return anything and processes the configuration itself. I think a better name might be something like `fetchRemoteConfiguration`.
Also I think we should persist the last fetched values and read them on launch, otherwise it will fallback to the default value every time.

vbabenkoru: Re: persistence, I see this is also something that's not in iOS, so it's not something that's necessary for this PR. I've opened a separate thread on Slack to discuss whether we need persistence for beta (both iOS and Android).

iterableapi/src/main/java/com/iterable/iterableapi/IterableApiClient.java
```diff
@@ -49,13 +49,31 @@ private RequestProcessor getRequestProcessor() {
     void setOfflineProcessingEnabled(boolean offlineMode) {
         if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
             if (offlineMode) {
-                this.requestProcessor = new OfflineRequestProcessor(authProvider.getContext());
+                if (this.requestProcessor == null || this.requestProcessor.getClass() != OfflineRequestProcessor.class) {
```
vbabenkoru, line 52: Maybe we should go further and two request processors should be lazily instantiated singletons, so we don't create another one when switching back and forth. Not likely to happen in reality, but still..

iterableapi/src/main/java/com/iterable/iterableapi/IterableApiClient.java
```diff
@@ -49,13 +49,31 @@ private RequestProcessor getRequestProcessor() {
...
+                    this.requestProcessor = new OnlineRequestProcessor();
+                }
             }
         }
     }
 
+    void getRemoteConfiguration(IterableHelper.SuccessHandler successHandler, IterableHelper.FailureHandler failureHandler) {
+        JSONObject requestJSON = new JSONObject();
+        try {
+            //TODO: Add data and make request
```
vbabenkoru, line 66: ?

iterableapi/src/test/java/com/iterable/iterableapi/IterableApiTest.java
```diff
@@ -668,4 +668,25 @@ public void testTrackInAppDeleteWithNullParameters() throws Exception {
...
 
 
+    @Test
+    public void testFetchRemoteConfigurationCalledWhenInForeground() throws Exception {
+
+        server.enqueue(new MockResponse().setResponseCode(200).setBody("{\n" +
+                "            \"offlineMode\": false,\n" +
+                "                \"offlineModeBeta\": true,\n" +
+                "                \"someOtherKey1\": \"someOtherValue1\",\n" +
+                "            // ....\n" +
```
vbabenkoru, line 678: Invalid JSON

iterableapi/src/test/java/com/iterable/iterableapi/IterableApiTest.java
```diff
@@ -668,4 +668,25 @@ public void testTrackInAppDeleteWithNullParameters() throws Exception {
...
+                "        }"));
+
```
vbabenkoru, line 680: Add
```
        IterableActivityMonitor.getInstance().unregisterLifecycleCallbacks(getContext());
        IterableActivityMonitor.instance = new IterableActivityMonitor();
```

iterableapi/src/test/java/com/iterable/iterableapi/IterableApiTest.java
```diff
@@ -668,4 +668,25 @@ public void testTrackInAppDeleteWithNullParameters() throws Exception {
...
+        IterableApi.initialize(getContext(), "apiKey", new IterableConfig.Builder().setAutoPushRegistration(false).build());
+        verify(mockApiClient).setOfflineProcessingEnabled(false);
+        clearInvocations(mockApiClient);
+        Robolectric.buildActivity(Activity.class).create().start().resume();
+        shadowOf(getMainLooper()).idle();
+        RecordedRequest trackInAppConsumeRequest = server.takeRequest(1, TimeUnit.SECONDS);
+        assertNotNull(trackInAppConsumeRequest);
+        assertTrue(trackInAppConsumeRequest.getRequestUrl().toString().contains("/getRemoteConfiguration"));
+//        verify(mockApiClient).setOfflineProcessingEnabled(true);
```
vbabenkoru, line 689: Add at the end:
```                
        IterableActivityMonitor.getInstance().unregisterLifecycleCallbacks(getContext());
        IterableActivityMonitor.instance = new IterableActivityMonitor();
```

vbabenkoru (changes requested):

iterableapi/src/main/java/com/iterable/iterableapi/IterableApiClient.java
Ayyanchira: Yeah. Looks like we will eventually have both offline and online request processor when it goes GA

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -966,6 +980,7 @@ private void onForeground() {
                 IterableLogger.d(TAG, "Performing automatic push registration");
                 sharedInstance.registerForPush();
             }
+            fetchRemoteConfiguration();
```
vbabenkoru, line 983: I think this should be run once on launch (within the `if` block above), instead of on every foreground event.

iterableapi/src/test/java/com/iterable/iterableapi/IterableApiTest.java
```diff
@@ -668,4 +668,29 @@ public void testTrackInAppDeleteWithNullParameters() throws Exception {
     }
 
 
+    @Test
+    public void testFetchRemoteConfigurationCalledWhenInForeground() throws Exception {
+
+        server.enqueue(new MockResponse().setResponseCode(200).setBody("{\n" +
+                "            \"offlineMode\": false,\n" +
+                "                \"offlineModeBeta\": true,\n" +
```
vbabenkoru, line 676: Does this need extra indentation?

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
Ayyanchira: If it goes to if condition above it, the function will have to depend on autopushregistration flag as well. 
By having it here, looks like it wont be called on every foreground as its inside `if (!_firstForegroundHandled)` condition.

iterableapi/src/test/java/com/iterable/iterableapi/IterableApiTest.java
Ayyanchira: adjusted. Thanks :)

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -310,6 +310,20 @@ public static void initialize(@NonNull Context context, @NonNull String apiKey,
         IterablePushActionReceiver.processPendingAction(context);
     }
 
+    void fetchRemoteConfiguration() {
+        apiClient.getRemoteConfiguration(new IterableHelper.IterableActionHandler() {
+            @Override
+            public void execute(@Nullable String data) {
+                try {
+                    JSONObject jsonData = new JSONObject(data);
+                    sharedInstance.apiClient.setOfflineProcessingEnabled(jsonData.getBoolean("offlineModeBeta"));
```
tapashmajumder, line 319: This doesn't seem to the right flag coming from the backend. It should be `SDKOfflineModeBeta`

tapashmajumder: Oh. Wait

tapashmajumder: You are right. Never mind.

tapashmajumder: Nit: Not needed for +1. This is the correct flag. But we should probably load it from some constant file.

Ayyanchira: Added them in constants in other PR where its getting stored persistently

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
vbabenkoru: Ah, yeah, sorry, you're right! I missed the extra bracket and thought it was outside the conditional block.

vbabenkoru (approved):

PR #311, opened 2021-01-25
[MOB-2531] - Persist offline configuration

Ayyanchira:
## 🔹 Jira Ticket(s) if any

https://iterable.atlassian.net/browse/MOB-2531

## ✏️ Description

1. Storing offline feature flag to shared preference.
2. Loading offlineconfiguration from shared preference during initialization. Default set to false. If nothing in shared preference, it will be false again.

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
@@ -306,8 +306,14 @@ private static final String TAG = "IterableApi";
             sharedInstance.inAppManager = new IterableInAppManager(sharedInstance, sharedInstance.config.inAppHandler,
                     sharedInstance.config.inAppDisplayInterval);
         }
-        sharedInstance.apiClient.setOfflineProcessingEnabled(sharedInstance.config.offlineProcessing);
+        loadLastSavedConfiguration(context);
         IterablePushActionReceiver.processPendingAction(context);
+    }
+
+    static void loadLastSavedConfiguration(Context context) {
+        SharedPreferences sharedPref = context.getSharedPreferences(IterableConstants.SHARED_PREFS_SAVED_CONFIGURATION, Context.MODE_PRIVATE);
+        boolean offlineMode = sharedPref.getBoolean(IterableConstants.SHARED_PREFS_OFFLINE_MODE_BETA_KEY, false);
+        sharedInstance.apiClient.setOfflineProcessingEnabled(offlineMode);
     }
 
     void fetchRemoteConfiguration() {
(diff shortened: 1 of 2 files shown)


iterableapi/src/main/java/com/iterable/iterableapi/IterableConstants.java
```diff
@@ -107,6 +107,8 @@
     public static final String SHARED_PREFS_ATTRIBUTION_INFO_KEY = "itbl_attribution_info";
     public static final int    SHARED_PREFS_ATTRIBUTION_INFO_EXPIRATION_HOURS = 24;
     public static final String SHARED_PREFS_FCM_MIGRATION_DONE_KEY = "itbl_fcm_migration_done";
+    public static final String SHARED_PREFS_SAVED_CONFIGURATION = "itbl_saved_configuration";
+    public static final String SHARED_PREFS_OFFLINE_MODE_BETA_KEY = "offlineModeBeta";
```
davidtruong, line 111: Since this is refreshed on every app launch, is there a reason this is a Beta? Were you planning on changing this after beta?

Ayyanchira: Yes. These flags will be adjusted when going to GA.

davidtruong: Why do we need to adjust them if they will get overwritten anyways on launch?

vbabenkoru: @davidtruong There are two feature flags: one for beta, one for release (GA). We're only going to enable the beta feature flag during beta, and then switch to using the GA flag on GA release.

davidtruong: Gotcha, didn't realize we have 2 pulled from different feature flags.

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -306,17 +306,28 @@ public static void initialize(@NonNull Context context, @NonNull String apiKey,
             sharedInstance.inAppManager = new IterableInAppManager(sharedInstance, sharedInstance.config.inAppHandler,
                     sharedInstance.config.inAppDisplayInterval);
         }
-        sharedInstance.apiClient.setOfflineProcessingEnabled(sharedInstance.config.offlineProcessing);
```
vbabenkoru, line 309: Let's remove `offlineProcessing` from `IterableConfig` then, I think iOS already removed it.

Ayyanchira: Makes sense... 👍

Ayyanchira: I stumbled upon this case where -> we clear the database on logout, only if the offlineConfiguration was enabled.
Now that its is remote configuration. We might have already stored some data and then offline configuration could be set to false at later point. 
We can always clear the database regardless of this feature but we do initialize and create a database for it. 
Thinking of what check I can do...

vbabenkoru: Synced over Zoom, split this into a separate ticket/task.

vbabenkoru (approved):

PR #313, opened 2021-01-26
[MOB-2549] - createdAt in all the request

Ayyanchira:
## 🔹 Jira Ticket(s) if any

https://iterable.atlassian.net/browse/MOB-2549

## ✏️ Description

Added createdAt field in Online Request Processor. Thus now all the requests will have `createdAt` timestamp in them.

### DIFF
--- a/iterableapi/src/androidTest/java/com/iterable/iterableapi/IterableApiRequestsTest.java
+++ b/iterableapi/src/androidTest/java/com/iterable/iterableapi/IterableApiRequestsTest.java
@@ -10,6 +10,7 @@ import org.junit.Test;
 import org.junit.runner.RunWith;
 
 import java.util.ArrayList;
+import java.util.Date;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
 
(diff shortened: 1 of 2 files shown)


iterableapi/src/androidTest/java/com/iterable/iterableapi/IterableApiRequestsTest.java
```diff
@@ -65,7 +66,7 @@ public void testGetRequestHeaders() throws Exception {
 
     @Test
     public void testTrackPurchase() throws Exception {
-        String expectedRequest = "{\"user\":{\"email\":\"test_email\"},\"items\":[{\"id\":\"sku123\",\"name\":\"Item\",\"price\":50,\"quantity\":2}],\"total\":100}";
+        String expectedRequest = new StringBuilder(new StringBuffer("{\"user\":{\"email\":\"test_email\"},\"items\":[{\"id\":\"sku123\",\"name\":\"Item\",\"price\":50,\"quantity\":2}],\"total\":100").append(",\"createdAt\":").append(new Date().getTime()/1000).append("}")).toString();
```
vbabenkoru, line 69: This may fail every once in a while if there is significant enough difference between when the request was sent and when it was checked. We'll need to change this test if it becomes flaky.

vbabenkoru (approved):

PR #314, opened 2021-01-26
[MOB-2548] - send SDK Request Processor in Header

Ayyanchira:
## 🔹 Jira Ticket(s) if any
https://iterable.atlassian.net/browse/MOB-2548

## ✏️ Description

1. Added new enum ProcessorType in `IterableApiRequest`
2. Added the field in `IterableConstants`
3. Added a setter for processing type. If setter is not used, it will be defaulted to Online Processing.
4. All the cascading changes to other classes - Now offline processor's get and post request uses this setter before executing the request. Also same changes for TaskRunner where it fetches the task from database and executes the request.

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableRequestTask.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableRequestTask.java
@@ -89,7 +89,7 @@ class IterableRequestTask extends AsyncTask<IterableApiRequest, Void, IterableAp
                     urlConnection.setRequestProperty(IterableConstants.HEADER_SDK_PLATFORM, "Android");
                     urlConnection.setRequestProperty(IterableConstants.HEADER_SDK_VERSION, IterableConstants.ITBL_KEY_SDK_VERSION_NUMBER);
                     urlConnection.setRequestProperty(IterableConstants.KEY_SENT_AT, String.valueOf(new Date().getTime() / 1000));
-                    urlConnection.setRequestProperty(IterableConstants.HEADER_SDK_PROCESSOR_TYPE, iterableApiRequest.processorType.toString());
+                    urlConnection.setRequestProperty(IterableConstants.HEADER_SDK_PROCESSOR_TYPE, iterableApiRequest.getProcessorType().toString());
                     if (iterableApiRequest.authToken != null) {
                         urlConnection.setRequestProperty(IterableConstants.HEADER_SDK_AUTHORIZATION, IterableConstants.HEADER_SDK_AUTH_FORMAT + iterableApiRequest.authToken);
                     }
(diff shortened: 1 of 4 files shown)


iterableapi/src/main/java/com/iterable/iterableapi/IterableRequestTask.java
```diff
@@ -311,11 +311,29 @@ protected void setRetryCount(int count) {
...
+        OFFLINE {
+            @NonNull
+            @Override
+            public String toString() {
+                return "Offline";
+            }
+        }
+    }
+
     IterableApiRequest(String apiKey, String baseUrl, String resourcePath, JSONObject json, String requestType, String authToken, IterableHelper.SuccessHandler onSuccess, IterableHelper.FailureHandler onFailure) {
```
vbabenkoru, line 337: What's calling this constructor now?

iterableapi/src/androidTest/java/com/iterable/iterableapi/IterableApiResponseTest.java
```diff
@@ -204,7 +204,7 @@ public void testResponseCode500() throws Exception {
             stubAnyRequestReturningStatusCode(500, "{}");
         }
 
-        IterableApiRequest request = new IterableApiRequest("fake_key", "", new JSONObject(), IterableApiRequest.POST, null, null, null);
+        IterableApiRequest request = new IterableApiRequest("fake_key", "", new JSONObject(), IterableApiRequest.POST, null, (IterableHelper.SuccessHandler) null, null);
```
vbabenkoru, line 207: Looks weird

iterableapi/src/main/java/com/iterable/iterableapi/IterableRequestTask.java
```diff
@@ -336,6 +368,19 @@ protected void setRetryCount(int count) {
         this.authToken = authToken;
         this.successCallback = onSuccess;
         this.failureCallback = onFailure;
+        this.processorType = ProcessorType.ONLINE;
+    }
+
+    IterableApiRequest(String apiKey, String resourcePath, JSONObject json, String requestType, String authToken, IterableHelper.SuccessHandler onSuccess, IterableHelper.FailureHandler onFailure, ProcessorType processorType) {
```
vbabenkoru, line 374: Can we reduce the number of constructors?

iterableapi/src/androidTest/java/com/iterable/iterableapi/IterableApiResponseTest.java
Ayyanchira: Yeah super weird.. 😓 Six constructor... 
Or should I pass Processor type from each request maker?

iterableapi/src/main/java/com/iterable/iterableapi/IterableRequestTask.java
```diff
@@ -374,6 +430,25 @@ static IterableApiRequest fromJSON(JSONObject jsonData, @Nullable IterableHelper
         }
         return null;
     }
+
+    static IterableApiRequest fromJSON(JSONObject jsonData, @Nullable IterableHelper.SuccessHandler onSuccess, @Nullable IterableHelper.FailureHandler onFailure, ProcessorType processorType) {
```
vbabenkoru, line 434: Do we need two separate copies of this method? Wondering if we can reduce duplication.

Ayyanchira: Yes.. yes... the other one can in turn use this one inside..

iterableapi/src/main/java/com/iterable/iterableapi/IterableRequestTask.java
Ayyanchira: basically the ones which were calling constructor will null parameters are now confused... 
Will try to reduce number of constructors

iterableapi/src/main/java/com/iterable/iterableapi/IterableRequestTask.java
```diff
@@ -89,7 +89,7 @@ static IterableApiResponse executeApiRequest(IterableApiRequest iterableApiReque
                     urlConnection.setRequestProperty(IterableConstants.HEADER_SDK_PLATFORM, "Android");
                     urlConnection.setRequestProperty(IterableConstants.HEADER_SDK_VERSION, IterableConstants.ITBL_KEY_SDK_VERSION_NUMBER);
                     urlConnection.setRequestProperty(IterableConstants.KEY_SENT_AT, String.valueOf(new Date().getTime() / 1000));
-
+                    urlConnection.setRequestProperty(IterableConstants.HEADER_SDK_PROCESSOR_TYPE, iterableApiRequest.processorType.toString());
```
tapashmajumder, line 92: Looks like `processorType` is both an internal field if iterableApiRequest and it is also exposed as a setter via `setProcessorType`.  Why not set the field as private and add a getter `getProcessorType` in addition to the setter.

tapashmajumder (approved):
+1

PR #315, opened 2021-01-26
[MOB - 2541] - Remove offline configuration from IterableConfig

Ayyanchira:
## 🔹 Jira Ticket(s) if any

https://iterable.atlassian.net/browse/MOB-2541

## ✏️ Description

1. Removed offlineProcessing flag from IterableConfig
2. New interface method `onLogout(context)` introduced in `RequestProcessor`
3.  `apiClient` now has onLogout() method which makes calls to its requestProcessor's `onLogout(context)`
4. `IterableAPI` calls `onLogout()` of `apiClient`
5. Updated test method to check if database is cleared by calling fetch remote configuration.

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/OfflineRequestProcessor.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/OfflineRequestProcessor.java
@@ -18,6 +18,7 @@ import java.util.Set;
 class OfflineRequestProcessor implements RequestProcessor {
     private TaskScheduler taskScheduler;
     private IterableTaskRunner taskRunner;
+    private IterableTaskStorage taskStorage;
     private static final Set<String> offlineApiSet = new HashSet<>(Arrays.asList(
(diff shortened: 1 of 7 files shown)


iterableapi/src/main/java/com/iterable/iterableapi/OfflineRequestProcessor.java
```diff
@@ -58,6 +58,11 @@ public void processPostRequest(@Nullable String apiKey, @NonNull String resource
         }
     }
 
+    @Override
+    public void onLogout(Context context) {
+        IterableTaskStorage.sharedInstance(context).deleteAllTasks();
```
vbabenkoru, line 63: Make `taskStorage` a field in this class (line 33 in this file) and access it directly here. It's better to use class fields instead of global instances.

vbabenkoru (changes requested):

vbabenkoru (approved):

PR #323, opened 2021-03-02
[MOB-2606] persistent read state for inbox in-apps

roninopf:
## 🔹 Jira Ticket(s) if any

* [MOB-2606](https://iterable.atlassian.net/browse/MOB-2606)

## ✏️ Description

* adds SDK support for persisting read state per user rather than local device only

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppManager.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppManager.java
@@ -307,20 +307,31 @@ public class IterableInAppManager implements IterableActivityMonitor.AppStateCal
     private void syncWithRemoteQueue(List<IterableInAppMessage> remoteQueue) {
         boolean changed = false;
         Map<String, IterableInAppMessage> remoteQueueMap = new HashMap<>();
+
         for (IterableInAppMessage message : remoteQueue) {
             remoteQueueMap.put(message.getMessageId(), message);
+
             if (storage.getMessage(message.getMessageId()) == null) {
                 storage.addMessage(message);
                 onMessageAdded(message);
                 changed = true;
             }
+
+            if (storage.getMessage(message.getMessageId()) != null && message.isRead()) {
+                IterableInAppMessage localMessage = storage.getMessage(message.getMessageId());
+                localMessage.setRead(message.isRead());
+
+                changed = true;
+            }
         }
+
         for (IterableInAppMessage localMessage : storage.getMessages()) {
             if (!remoteQueueMap.containsKey(localMessage.getMessageId())) {
                 storage.removeMessage(localMessage);
                 changed = true;
             }
         }
+
         scheduleProcessing();
         if (changed) {
             notifyOnChange();
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppManager.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppManager.java
@@ -317,11 +317,14 @@ public class IterableInAppManager implements IterableActivityMonitor.AppStateCal
                 changed = true;
             }
 
-            if (storage.getMessage(message.getMessageId()) != null && message.isRead()) {
+            if (storage.getMessage(message.getMessageId()) != null) {
                 IterableInAppMessage localMessage = storage.getMessage(message.getMessageId());
-                localMessage.setRead(message.isRead());
+                
+                if (!localMessage.isRead() && message.isRead()) {
+                    localMessage.setRead(message.isRead());
 
-                changed = true;
+                    changed = true;
+                }
             }
         }
 
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppManager.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppManager.java
@@ -311,16 +311,20 @@ public class IterableInAppManager implements IterableActivityMonitor.AppStateCal
         for (IterableInAppMessage message : remoteQueue) {
             remoteQueueMap.put(message.getMessageId(), message);
 
-            if (!isStoringInApp(message)) {
+            boolean isInAppStored = storage.getMessage(message.getMessageId()) != null;
+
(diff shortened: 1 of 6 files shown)


iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppManager.java
```diff
@@ -307,20 +307,31 @@ private boolean isMessageExpired(IterableInAppMessage message) {
...
         for (IterableInAppMessage message : remoteQueue) {
             remoteQueueMap.put(message.getMessageId(), message);
+
             if (storage.getMessage(message.getMessageId()) == null) {
                 storage.addMessage(message);
                 onMessageAdded(message);
                 changed = true;
             }
+
+            if (storage.getMessage(message.getMessageId()) != null && message.isRead()) {
```
vbabenkoru, line 320: I think this should only be run when the read state is different in the server response vs. local state. If the read flag didn't change, we shouldn't set `changed` to `true`.

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppManager.java
```diff
@@ -307,20 +307,34 @@ private boolean isMessageExpired(IterableInAppMessage message) {
...
         for (IterableInAppMessage message : remoteQueue) {
             remoteQueueMap.put(message.getMessageId(), message);
+
             if (storage.getMessage(message.getMessageId()) == null) {
                 storage.addMessage(message);
                 onMessageAdded(message);
                 changed = true;
             }
+
+            if (storage.getMessage(message.getMessageId()) != null) {
```
vbabenkoru, line 320: `else`?

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppManager.java
```diff
@@ -358,7 +372,7 @@ private void processMessages() {
         List<IterableInAppMessage> messagesByPriorityLevel = getMessagesSortedByPriorityLevel(messages);
 
         for (IterableInAppMessage message : messagesByPriorityLevel) {
-            if (!message.isProcessed() && !message.isConsumed() && message.getTriggerType() == TriggerType.IMMEDIATE) {
+            if (!message.isProcessed() && !message.isConsumed() && message.getTriggerType() == TriggerType.IMMEDIATE && !message.isRead()) {
```
Ayyanchira, line 375: Also wondering how `isProcessed()` will hold its significance? Wont `isRead()` serve the purpose now?

Ayyanchira: @vbabenkoru ?

Ayyanchira: Just thinking if isProcessed can be eliminated... 🤔

roninopf: `isProcessed` is different than being `read` though because `isProcessed` applies to every in-app message, whereas `read` is relevant to inbox.

Ayyanchira: ohh yeah. thats correct 👍

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppManager.java
vbabenkoru: If the reason is for readability, I think the check should be stored in a named variable to express intent. It takes a while to understand what `storage.getMessage(message.getMessageId()) == null` and `storage.getMessage(message.getMessageId()) != null` mean.

roninopf: Will add, that's a good idea.

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppManager.java
```diff
@@ -307,20 +307,34 @@ private boolean isMessageExpired(IterableInAppMessage message) {
     private void syncWithRemoteQueue(List<IterableInAppMessage> remoteQueue) {
         boolean changed = false;
         Map<String, IterableInAppMessage> remoteQueueMap = new HashMap<>();
+
         for (IterableInAppMessage message : remoteQueue) {
             remoteQueueMap.put(message.getMessageId(), message);
-            if (storage.getMessage(message.getMessageId()) == null) {
+
+            if (!isStoringInApp(message)) {
```
vbabenkoru, line 314: I think you don't need a separate method. Just create a local variable like this:
`boolean isExistingInApp = storage.getMessage(message.getMessageId()) != null`

vbabenkoru: If a boolean is only used within a single method, I think it's easier to define it inline rather than declaring a separate method.

Ayyanchira (approved):

iterableapi/src/test/java/com/iterable/iterableapi/IterableInAppManagerTest.java
```diff
@@ -382,12 +384,28 @@ public void testInAppAutoDisplayPause() throws Exception {
         verify(inAppHandler, times(1)).onNewInApp(inAppMessageCaptor.capture());
     }
 
+    @Test
+    public void testMessagePersistentReadStateFromServer() throws Exception {
+        // load the in-app that has not been synchronized with the server yet (read state is set to false)
+        dispatcher.enqueueResponse("/inApp/getMessages", new MockResponse().setBody(IterableTestUtils.getResourceString("inapp_payload_inbox_read_state_1.json")));
+        IterableInAppManager inAppManager = IterableApi.getInstance().getInAppManager();
+        inAppManager.syncInApp();
+
```
vbabenkoru, line 393: I think it'd be good to add an assert to verify that `read` was false initially.

vbabenkoru (approved):
Approved with one minor comment.

PR #332, opened 2021-03-28
[MOB-2762] - Bug fix - In App read state

Ayyanchira:
## 🔹 Jira Ticket(s) if any

https://iterable.atlassian.net/browse/MOB-2762

## ✏️ Description

1. Dont show inapp (process message) right on switching to foreground. Wait for sync to happen and it will trigger scheduleProcessing automatically.
2. Schedule Processing is manually called if the app comes to foreground within 60 seconds of its last sync.
3. Schedule processing is also called when getMessage call fails. It compensates for network being offline

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppManager.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppManager.java
@@ -428,7 +428,6 @@ public class IterableInAppManager implements IterableActivityMonitor.AppStateCal
 
     @Override
     public void onSwitchToForeground() {
-        scheduleProcessing();
         if (IterableUtil.currentTimeMillis() - lastSyncTime > MOVE_TO_FOREGROUND_SYNC_INTERVAL_MS) {
             syncInApp();
         }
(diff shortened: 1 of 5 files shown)


iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppManager.java
```diff
@@ -428,7 +428,6 @@ private void handleIterableCustomAction(String actionName, IterableInAppMessage
 
     @Override
     public void onSwitchToForeground() {
-        scheduleProcessing();
```
vbabenkoru, line 431: Are you sure this won't have any adverse effects?

vbabenkoru: Do we always call `scheduleProcessing` after updating the queue?

Ayyanchira: As you mentioned. If network fails on the app, it will not be able to show an inapp by itself. It will rely on network availability.

Ayyanchira: Made some more changes. 
Now, 
Schedule Processing is manually called if the app comes to foreground within 60 seconds of its last sync.
Schedule processing is also called when getMessage call fails. It compensates for network being offline

vbabenkoru (approved):

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppManager.java
vbabenkoru: Awesome!

PR #339, opened 2021-04-15
[MOB-2579] - Maven Push essentials

Ayyanchira:
## 🔹 Jira Ticket(s) if any

https://iterable.atlassian.net/browse/MOB-2579

## ✏️ Description

1. Added maven-push.gradle which does the essential job of generating POM files and signing and uploading.
NOTE: This assumes that developer now has GPG key, password and location of the file store in local.properties. Refer here: http://sorcersoft.org/project/site/gradle/userguide/signing_plugin.html

2. Updated existing gradle file to include maven-push as a part of its tasks

3. gradle.properties now has common variables across api and api-ui project

4. Gitignore changes to not include pom files generated

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,6 +1,7 @@
 // Top-level build file where you can add configuration options common to all sub-projects/modules.
 buildscript {
     ext.kotlin_version = '1.4.32'
+    ext.mavenPublishEnabled = true
     repositories {
         google()
         jcenter()
--- a/gradle.properties
+++ b/gradle.properties
@@ -18,3 +18,20 @@
 # org.gradle.parallel=true
 android.enableJetifier=true
 android.useAndroidX=true
+
+VERSION_CODE=2
+GROUP=com.iterable
+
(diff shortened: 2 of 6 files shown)


gradle.properties
```diff
@@ -17,4 +17,21 @@
 # http://www.gradle.org/docs/current/userguide/multi_project_builds.html#sec:decoupled_projects
 # org.gradle.parallel=true
 android.enableJetifier=true
-android.useAndroidX=true
\ No newline at end of file
+android.useAndroidX=true
+
+VERSION_CODE=2
```
vbabenkoru, line 22: What does a version code mean? Is this something that is going to change often?

build.gradle
```diff
@@ -1,6 +1,7 @@
 // Top-level build file where you can add configuration options common to all sub-projects/modules.
 buildscript {
     ext.kotlin_version = '1.4.32'
+    ext.mavenPublishEnabled = true
```
vbabenkoru, line 4: Nice 👍

gradle.properties
vbabenkoru: It looks like it's not even referenced in the Maven script.. do we need it?

vbabenkoru: It looks like it's not even referenced in the Maven script.. do we need it?

Ayyanchira: Hmm.. we dont actually need it. I somehow thought its been used.

vbabenkoru (approved):

PR #340, opened 2021-04-28
[MOB-2662] Validating context existence before action execution

hoofmen:
## 🔹 Jira Ticket(s) if any

* [MOB-2662](https://iterable.atlassian.net/browse/MOB-2662)

## ✏️ Description

> Please provide a brief description of what this pull request does.

Validating context existence before trying to execute an action. If context does not exits, assign it to a pending state that will execute once context has been initialized

### DIFF
--- a/iterableapi/src/androidTest/java/com/iterable/iterableapi/IterableActionRunnerTest.java
+++ b/iterableapi/src/androidTest/java/com/iterable/iterableapi/IterableActionRunnerTest.java
@@ -81,7 +81,7 @@ public class IterableActionRunnerTest {
         actionData.put("type", "openUrl");
         actionData.put("data", "https://example.com");
         IterableAction action = IterableAction.from(actionData);
-        IterableActionRunner.pendingAction = new IterableActionRunner.PendingAction(action, IterableActionSource.APP_LINK);;
+        IterableActionRunner.pendingAction = new IterableActionRunner.PendingAction(action, IterableActionSource.APP_LINK);
         IterableActionRunner.processPendingAction(getApplicationContext());
 
         intended(allOf(hasAction(Intent.ACTION_VIEW), hasData("https://example.com")));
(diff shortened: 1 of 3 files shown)


iterableapi/src/androidTest/java/com/iterable/iterableapi/IterableActionRunnerTest.java
```diff
@@ -34,6 +36,7 @@
     @Before
     public void setUp() {
         Intents.init();
+        IterableActionRunner.pendingAction = null;
```
hoofmen, line 39: Needed as otherwise the static `IterableActionRunner` will have a pending action in between tests. Not sure if this could happen in real-life since the pending action should be cleared once context is initialized.

iterableapi/src/androidTest/java/com/iterable/iterableapi/IterableActionRunnerTest.java
```diff
@@ -55,6 +58,36 @@ public void testOpenUrlAction() throws Exception {
...
+    public void testOpenUrlActionWithoutContext() throws Exception {
+        IterableTestUtils.initIterableApi(null);
+        intending(anyIntent()).respondWith(new Instrumentation.ActivityResult(0, null));
+        JSONObject actionData = new JSONObject();
+        actionData.put("type", "openUrl");
+        actionData.put("data", "https://example.com");
+        IterableAction action = IterableAction.from(actionData);
+        assertNull(IterableActionRunner.pendingAction);
+        IterableActionRunner.executeAction(null, action, IterableActionSource.APP_LINK);
+
```
hoofmen, line 71: This test should make sure that w/o a context no VIEW should be opened and any action would be marked as pending. I could not find a way to validate that now VIEW should open.

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -307,7 +307,7 @@ public static void initialize(@NonNull Context context, @NonNull String apiKey,
                     sharedInstance.config.inAppDisplayInterval);
         }
         loadLastSavedConfiguration(context);
-        IterablePushActionReceiver.processPendingAction(context);
+        processPendingActions(context);
```
hoofmen, line 310: Extracted this to a separated method of for `pending actions`. Wondering if we might have other pending actions we're not covering, if so we might have to implement something more abstract like a collection of `pending actions`

vbabenkoru: Good idea!

PR #344, opened 2021-05-06
squash changes from stable

roninopf:
## 🔹 Jira Ticket(s) if any

* N/A

## ✏️ Description

* squashes and merges changes from `stable` to `master`, which is only the changelog


vbabenkoru:
Why squash and not merge?

roninopf:
> Why squash and not merge?

It would duplicate all of the cherry-picked PRs, I didn't think that was necessary, or all that sensible for the history.

vbabenkoru (approved):
Ah, right! That makes sense.

PR #345, opened 2021-05-12
[MOB-2899] - Health Monitoring for Offline Processor

Ayyanchira:
## 🔹 Jira Ticket(s) if any

https://iterable.atlassian.net/browse/MOB-2899

## ✏️ Description

1. Dont schedule for offline processing if row count exceeds max allowed number
2. Max offline operation is set to 1000
3. Added tests for Healthmonitoring and offline processing
4. DB error checks and fallback logic

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/HealthMonitor.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/HealthMonitor.java
@@ -1,0 +1,47 @@
+package com.iterable.iterableapi;
+
+public class HealthMonitor {
+    private static final String TAG = "HealthMonitor";
+
+    public HealthMonitor(IterableTaskStorage database) {
+        this.database = database;
+    }
+
+    public boolean canSchedule() {
+        IterableLogger.d(TAG, "canSchedule");
+
+        if (errored) {
+            return false;
+        }
+
+        return database.numberOfTasks() < IterableConstants.MAX_OFFLINE_OPERATION;
+    }
+
+    public boolean canProcess() {
+        IterableLogger.d(TAG, "canProcess");
+        return !errored;
+    }
+
+    public void onScheduleError() {
+        IterableLogger.w(TAG, "onScheduleError");
+
+    }
+
+    void onNextTaskError() {
+        IterableLogger.w(TAG, "onNextTaskError");
+        onError();
+    }
+
+    void onDeleteAllTasksError() {
+        IterableLogger.w(TAG, "onDeleteAllTasksError");
+        onError();
+    }
+
+    private void onError() {
+        errored = true;
+
+    }
+
+    private IterableTaskStorage database;
+    private boolean errored = false;
+}
--- a/iterableapi/src/main/java/com/iterable/iterableapi/HealthMonitor.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/HealthMonitor.java
@@ -1,31 +1,23 @@
 package com.iterable.iterableapi;
 
-import java.util.ArrayList;
-
-public class HealthMonitor implements HealthMonitorHandler{
+public class HealthMonitor implements IterableTaskStorage.IterableDatabaseStatusListeners {
     private static final String TAG = "HealthMonitor";
 
     private IterableTaskStorage database;
     private boolean errored = false;
-    private ArrayList<HealthMonitorHandler> healthMonitorListeners = new ArrayList<>();
+
     public HealthMonitor(IterableTaskStorage database) {
         this.database = database;
+        database.addDataBaseListener(this);
     }
 
     public boolean canSchedule() {
         IterableLogger.d(TAG, "canSchedule");
-
-        return database.isDatabaseReady() && database.numberOfTasks() < IterableConstants.MAX_OFFLINE_OPERATION;
-        //TODO: errored has to be set by DB errors being found during Taskstorage initialization or during other DB operations.
-//        if (errored) {
-//            return false;
-//        }
-//
-//        return database.numberOfTasks() < IterableConstants.MAX_OFFLINE_OPERATION;
+        return !errored;
     }
 
     public boolean canProcess() {
-        IterableLogger.d(TAG, "canProcess");
+        IterableLogger.d(TAG, "Health monitor can process: " + !errored);
         return !errored;
     }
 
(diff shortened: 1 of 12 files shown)


iterableapi/src/main/java/com/iterable/iterableapi/IterableTaskStorage.java
```diff
@@ -254,13 +255,24 @@ private IterableTask createTaskFromCursor(Cursor cursor) {
         return taskIds;
     }
 
+    long numberOfTasks() throws Exception {
+        if (!isDatabaseReady()) {
+            notifyDBNotReady();
+            throw new Exception("Database is not ready");
```
davidtruong, line 261: Consider using a IllegalStateException to make the exception more specific.

iterableapi/src/main/java/com/iterable/iterableapi/IterableTaskStorage.java
```diff
@@ -445,14 +457,87 @@ private boolean updateTaskWithContentValues(String id, ContentValues contentValu
 
     private boolean isDatabaseReady() {
         if (database == null) {
+            notifyDBNotReady();
```
davidtruong, line 460: What's the use case for calling both notifyDBNotReady and notifyDBError?

roninopf: That's definitely an error, from before we formalized the error notifications. Removed.

iterableapi/src/main/java/com/iterable/iterableapi/IterableConstants.java
```diff
@@ -235,6 +235,9 @@
     // Custom actions handled by the SDK
     public static final String ITERABLE_IN_APP_ACTION_DELETE    = "delete";
 
+    //Offline operation
+    public static final long MAX_OFFLINE_OPERATION              = 1000;
```
vbabenkoru, line 239: The name looks confusing.. What is 'max operation'?

iterableapi/src/main/java/com/iterable/iterableapi/HealthMonitor.java
```diff
@@ -0,0 +1,68 @@
+package com.iterable.iterableapi;
+
+public class HealthMonitor implements IterableTaskStorage.IterableDatabaseStatusListeners {
+    private static final String TAG = "HealthMonitor";
+
+    private boolean errored = false;
```
vbabenkoru, line 6: Naming: It's not clear what exactly errored and what this variable is for.

iterableapi/src/main/java/com/iterable/iterableapi/IterableTaskStorage.java
```diff
@@ -254,13 +255,24 @@ private IterableTask createTaskFromCursor(Cursor cursor) {
         return taskIds;
     }
 
+    long numberOfTasks() throws Exception {
```
vbabenkoru, line 258: Methods in Java are usually verbs: `getTasksCount`/`getNumberOfTasks`

iterableapi/src/main/java/com/iterable/iterableapi/HealthMonitor.java
```diff
@@ -0,0 +1,68 @@
...
+    public void isClosed() {
+        IterableLogger.v(TAG, "DB Closed notified to healthMonitor");
+        errored = true;
+    }
+
+    @Override
+    public void isReady() {
+        IterableLogger.v(TAG, "DB Ready notified to healthMonitor");
+        errored = false;
+    }
```
vbabenkoru, line 67: Do these have to be separate methods? Looks like they're doing the same thing. If the logic is the same, I think it's better to have one method, and if we need to have different messages in the log, a method parameter should be enough.

iterableapi/src/main/java/com/iterable/iterableapi/IterableTaskStorage.java
```diff
@@ -445,14 +457,86 @@ private boolean updateTaskWithContentValues(String id, ContentValues contentValu
 
     private boolean isDatabaseReady() {
```
vbabenkoru, line 458: Is `isDatabaseReady` the only place where we can detect DB errors? One of the failing tests was crashing in `createTask` when calling `database.insert`, and that wasn't being caught by the health monitor. I think we should catch errors in every method that calls the database and log that and flip the flag. The current implementation probably won't catch database corruption errors or low storage errors.

iterableapi/src/main/java/com/iterable/iterableapi/IterableTaskStorage.java
```diff
@@ -445,14 +457,86 @@ private boolean updateTaskWithContentValues(String id, ContentValues contentValu
...
+        void onDBError();
+
+        void isClosed();
+
+        void isReady();
+    }
+
+    private ArrayList<IterableDatabaseStatusListeners> databaseStatusListeners = new ArrayList<>();
+
+    void addDataBaseListener(IterableDatabaseStatusListeners listener) {
```
vbabenkoru, line 486: * `database` is one word, no need to capitalize as `DataBase`.
* "Database listener" sounds too generic, I think "database state listener" or "database error listener" would be more specific.

iterableapi/src/main/java/com/iterable/iterableapi/OfflineRequestProcessor.java
```diff
@@ -70,7 +78,7 @@ boolean isRequestOfflineCompatible(String baseUrl) {
     }
 }
 
-//Placeholder Taskschedular for testing purpose.
+//Mock TaskScheduler for testing purpose.
```
vbabenkoru, line 81: Why is this a mock? Looks like a code that's used in production.

iterableapi/src/main/java/com/iterable/iterableapi/HealthMonitor.java
Ayyanchira: Makes sense.. will name it databaseErrored

iterableapi/src/main/java/com/iterable/iterableapi/IterableConstants.java
Ayyanchira: Maximum number of offline operations supported?
hmm.. may be OFFLINE_OPERATION_THRESHOLD or
OFFLINE_LIMIT or
OFFLINE_OPERATIONS_LIMIT

iterableapi/src/main/java/com/iterable/iterableapi/IterableTaskStorage.java
Ayyanchira: Removed similar methods and now it only notifies for two states, Ready and Error

iterableapi/src/main/java/com/iterable/iterableapi/HealthMonitor.java
Ayyanchira: Changed it to notify only two states -> errored and ready

iterableapi/src/main/java/com/iterable/iterableapi/OfflineRequestProcessor.java
Ayyanchira: Removed the line

iterableapi/src/main/java/com/iterable/iterableapi/IterableTaskStorage.java
Ayyanchira: Added code to read the returned value when inserting into database and then requesting via online mode.
Also flagging Healthmonitor in such cases

davidtruong (approved):
LFTM

vbabenkoru (approved):

PR #349, opened 2021-06-03
[MOB-2912] - Firebase Crash Fix

Ayyanchira:
## 🔹 Jira Ticket(s) if any

https://iterable.atlassian.net/browse/MOB-2912

## ✏️ Description

1. Upgrading firebase to 20.3.0 to use FirebaseMessaging instead of FirebaseInstanceId to get Token.
2. Updated related methods.

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableFirebaseMessagingService.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableFirebaseMessagingService.java
@@ -5,11 +5,13 @@ import android.os.AsyncTask;
 import android.os.Bundle;
 import androidx.annotation.NonNull;
 
-import com.google.firebase.iid.FirebaseInstanceId;
+import com.google.android.gms.tasks.Tasks;
+import com.google.firebase.messaging.FirebaseMessaging;
 import com.google.firebase.messaging.FirebaseMessagingService;
 import com.google.firebase.messaging.RemoteMessage;
 
 import java.util.Map;
+import java.util.concurrent.ExecutionException;
 
 public class IterableFirebaseMessagingService extends FirebaseMessagingService {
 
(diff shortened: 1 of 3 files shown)


iterableapi/src/main/java/com/iterable/iterableapi/IterableFirebaseMessagingService.java
```diff
@@ -86,7 +88,16 @@ public static boolean handleMessageReceived(@NonNull Context context, @NonNull R
      * Call this from a custom {@link FirebaseMessagingService} to register the new token with Iterable
      */
     public static void handleTokenRefresh() {
-        String registrationToken = FirebaseInstanceId.getInstance().getToken();
+        String registrationToken;
+        try {
+            registrationToken = Tasks.await(FirebaseMessaging.getInstance().getToken());
```
vbabenkoru, line 93: This code is duplicated. It's better to move it to a class accessible from both places where it's used and call that instead.

roninopf (approved):

iterableapi/src/main/java/com/iterable/iterableapi/IterableFirebaseMessagingService.java
```diff
@@ -86,11 +88,25 @@ public static boolean handleMessageReceived(@NonNull Context context, @NonNull R
...
+        String registrationToken = getFirebaseToken();
         IterableLogger.d(TAG, "New Firebase Token generated: " + registrationToken);
         IterableApi.getInstance().registerForPush();
     }
 
+    public static String getFirebaseToken() {
+        String registrationToken = null;
+        try {
+            registrationToken = Tasks.await(FirebaseMessaging.getInstance().getToken());
+        } catch (ExecutionException e) {
```
davidtruong, line 100: Is there a reason why we are only catching these 2 specific exceptions? What happens if other exceptions occur?

davidtruong:
Other than the comment about the error handling, overall it looks good.

iterableapi/src/main/java/com/iterable/iterableapi/IterableFirebaseMessagingService.java
Ayyanchira: Will add a general exception handling as well. 👍
Tasks.await throws those two exception. Hence added these two only initially

davidtruong (approved):
LFTM

PR #353, opened 2021-06-10
[MOB - 2976] - Adding time stamp to notification

Ayyanchira:
## 🔹 Jira Ticket(s) if any

https://iterable.atlassian.net/browse/MOB-2976

## ✏️ Description

1. Removed the code which set the when time to 0. This disabled our notification to have timestamp.

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableNotificationHelper.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableNotificationHelper.java
@@ -140,7 +140,7 @@ class IterableNotificationHelper {
 
             notificationBuilder
                     .setSmallIcon(getIconId(context))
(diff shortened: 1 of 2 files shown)


iterableapi/src/main/java/com/iterable/iterableapi/IterableNotificationHelper.java
```diff
@@ -140,7 +140,7 @@ public IterableNotificationBuilder createNotification(Context context, Bundle ex
 
             notificationBuilder
                     .setSmallIcon(getIconId(context))
-                    .setTicker(applicationName).setWhen(0)
```
Ayyanchira, line 143: @davidtruong , Was there any reason we set the notification timing to 0? 
I saw the history of file and this code was there from the beginning

roninopf (approved):

PR #355, opened 2021-06-14
[MOB-2971] - Set Context method

Ayyanchira:
## 🔹 Jira Ticket(s) if any

https://iterable.atlassian.net/browse/MOB-2971

## ✏️ Description

This method `setContext` will be called from RN Android Application class to just keep context ready with Iterable. This way, by the time IterableSDK is initialized and starts processing request, IterableActivitymonitor will already be aware of activities on the screen.


roninopf (approved):

davidtruong (approved):
LFTM

PR #393, opened 2021-11-02
[MOB-3386] android session tracking

roninopf:
## 🔹 Jira Ticket(s) if any

* [MOB-3386](https://iterable.atlassian.net/browse/MOB-3386)

## ✏️ Description

* opens and slightly refactors inbox session tracking related code
* also creates `updateVisibleRows` which does the [equivalent to the iOS SDK's version](https://github.com/Iterable/swift-sdk/blob/0dd3ce2f4e0a6904c51a67007a79c366a15b4a51/swift-sdk/Internal/InboxImpressionTracker.swift#L36)


iterableapi/src/main/java/com/iterable/iterableapi/ImpressionData.java
```diff
@@ -0,0 +1,32 @@
+package com.iterable.iterableapi;
+
+import androidx.annotation.RestrictTo;
+
+import java.util.Date;
+
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+public class ImpressionData {
```
Ayyanchira, line 8: Different files are good to keep it tidy but ImpressionData and SessionManager can stay package private. This should still be accessible to RN bridge through RNIterableInternal class.

roninopf: I tried removing `public` but it caused errors inside the native Android SDK. This doesn't seem to be the right approach?

Ayyanchira (changes requested):
Added comment regarding ImpressionData being public

iterableapi/src/main/java/com/iterable/iterableapi/ImpressionData.java
roninopf: Got it to work, my SDKs were probably out of sync the last time I tried

Ayyanchira (approved):

PR #400, opened 2021-11-11
[MOB - 3703] - Sensitive log information

Ayyanchira:
## 🔹 Jira Ticket(s) if any

(https://iterable.atlassian.net/browse/MOB-XXXX)](https://iterable.atlassian.net/browse/MOB-3703)

## ✏️ Description

1. Removed logging of JWT keys and API keys from Headers and also from error case scenarios. 
2. Added checks to leverage config.loglevel only if App is in debug mode.

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableLogger.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableLogger.java
@@ -8,7 +8,7 @@ import android.util.Log;
 public class IterableLogger {
 
     public static void d(String tag, String msg) {
-        if (isLoggableLevel(Log.DEBUG)) {
+        if (isLoggable(Log.DEBUG)) {
             Log.d(tag, " 💚 " + msg);
         }
     }
@@ -59,7 +59,7 @@ public class IterableLogger {
 
     private static boolean isLoggable(int messageLevel) {
         boolean isDebug = ((IterableApi.getInstance().getMainActivityContext().getApplicationInfo().flags & IterableApi.getInstance().getMainActivityContext().getApplicationInfo().FLAG_DEBUGGABLE) != 0);
-        if(isDebug){
+        if (isDebug) {
             return isLoggableLevel(messageLevel);
         }
         // Log level will be set to WARNING and above if in release mode.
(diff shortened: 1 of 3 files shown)


iterableapi/src/main/java/com/iterable/iterableapi/IterableRequestTask.java
```diff
@@ -245,12 +245,19 @@ private static String buildHeaderString(HttpURLConnection urlConnection) {
         Iterator<?> headerKeys = urlConnection.getRequestProperties().keySet().iterator();
         while (headerKeys.hasNext()) {
             String key = (String) headerKeys.next();
-            headerString.append(key + " : " + urlConnection.getRequestProperties().get(key) + "\n");
+            if (isSensitive(key)) {
+                continue;
```
roninopf, line 249: Not needed for approval, but would it be more useful here to print something like `redacted` or `******` here instead of skipping it?

iterableapi/src/main/java/com/iterable/iterableapi/IterableLogger.java
```diff
@@ -57,6 +57,15 @@ public static void printInfo() {
         }
     }
 
+    private static boolean isLoggable(int messageLevel) {
+        boolean isDebug = ((IterableApi.getInstance().getMainActivityContext().getApplicationInfo().flags & IterableApi.getInstance().getMainActivityContext().getApplicationInfo().FLAG_DEBUGGABLE) != 0);
```
roninopf, line 61: Not needed for approval, should `isDebug` be split out as a universal utility function?

roninopf (approved):

iterableapi/src/main/java/com/iterable/iterableapi/IterableLogger.java
Ayyanchira: In fact we do have a global state in IterableAPI class which is referred in IterableLogger. However, I dont see `setDebugMode` called anywhere in the code. 

https://github.com/Iterable/iterable-android-sdk/blob/788e949c96e990b6f38bb4f3eecf6f495f6170f2/iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java#L183

This PR will add explicit check. But will give it a look if it can be optimized 👍

iterableapi/src/main/java/com/iterable/iterableapi/IterableRequestTask.java
Ayyanchira: adHoc redacting was not recommended by security team. So skipping that

vbabenkoru (approved):

PR #404, opened 2021-12-01
[MOB - 3754] - Handling Trampoline Restriction

Ayyanchira:
## 🔹 Jira Ticket(s) if any

[[MOB-3754]](https://iterable.atlassian.net/browse/MOB-3754)

## ✏️ Description

1. Introduced new Activity - `IterableTrampolineActivity` with transparent style which now replaces `IterablePushActionReceiver` as a starting point when user taps notification replacing.
2. Notification Builder creates PendingIntent from `getActivity` instead of `getBroadcast` to accommodate to above change. [Except for custom action with openApp false]
3. Data from notification is thus, now routed through `IterableTrampolineActivity `
4. ~~`IterableTrampolineActivity ` passes the data as is to `IterablePushActionReceiver` and dismisses itself`~~
5. `IterablePushActionReceiver` is still applicable for those actions which invoke custom actions with openApp flag unchecked. `IterableTrampolineActivity` will not be invoked in this flow.
6. `IterablePushNotificationUtil` class now performs handling the operations post tap of push notification/ push notification buttons. Both `IterableTrampolineActivity ` and `IterablePushActionReceiver` use this class to accomplish common functionality.

[MOB-3754]: https://iterable.atlassian.net/browse/MOB-3754?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/iterableapi/src/main/AndroidManifest.xml
+++ b/iterableapi/src/main/AndroidManifest.xml
@@ -23,7 +23,7 @@
         </receiver>
 
         <activity
-            android:name=".TrampolineActivity"
+            android:name=".IterableTrampolineActivity"
             android:exported="false"
             android:launchMode="singleInstance"
             android:theme="@style/TrampolineActivity.Transparent"/>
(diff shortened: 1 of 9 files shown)


davidtruong (approved):
LFTM

PR #418, opened 2022-01-06
[Issue #377] Android 12: Set flag PendingIntent.FLAG_IMMUTABLE

scaires:
## 🔹 Jira Ticket(s) if any

* #377

## ✏️ Description

After deploying iterable SDK 3.4.0 with the Kiva Android app, I noticed [Android 12 specific crashes](https://github.com/Iterable/iterable-android-sdk/issues/377#issuecomment-1006741121) in our crash logging caused by a missing PendingIntent flag. 

It appears as if the fixes for this issue in previous releases (https://github.com/Iterable/iterable-android-sdk/pull/380 and https://github.com/Iterable/iterable-android-sdk/pull/390) regressed in SDK 3.4.0 (a278c7d773a62a1e356ac47263b4e12bea2373d7), as the change to these files was not called out as intentional in https://github.com/Iterable/iterable-android-sdk/pull/413.

I restored both [IterableNotificationBuilder.java](https://github.com/Iterable/iterable-android-sdk/blob/master/iterableapi/src/main/java/com/iterable/iterableapi/IterableNotificationBuilder.java) and [IterableNotificationHelper.java](https://github.com/Iterable/iterable-android-sdk/blob/master/iterableapi/src/main/java/com/iterable/iterableapi/IterableNotificationHelper.java) to revision 9e0a29fa278f674f23f3b72b92c55e1b3fd2e646 as part of this PR.

Workaround for those encountering this issue until a fix is released: reverting to [SDK 3.3.9](https://github.com/Iterable/iterable-android-sdk/releases/tag/3.3.9) should restore the previous behavior.

Thank you!
~ SteveC @ Kiva

Edit: Fixed conflicts w/ master's Trampoline changes. The PendingIntent mutablity is now specified in IterableNotificationHelper so did not need to be changed.

PR #420, opened 2022-01-13
[MOB-3897] Add regression test for PendingIntent.FLAG_IMMUTABLE

vbabenkoru:
## 🔹 Jira Ticket(s) if any

* [MOB-3897](https://iterable.atlassian.net/browse/MOB-3897)

## ✏️ Description

Add a test to make sure `PendingIntent.FLAG_IMMUTABLE` is present in both the content intent and action (button) intents to prevent any future regressions due to this flag being missing.


vbabenkoru:
This is a matching test for the fix in #418.

PR #425, opened 2022-02-02
[MOB - 3970] - allow setShowBadge on notificationChannel

Ayyanchira:
## 🔹 Jira Ticket(s) if any

* [MOB-3970](https://iterable.atlassian.net/browse/MOB-3970)

## ✏️ Description

1. Introduced a notificationBadge boolean value in IterableConfig. This has default set to true. setting it to false will hide the nofitication badge on app icon.

Note:
Notification badge is applied to a channel. And this configuration can disable all the notification badge passed through Iterable.
It is a single boolean value. As Iterable supports multiple Android Notification Channels, it will make sense to enable and disable badging for selected channels.

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableNotificationHelper.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableNotificationHelper.java
@@ -272,6 +272,9 @@ class IterableNotificationHelper {
                 notificationChannel = new NotificationChannel(channelId, channelName, NotificationManager.IMPORTANCE_HIGH);
                 notificationChannel.setDescription(channelDescription);
                 notificationChannel.enableLights(true);
+
+                //TODO: Is it a good idea to have the value fetched from config like this? Or should we have the value passed through a utility method? which will basically do the same thing. OR store it in sharedPreferences? and load it next time
+                notificationChannel.setShowBadge(IterableApi.getInstance().config.notificationBadge);
             }
             return notificationChannel;
         }
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableConfig.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableConfig.java
@@ -70,6 +70,8 @@ public class IterableConfig {
      */
     final String[] allowedProtocols;
 
+    final boolean notificationBadge;
+
     private IterableConfig(Builder builder) {
         pushIntegrationName = builder.pushIntegrationName;
         urlHandler = builder.urlHandler;
@@ -82,6 +84,7 @@ public class IterableConfig {
         authHandler = builder.authHandler;
         expiringAuthTokenRefreshPeriod = builder.expiringAuthTokenRefreshPeriod;
         allowedProtocols = builder.allowedProtocols;
+        notificationBadge = builder.notificationBadge;
     }
 
     public static class Builder {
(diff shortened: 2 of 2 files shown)


iterableapi/src/main/java/com/iterable/iterableapi/IterableNotificationHelper.java
```diff
@@ -272,6 +272,9 @@ private NotificationChannel createNotificationChannel(String channelId, String c
                 notificationChannel = new NotificationChannel(channelId, channelName, NotificationManager.IMPORTANCE_HIGH);
                 notificationChannel.setDescription(channelDescription);
                 notificationChannel.enableLights(true);
+
+                //TODO: Is it a good idea to have the value fetched from config like this? Or should we have the value passed through a utility method? which will basically do the same thing. OR store it in sharedPreferences? and load it next time
```
davidtruong, line 276: Let's chat about this later. We have multiple ways of passing in configurations. This way as well as the Android Manifest.

iterableapi/src/main/java/com/iterable/iterableapi/IterableNotificationHelper.java
```diff
@@ -272,6 +272,9 @@ private NotificationChannel createNotificationChannel(String channelId, String c
...
+                notificationChannel.setShowBadge(IterableApi.getInstance().config.notificationBadge);
```
davidtruong, line 277: Unfortunately this won't work as is because the channel has already been created once on device and it is tied to a specific ChannelId.

That means for it to take affect on existing devices you would need to change the channelId to be different. 
IE: String channelId = context.getPackageName() + "_noBadge";

davidtruong (changes requested):
See change comments

iterableapi/src/main/java/com/iterable/iterableapi/IterableNotificationHelper.java
Ayyanchira: Wow.. would like to explore how passing configurations through Manifest would work.

iterableapi/src/main/java/com/iterable/iterableapi/IterableNotificationHelper.java
Ayyanchira: Hmm.. looks like it can work as long as the notificationChannel is not submitted to channel is submitted to [NotificationManager#createNotificationChannel(NotificationChannel)](https://developer.android.com/reference/android/app/NotificationManager#createNotificationChannel(android.app.NotificationChannel))

[Doc](https://developer.android.com/reference/android/app/NotificationChannel#setShowBadge(boolean))

But also one observation is ->
Once I - set it to false - , I had to uninstall and install the app with - set to true - to make the badges work again. Just changing the code and installing over the existing app did not make it work.

iterableapi/src/main/java/com/iterable/iterableapi/IterableNotificationHelper.java
davidtruong: https://support.iterable.com/hc/en-us/articles/115000331943-Setting-up-Android-Push-Notifications-#customizing-android-push-notifications

We'll also need to update our support docs with the change after it is released.

Ayyanchira:
Closing this one. Implementing Android Manifest approach

PR #431, opened 2022-02-22
[MOB-4068] - Reset jwtToken when logging out

Ayyanchira:
## 🔹 Jira Ticket(s) if any

* [MOB-4068](https://iterable.atlassian.net/browse/MOB-4068)

## ✏️ Description

For JWT enabled projects using Android SDK, a logout and login ends up disabling the user but not registering the user again(stopping them from getting push notification). Reason being the auth not reset because the authHandler kept returning the same authtoken and was not set again. Instead it should be set to null and then next time user logs in, the code can continue normally


davidtruong (approved):
Changes look good to me. We'll just need some tests later on to verify the functionality for the future.

PR #439, opened 2022-03-16
Checking duplicate messageIds before showing push

Ayyanchira:
## 🔹 Jira Ticket(s) if any

Hack Week - Avoid showing duplicate push messages

## ✏️ Description

1. Comma seperated message Ids stored in shared preference
2. Arraylist implementation where New messageIds added at last and removed from index0 to keep size less than 10
3. All this functionality is clubbed into the new method: `isDuplicateMessageId` which now returns true if messageId duplicate is found. Else returns false where it successfully stores the last push messageId.

Remaining:
1. Call trackDuplicate Api call when return value of true is returned from this new method.
2. return false and do not proceed with showing the notification and discard the notification object created so far

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableFirebaseMessagingService.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableFirebaseMessagingService.java
@@ -73,12 +73,13 @@ public class IterableFirebaseMessagingService extends FirebaseMessagingService {
                 if (isDuplicateMessageId(context, notificationBuilder.iterableNotificationData.getMessageId())) {
                     IterableLogger.d(TAG, "Notification is duplicate .Proceed to call dedeup API");
                     IterableApi.getInstance().trackDuplicateSend(notificationBuilder.iterableNotificationData.getMessageId());
+                    notificationBuilder = null;
                     return false;
                 }
-                if(BuildConfig.DEBUG && notificationBuilder.iterableNotificationData.getCampaignId() == 0) {
-                    IterableLogger.d(TAG, "Marking Push proof as duplicates. Proofs still continues to be shown on device. It wont be treated and skipped as duplicates");
-                    IterableApi.getInstance().trackDuplicateSend(notificationBuilder.iterableNotificationData.getMessageId());
-                }
+//                if(BuildConfig.DEBUG && notificationBuilder.iterableNotificationData.getCampaignId() == 0) {
+//                    IterableLogger.d(TAG, "Marking Push proof as duplicates. Proofs still continues to be shown on device. It wont be treated and skipped as duplicates");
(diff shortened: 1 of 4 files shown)


iterableapi/src/main/java/com/iterable/iterableapi/IterableFirebaseMessagingService.java
```diff
@@ -83,6 +97,46 @@ public static boolean handleMessageReceived(@NonNull Context context, @NonNull R
...
+            while (recentMessageIdArrayList.size() > 9) {
+                IterableLogger.d(TAG, "Removing old messageId..");
+                recentMessageIdArrayList.remove(0);
+            }
+        }
+        //Add new messageId to list
+        recentMessageIdArrayList.add(messageId);
+
+        // Traversing the ArrayList
+        StringBuilder str = new StringBuilder();
```
roninopf, line 122: What do you think of using a [Set of String](https://developer.android.com/reference/android/content/SharedPreferences#getStringSet(java.lang.String,%20java.util.Set%3Cjava.lang.String%3E))s to store and save to the `SharedPreferences`? That way you won't have to do this String manipulation and it translates to a more usable data structure.

iterableapi/src/main/java/com/iterable/iterableapi/IterableFirebaseMessagingService.java
```diff
@@ -83,6 +97,46 @@ public static boolean handleMessageReceived(@NonNull Context context, @NonNull R
...
+        ArrayList<String> recentMessageIdArrayList = new ArrayList<>();
+        if (recentMessageIds != null) {
+            //Only if there are previous messageIds, check for duplicates
+            IterableLogger.d(TAG, "No records of message ids in sharedPreference");
+            recentMessageIdArrayList = new ArrayList<>(Arrays.asList(recentMessageIds.split(",")));
+            if (recentMessageIdArrayList.contains(messageId)) {
+                IterableLogger.d(TAG, "Duplicate message id found");
+                return true;
+            }
+            while (recentMessageIdArrayList.size() > 9) {
```
roninopf, line 113: I think this should be 10, as in, we keep track of 10 IDs.

iterableapi/src/main/java/com/iterable/iterableapi/IterableFirebaseMessagingService.java
Ayyanchira: Set wont be able to give the right sequence, while in case comma seperated string, I was able to maintain the order.

iterableapi/src/main/java/com/iterable/iterableapi/IterableFirebaseMessagingService.java
Ayyanchira: It will be max 10 with this logic as this check happens before entering the value to the list. So this condition basically checks -> "If the list is greater than 9, remove the first one to make space for 10th one"

PR #440, opened 2022-03-19
Disable logging in production

UMFsimke:
## 🔹 Jira Ticket(s) if any

* None

## ✏️ Description

Disables logging in production. Per official [Android's documentation](https://developer.android.com/studio/publish/preparing#turn-off-logging-and-debugging) all logging should be disabled in production. Additionally, current logger is not HIPAA compliant as errors are logged by default and sensitive information might leak by mistake. Example of this would be a code in the `IterableActionRunner.java` where is a code:
```
IterableLogger.e(TAG, "Could not find activities to handle deep link:" + uri);
```

In case a deep link contains user ID or other PHI that string will be logged onto the device and anyone can read it with LogCat or by observing logs.

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableConfig.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableConfig.java
@@ -1,7 +1,8 @@
 package com.iterable.iterableapi;
 
 import androidx.annotation.NonNull;
-import android.util.Log;
+
+import com.iterable.iterableapi.util.LogLevel;
 
 /**
  *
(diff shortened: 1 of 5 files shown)


UMFsimke:
True, my mistake. I've added custom mapping to hide default ones and introduced a new `NONE` constant. There are various ways to solve this, I'm not sure what is the preferable one for your team. One could add a logger delegate, just like in the iOS SDK, and implementations of `DefaultLoggerDelegage` and `NoneLoggerDelegage`.

PaulWoitaschek:
Imo the correct approach would be to utilize
adb shell setprop

And use the systems isLoggable

UMFsimke:
That would require every developer to set **every Iterable's tag** on their own, including CI tools. Those tags will increase through time, it shouldn't be responsibility of other developers to take care of internal logic.

More over, wouldn't that be a global property for all apps? What if I use `adb shell setprop` on a stolen device, wouldn't that enable logging? Even though one could say that is almost impossible to happen HIPAA is very strict and complexity of the flow is not considered and security needs to be taken into the account.

PR #445, opened 2022-04-18
[MOB-4248] Support JSON-only in-app message rendering

pauljung14:
## 🔹 Jira Ticket(s) if any

* [MOB-4247](https://iterable.atlassian.net/browse/MOB-4247)
* [MOB-4248](https://iterable.atlassian.net/browse/MOB-4248)

## ✏️ Description

* Update Android SDK to support rendering of JSON-only in-app messages.
* Adds a new `renderJsonHandler` and updates the `IterableInAppManager`'s `showMessage` method to make use of the new handler whenever we want to show a JSON-only in-app message.
* Some refactoring also included

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppManager.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppManager.java
@@ -43,7 +43,6 @@ public class IterableInAppManager implements IterableActivityMonitor.AppStateCal
     private final IterableInAppStorage storage;
     private final IterableInAppHandler handler;
     private final IterableInAppDisplayer displayer;
-    private final IterableRenderJsonHandler renderJsonHandler;
     private final IterableActivityMonitor activityMonitor;
     private final double inAppDisplayInterval;
     private final List<Listener> listeners = new ArrayList<>();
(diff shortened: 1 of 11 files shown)


iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppManager.java
```diff
@@ -207,44 +211,64 @@ void reset() {
...
+                processClick(message, url.toString(), inAppLocation);
                 handleInAppClick(message, url);
                 lastInAppShown = IterableUtil.currentTimeMillis();
                 scheduleProcessing();
             }
-        })) {
+        };
+        // JSON-only web in-app messages have a payload but no html
+        if (messageCustomPayload.length() > 0 && messageHtml.length() == 0) {
+            renderJsonHandler.showMessage(messageCustomPayload, urlCallback);
```
davidtruong, line 246: It looks like renderJsonHandler and displayer have the same logic but a different implementation for showMessage. What if you create a base class for IterableInAppDisplayer and`IterableRenderJsonDisplayer` which just has the showMessage function. That way you have a single InAppDisplayer when you create the IterableInAppManager object and might would simplify this logic for the future.

pauljung14:
This pull request can be reopened at a point when we're ready to pick this back up.

PR #446, opened 2022-05-25
[MOB-4447] feature/MOB-4447-protocol-fix

davidtruong:
## 🔹 Jira Ticket(s) if any

* [MOB-4447](https://iterable.atlassian.net/browse/MOB-4447)

## ✏️ Description

> The list of protocols doesn't work on mParticle because it creates a new instance.
Changes the reference to use the shared instance.

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableUtil.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableUtil.java
@@ -285,7 +285,7 @@ class IterableUtil {
             return true;
         }
 
-        for (String allowedProtocol : IterableApi.getInstance().config.allowedProtocols) {
+        for (String allowedProtocol : IterableApi.sharedInstance.config.allowedProtocols) {
             if (urlProtocol.equals(allowedProtocol)) {
                 return true;
             }


iterableapi/src/main/java/com/iterable/iterableapi/IterableUtil.java
```diff
@@ -285,7 +285,7 @@ static boolean isUrlOpenAllowed(@NonNull String url) {
             return true;
         }
 
-        for (String allowedProtocol : IterableApi.getInstance().config.allowedProtocols) {
+        for (String allowedProtocol : IterableApi.sharedInstance.config.allowedProtocols) {
```
Ayyanchira, line 288: I wonder how this will work as getInstance() returns sharedInstance.

davidtruong: Apparently when integrating with mParticle the allowedProtocols don't show up. So something is unique about the configuration setup. This was my primary guess as to the cause since it is the only difference between this value and other configs that we have setup.

Ayyanchira (approved):

PR #447, opened 2022-05-25
[MOB-4439] add update email with auth token

roninopf:
## 🔹 Jira Ticket(s) if any

* [MOB-4439](https://iterable.atlassian.net/browse/MOB-4439)

## ✏️ Description

* adds `authToken` as a field that can be set when setting updating a user's email

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
@@ -582,6 +582,7 @@ private static final String TAG = "IterableApi";
                 failureHandler.onFailure("The Iterable SDK must be initialized with email or " +
                         "userId before calling updateEmail", null);
             }
+
             return;
         }
 
@@ -594,9 +595,9 @@ private static final String TAG = "IterableApi";
 
                 storeAuthData();
                 getAuthManager().requestNewAuthToken(false);
+
                 if (successHandler != null) {
                     successHandler.onSuccess(data);
-
                 }
             }
         }, failureHandler);
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
@@ -339,10 +339,6 @@ private static final String TAG = "IterableApi";
         });
     }
 
-    public void setEmail(@Nullable String email) {
-        setEmail(email, null);
-    }
-
     /**
      * Set user email used for API calls
      * Calling this or {@link #setUserId(String)} is required before making any API calls.
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
@@ -359,7 +359,8 @@ private static final String TAG = "IterableApi";
             return;
         }
 
-        onLogOut();
+        logoutPreviousUser();
+
         _email = email;
         _userId = null;
         storeAuthData();
@@ -383,7 +384,8 @@ private static final String TAG = "IterableApi";
             return;
         }
 
-        onLogOut();
+        logoutPreviousUser();
+
         _email = null;
         _userId = userId;
         storeAuthData();
@@ -1058,10 +1060,6 @@ private static final String TAG = "IterableApi";
         return _deviceId;
     }
 
-    private boolean isEitherUserIdOrEmailSet() {
-        return _email != null || _userId != null;
-    }
-
     private void storeAuthData() {
         try {
             SharedPreferences.Editor editor = getPreferences().edit();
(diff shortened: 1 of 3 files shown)


iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -558,7 +562,15 @@ public void trackPurchase(double total, @NonNull List<CommerceItem> items, @Null
      * @param newEmail New email
      */
     public void updateEmail(final @NonNull String newEmail) {
-        updateEmail(newEmail, null, null);
+        updateEmail(newEmail, null, null, null);
+    }
+
+    public void updateEmail(final @NonNull String newEmail, final @NonNull String authToken) {
```
davidtruong, line 568: Is this method ever used anywhere?

roninopf: It's intended to be usable by the customer, but it's not used internally, yes.

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -570,29 +578,30 @@ public void updateEmail(final @NonNull String newEmail) {
...
+
             return;
         }
 
         apiClient.updateEmail(newEmail, new IterableHelper.SuccessHandler() {
             @Override
             public void onSuccess(@NonNull JSONObject data) {
                 if (_email != null) {
-                    _email = newEmail;
+                    setEmail(newEmail, authToken);
```
Ayyanchira, line 597: Instead of locally assigning the email, by this, we will be going through setEmail flow where there will be logout and reregistration. I'm not sure if updateEmail should follow all that? 
Shouldnt changing the email and authToken locally be enough?

iterableapi/src/main/java/com/iterable/iterableapi/IterableAuthManager.java
```diff
@@ -34,35 +35,44 @@ public class IterableAuthManager {
...
+                                        queueExpirationRefresh(authToken);
+                                    }
+
+                                    IterableApi.getInstance().setAuthToken(authToken);
+                                    pendingAuth = false;
+                                    reSyncAuth();
+
+                                    if (onSuccess != null) {
+                                        onSuccess(authToken);
+                                    }
```
Ayyanchira, line 65: Success handler passed in authManager looks remote to this class's method and functionality. 
`successCallback` received from developers should be called when updateEmail as a module is successful and not when AuthManager's `onAuthTokenRequested` is successful. That way, it will have clear separation of concern. 
Also there is success and failure callbacks being introduced in AuthHandler.

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
roninopf: I don't think so, I wanted to ensure that any specific handlers related to changing the email and auth token would also fire. If we just set it, then things like the proper `getMessages` queue wouldn't happen.

iterableapi/src/main/java/com/iterable/iterableapi/IterableAuthManager.java
roninopf: This path won't get called with `updateEmail`, so it's already separated. This was added before we decided on the success/failure callbacks for `AuthHandler`, so I think I was actually fixing something that was missing back then?

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
Ayyanchira: messages can stay intact unless logout happens. Updating email looks like it can happen separately.

iterableapi/src/main/java/com/iterable/iterableapi/IterableAuthManager.java
Ayyanchira: Oh yes! UpdateEmail wont trigger this flow.. But it will place a handler through it inside the AuthManager. Yeah callbacks were missing but are being taken care of in the other PR.

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
roninopf: I realized we have to do it by variable anyways. Changed.

iterableapi/src/main/java/com/iterable/iterableapi/IterableAuthManager.java
roninopf: Feel free to change it over/overwrite it in the next PR!

Ayyanchira: Alright. Will have it modified in the other PR 👍🏼

Ayyanchira (approved):
LGTM. Will have the success handler adjusted in the PR where its been worked upon.

PR #463, opened 2022-07-25
[MOB-4641] - Add success and failure callback to AuthHandler

Ayyanchira:
## 🔹 Jira Ticket(s) if any

* [MOB-4641](https://iterable.atlassian.net/browse/MOB-4641)

## ✏️ Description

1. Added Success and Failure method to AuthHandler interface.
2. SuccessCallback is called when SDK receives the `authToken`. The success callback passes back the received authToken which developers can verify to see if its the correct one.
3. Failure callback is called when nothing is returned with possible throwable object which can be useful to understand the cause.


davidtruong (approved):
The pr looks good to add the new callbacks. Before we create a new release, let's add in the ability to set a new auth token for the existing logged in user.

davidtruong (approved):
LGTM

PR #467, opened 2022-08-20
[MOB-3876] - Try catch around network callback

Ayyanchira:
## 🔹 Jira Ticket(s) if any

* [MOB-3876](https://iterable.atlassian.net/browse/MOB-3876)

## ✏️ Description

`connectionManager.registerNetworkcallback` can throw Security exception on non Pixel device unless the manufacturer's fix it
Reference :https://issuetracker.google.com/issues/175055271

Until then, this try catch should avoid app crashes.

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableNetworkConnectivityManager.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableNetworkConnectivityManager.java
@@ -47,29 +47,35 @@ class IterableNetworkConnectivityManager {
         NetworkRequest.Builder networkBuilder = new NetworkRequest.Builder();
 
         if (connectivityManager != null) {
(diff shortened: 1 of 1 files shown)


vbabenkoru (approved):
Looks good.
The code around this, however, makes me think: if it can't register a listener, is it going to think that there is no connection? I assume this is used for offline mode. Could this cause it to stop making any calls? We might want to default `isConnected` to true I think..

Ayyanchira:
> Looks good.
> The code around this, however, makes me think: if it can't register a listener, is it going to think that there is no connection? I assume this is used for offline mode. Could this cause it to stop making any calls? We might want to default `isConnected` to true I think..

Great suggestion. That situation can totally happen. and not having it to true by default may just keep bloating the SDK with requests in DB. Will create a ticket for it 👍🏼

PR #469, opened 2022-09-13
[MOB-4549] keychain

roninopf:
## 🔹 Jira Ticket(s) if any

* [MOB-4549](https://iterable.atlassian.net/browse/MOB-4549)

## ✏️ Description

* creates a new `IterableKeychain` analogous to the one on the iOS SDK
* stores `email`, `userId` and `authToken` fields in that keychain
* migrates those fields that were previously in `SharedPreferences` to the keychain


iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -330,28 +341,55 @@ private String getDeviceId() {
...
+        _authToken = getKeychain().getAuthToken();
+
+        if (_authToken != null) {
+            getAuthManager().queueExpirationRefresh(_authToken);
+        }
+    }
+
+    private void updateSDKVersion() {
+        migrateAuthDataFromSharedPrefsToKeychain();
+    }
```
Ayyanchira, line 361: updateSDKVersion can be misleading...
How about:
checkForOutdatedKeychain()
checkForOldKeychain()
or something along that line?

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -330,28 +341,55 @@ private String getDeviceId() {
...
+        SharedPreferences prefs = getPreferences();
+        String sharedPrefsEmail = prefs.getString(IterableConstants.SHARED_PREFS_EMAIL_KEY, null);
+        String sharedPrefsUserId = prefs.getString(IterableConstants.SHARED_PREFS_USERID_KEY, null);
+        String sharedPrefsAuthToken = prefs.getString(IterableConstants.SHARED_PREFS_AUTH_TOKEN_KEY, null);
+
+        SharedPreferences.Editor editor = getPreferences().edit();
+
+        if (getKeychain().getEmail() == null && sharedPrefsEmail != null) {
+            getKeychain().saveEmail(sharedPrefsEmail);
+            editor.remove(IterableConstants.SHARED_PREFS_EMAIL_KEY);
```
Ayyanchira, line 373: For some reason email is still present in the old sharedPreferences...

Ayyanchira: May be it needs a commit() after remove

roninopf: Aha, that's probably right. Will add.

iterableapi/build.gradle
```diff
@@ -39,6 +39,7 @@ dependencies {
     api 'androidx.appcompat:appcompat:1.0.0'
     api 'androidx.annotation:annotation:1.0.0'
     api 'com.google.firebase:firebase-messaging:20.3.0'
+    implementation 'androidx.security:security-crypto-ktx:1.1.0-alpha03'
```
vbabenkoru, line 42: Do we really want to depend on an alpha version of a library in our SDK?

vbabenkoru: Also, do we really need this dependency? How much more difficult would it be if we didn't?

roninopf: I'm basing the inclusion on these pages:
https://developer.android.com/topic/security/data.md
https://developer.android.com/jetpack/androidx/releases/security

Should we use the 1.0.0 version?

I also would like to avoid adding yet another dependency, but I'm not sure how else to implement this without it.

vbabenkoru: The library is accessing lower-level APIs, so we could use them directly. Here is the source code:
https://android.googlesource.com/platform/frameworks/support/+/refs/heads/androidx-security-security-crypto-release/security/crypto/src/main/java/androidx/security/crypto

roninopf: Ooh, great find. I'll definitely make that change.

app/build.gradle
```diff
@@ -9,7 +9,7 @@ android {
 
     defaultConfig {
         applicationId "com.iterable.iterableapi.testapp"
-        minSdkVersion 16
+        minSdkVersion 23
```
Ayyanchira, line 12: Looks like a big bump to minSdkVersion. I couldn't find docs around EncryptedSharedPreferences needing minSdkVersion 23. What was the reference?

roninopf: I think it was for the cryptopgraphy libraries, but I'll see if this can be bumped back down after adding the source directly to our code base instead.

iterableapi/build.gradle
roninopf: @vbabenkoru I got started on manually putting in the libraries required, but it's leading to needing to insert several (tink, tink.proto which I'm still looking for, protobuf, that's as far as I've looked for now), do you think this still makes sense with this many required sources?

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -488,6 +527,8 @@ public static void initialize(@NonNull Context context, @NonNull String apiKey,
             sharedInstance.config = new IterableConfig.Builder().build();
         }
 
+        sharedInstance.updateSDKVersion();
```
davidtruong, line 530: Looks like updateSDKVersion is called on every initialize call. 
Do we want to store the existing SDK version and only update when it changes?

iterableapi/src/main/java/com/iterable/iterableapi/androidx.security.crypto/EncryptedSharedPreferences.java
```diff
@@ -0,0 +1,569 @@
+/*
```
davidtruong, line 1: Can you pull this library directly in the dependencies without including the source code?

Looks like there are public dependencies we can add: https://developer.android.com/jetpack/androidx/releases/security#security-crypto-1.0.0

davidtruong: Haha, Looks like Victor suggested the opposite of including the entire dependency on the crypto repo. Let me take a look at his suggestions there

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
roninopf: Ah, yeah, we should do exactly that. Will add. We do that with the iOS SDK too.

roninopf: I decided to do this as a post-release ticket (MOB-4916), it turns out we don't have a way to check on the current SDK's version number that I could easily find (since it's all inside config files, and not the source). I'm wary of doing that work before our next release, since it's also a small optimization rather than a bug fix or new feature.

iterableapi/src/main/AndroidManifest.xml
```diff
@@ -1,9 +1,10 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:tools="http://schemas.android.com/tools"
```
Ayyanchira, line 2: ^ is this one necessary? removing that also made the app run ok.

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -330,29 +341,85 @@ private String getDeviceId() {
...
-            IterableLogger.e(TAG, "Error while retrieving email/userId/authToken", e);
         }
+
+        if (_authToken != null) {
+            getAuthManager().queueExpirationRefresh(_authToken);
+        }
+    }
+
+    private void updateSDKVersion() {
+        IterableLogger.v(TAG, "jay debug - has encryption dependency: " + hasEncryptionDependency());
```
Ayyanchira, line 379: Can remove the debugging logs

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -330,29 +341,85 @@ private String getDeviceId() {
...
+            IterableLogger.v(TAG, "UPDATED: migrated authToken from SharedPreferences to IterableKeychain");
+        } else if (sharedPrefsAuthToken != null) {
+            editor.remove(IterableConstants.SHARED_PREFS_AUTH_TOKEN_KEY);
+        }
+
+        editor.apply();
+    }
+
+    private boolean hasEncryptionDependency() {
+        return Build.VERSION.SDK_INT >= 23;
```
Ayyanchira, line 422: Is API level - the only check we need to do? 
There was something which developer needs to add right? I dont remember what exactly it was, but do we need any checks around 'uses-library' for lower version?

iterableapi/src/main/java/com/iterable/iterableapi/IterableNetworkConnectivityManager.java
```diff
@@ -47,29 +47,35 @@ private void startNetworkCallback(Context context) {
...
+                        }
                     }
-                }
-            });
+                });
+            } catch (SecurityException e) {
+                // This security exception seems to be affecting few devices.
+                // More information here: https://issuetracker.google.com/issues/175055271?pli=1
+                IterableLogger.e(TAG, e.getLocalizedMessage());
+            }
```
Ayyanchira, line 78: Might have to rebase/sync to remove this change from this branch.

iterableapi/src/main/java/com/iterable/iterableapi/IterableKeychain.kt
```diff
@@ -0,0 +1,57 @@
+package com.iterable.iterableapi
+
+import android.content.Context
+import android.content.SharedPreferences
+import androidx.security.crypto.EncryptedSharedPreferences
+import androidx.security.crypto.MasterKeys
+
+class IterableKeychain {
```
Ayyanchira, line 8: Might have to add annotations of requiring higher APIs to this class or methods in it, so other classes and Junit tests can process it correctly?

Ayyanchira (commented):
Looking great.
Tested on emulator with API22, And API 33. Works well!
Migration is smooth, old files are deleted. Functionality for SDK < 23 intact 👍🏼 

Added nit pick comments and questions.
Can be approved once CI passes.

iterableapi/src/main/AndroidManifest.xml
roninopf: That was automatically added by Android Studio, so I think it'll get re-added if it's removed anyways.

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
roninopf: This is the only check needed in our SDK, yes. If the developer supports below API 23, there are specific lines to add, but that's not relevant to the SDK internally.

iterableapi/src/main/java/com/iterable/iterableapi/IterableKeychain.kt
roninopf: I think it might be fine, since testing will use different configs than customer apps.

roninopf:
Unit tests will be worked on in another ticket; I'll be merging this in to make this manageable.

PR #473, opened 2022-09-28
[MOB-4666] in-app in memory config option

roninopf:
## 🔹 Jira Ticket(s) if any

* [MOB-4666](https://iterable.atlassian.net/browse/MOB-4666)

## ✏️ Description

* adds an `IterableConfig` option to switch between storing on file (as before, and still default) or just in memory


Ayyanchira (commented):
Looking good so far. Will wait for the file deletion for approval.

evangreer91 (commented):
Looks good to me. The IterableInAppMemoryStorage method is not implemented, but I assume this is planned to be implemented in another PR.

roninopf:
> Looks good to me. The IterableInAppMemoryStorage method is not implemented, but I assume this is planned to be implemented in another PR.

No, it just has an empty constructor, since nothing else needs to be setup when in memory storage is used.

evangreer91 (approved):
LGTM

PR #475, opened 2022-10-19
[MOB-4986] - Stop redraw listener on webview load

Ayyanchira:
## 🔹 Jira Ticket(s) if any

* [MOB-4986](https://iterable.atlassian.net/browse/MOB-4986)

## ✏️ Description

> onPreDraw() may keep getting called even after the inapp is being displayed and loaded completely. In some instance, getHeight() returns different values after the page has loaded which makes it go to the top. Hence stopping the continuous predraw calls and resizeScript calls after webview is loaded and visible. Might not be a perfect solution but looks like a temporary work around for current implementation.


Ayyanchira:
Closing this PR. The issue is addressed adjusting other variable

PR #476, opened 2022-10-20
Mob 4998 connectivity manager offline mode improvements

Ayyanchira:
## 🔹 Jira Ticket(s) if any

* [MOB-4998](https://iterable.atlassian.net/browse/MOB-4998)

## ✏️ Description

> **Improvements in ConnectivityManager for offline mode.** 
-----
1. Default value to isConnected is set to true.

2. Extra conditions for when to set isConnected false.
Device can be connected to multiple network at once. Usually Wifi and cellular. Previous implementation would switch SDK mode to offline if either of the one got disconnected, thereby falsely picturing device being offline.
This implementation takes multiple network into consideration and saves them in a Set. Only if all the network are disconnected, only then `isConnected` is flipped to false.
onCapabilitiesChanged is another callback that can give even granular controls over if internet is actually present or not, but to keep it simple, this is a good improvement to existing functionality of offline mode.


iterableapi/src/main/java/com/iterable/iterableapi/IterableTaskRunner.java
```diff
@@ -87,7 +87,7 @@ public void onSwitchToBackground() {
 
     }
 
-    private void runNow() {
+    synchronized private void runNow() {
```
Ayyanchira, line 90: Making this method `synchronized` as it invokes DB operations to avoid any possible db lock issues

iterableapi/src/main/java/com/iterable/iterableapi/IterableNetworkConnectivityManager.java
```diff
@@ -44,14 +48,17 @@ private IterableNetworkConnectivityManager(Context context) {
     @RequiresApi(api = Build.VERSION_CODES.LOLLIPOP)
     private void startNetworkCallback(Context context) {
         ConnectivityManager connectivityManager = (ConnectivityManager) context.getSystemService(Context.CONNECTIVITY_SERVICE);
-        NetworkRequest.Builder networkBuilder = new NetworkRequest.Builder();
-
+        NetworkRequest networkRequest = new NetworkRequest.Builder().build();
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
+            isConnected = networkRequest.hasCapability(NetworkCapabilities.NET_CAPABILITY_INTERNET);
```
Ayyanchira, line 53: Leveraging new API if available which can give initial status of whether the connected network has internet or not.

iterableapi/src/main/java/com/iterable/iterableapi/IterableNetworkConnectivityManager.java
```diff
@@ -44,14 +48,17 @@ private IterableNetworkConnectivityManager(Context context) {
     @RequiresApi(api = Build.VERSION_CODES.LOLLIPOP)
     private void startNetworkCallback(Context context) {
         ConnectivityManager connectivityManager = (ConnectivityManager) context.getSystemService(Context.CONNECTIVITY_SERVICE);
-        NetworkRequest.Builder networkBuilder = new NetworkRequest.Builder();
-
+        NetworkRequest networkRequest = new NetworkRequest.Builder().build();
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
+            isConnected = networkRequest.hasCapability(NetworkCapabilities.NET_CAPABILITY_INTERNET) ? true : false;
```
pauljung14, line 53: Doesn't `networkRequest.hasCapability` already return boolean? Sounds like it does

If not, wondering if double-bang (!!) operator works in Java? Looks like a perfect case for it if so... it'd turn this to
```java
isConnected = !!networkRequest.hasCapability(NetworkCapabilities.NET_CAPABILITY_INTERNET)
```

pauljung14 (approved):

iterableapi/src/main/java/com/iterable/iterableapi/IterableNetworkConnectivityManager.java
Ayyanchira: 🤦‍♂️ good catch 😅

iterableapi/src/main/java/com/iterable/iterableapi/IterableNetworkConnectivityManager.java
```diff
@@ -3,21 +3,25 @@
...
 import androidx.annotation.RequiresApi;
 
 import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.Set;
 
 class IterableNetworkConnectivityManager {
     private static final String TAG = "NetworkConnectivityManager";
-    private boolean isConnected;
+    private boolean isConnected = true;
```
vbabenkoru, line 19: What if the device is offline and the app is launched? Would we be able to detect that?

iterableapi/src/main/java/com/iterable/iterableapi/IterableNetworkConnectivityManager.java
vbabenkoru: Have you verified that this actually works? Looks a bit weird that a thing called 'request' returns the current status of the connection. When I googled how to check connectivity, people seem to be doing `connectivityManager.getNetworkCapabilities`, but I couldn't find this approach.

iterableapi/src/main/java/com/iterable/iterableapi/IterableNetworkConnectivityManager.java
```diff
@@ -44,14 +48,17 @@ private IterableNetworkConnectivityManager(Context context) {
...
+            isConnected = networkRequest.hasCapability(NetworkCapabilities.NET_CAPABILITY_INTERNET);
+        }
         if (connectivityManager != null) {
             try {
-                connectivityManager.registerNetworkCallback(networkBuilder.build(), new ConnectivityManager.NetworkCallback() {
+                connectivityManager.registerNetworkCallback(networkRequest, new ConnectivityManager.NetworkCallback() {
                     @Override
                     public void onAvailable(@NonNull Network network) {
                         super.onAvailable(network);
+                        networkSet.add(network);
```
vbabenkoru, line 61: Does `onAvailable` get called immediately when you subscribe? Are we sure we'll get all the networks into `networkSet`?

iterableapi/src/main/java/com/iterable/iterableapi/IterableNetworkConnectivityManager.java
vbabenkoru: See documentation for the request vs. network capabilities object:
https://developer.android.com/reference/android/net/NetworkRequest
https://developer.android.com/reference/android/net/NetworkCapabilities

Ayyanchira: Verified that it works. Helpful for first time launch with no network.
Portion of video where it is working as expected :

https://user-images.githubusercontent.com/22951141/202002403-0b59ee5c-43be-481b-98c1-aebbf49bf08a.mov

There are more ways to have granular access of connection status with network having data or not (for cases like where wifi is connected and data is not available), but it can be done later if needed. 

This implementation covers multiple type of network device gets connected to and also gets rid one bug where disconnecting from one network triggered offline mode.

iterableapi/src/main/java/com/iterable/iterableapi/IterableNetworkConnectivityManager.java
Ayyanchira: With `networkRequest.hasCapability` getting called during app launch, it indeed gets the status of network status as not connected and triggering the offline mode. 
However, this API is only available to Android.P (API 28) and above. 

For others, it will treat it as online connected which will lead to missed events. 

But if we keep the default value of `isConnected` to `false`, this case will work even for older devices, given that the callback of `onNetworkConnected` is always triggered - (which is what I am consistently observing. Could not repro situation of how it can be not triggered during app launch)

iterableapi/src/main/java/com/iterable/iterableapi/IterableNetworkConnectivityManager.java
Ayyanchira: Thats the behavior I am consistently seeing which also makes me think of reverting the default `isConnected` from `true` to `false`.

iterableapi/src/main/java/com/iterable/iterableapi/IterableNetworkConnectivityManager.java
vbabenkoru: I only see a `false` value in the video. Does it return `true` when the network is available?

vbabenkoru: I'd just replace this with a check that uses `connectivityManager.getNetworkCapabilities`.

iterableapi/src/main/java/com/iterable/iterableapi/IterableNetworkConnectivityManager.java
```diff
@@ -37,21 +41,32 @@ private IterableNetworkConnectivityManager(Context context) {
...
+            connectivityManager = (ConnectivityManager) context.getSystemService(Context.CONNECTIVITY_SERVICE);
+            checkInternetAvailabilityOnActiveNetwork();
             startNetworkCallback(context);
         }
     }
 
+    private void checkInternetAvailabilityOnActiveNetwork() {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
+            Network activeNetwork = connectivityManager.getActiveNetwork();
+            isConnected = activeNetwork == null ? false : connectivityManager.getNetworkCapabilities(activeNetwork).hasCapability(NetworkCapabilities.NET_CAPABILITY_INTERNET);
```
vbabenkoru, line 53: I would also wrap all of this stuff in a `try...catch`. It should be safe enough, but this is Android, who knows if it's going to blow up on some obscure Android device. :D

Ayyanchira: Good idea 👍. Done

evantk91 (approved):
LGTM

Ayyanchira:
This can be merged once its verified working on older version of Android SDK

PR #487, opened 2023-01-11
[MOB-5365] update to set custom sound at notification channel

evangreer91:
## 🔹 Jira Ticket(s) if any

* [MOB-5365](https://iterable.atlassian.net/browse/MOB-5365)

## ✏️ Description

This pull request updates the custom sound to be set at the notification channel rather than at the notification builder. See [here](https://developer.android.com/reference/android/app/NotificationChannel#setSound(android.net.Uri,%20android.media.AudioAttributes)) for the documentation. This resolves a part of the issue, but further logic needs to be implemented to address the situation for when the custom sound is updated for the push notification template.

[MOB-5365]: https://iterable.atlassian.net/browse/MOB-5365?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Ayyanchira (approved):
Approving to be added in custom sound branch

PR #488, opened 2023-01-16
[MOB-5406] create notification channel for each sound

evangreer91:
## 🔹 Jira Ticket(s) if any

* [MOB-5406](https://iterable.atlassian.net/browse/MOB-5406)

## ✏️ Description

This pull request provides unique notification channels for each sound sent from the Iterable UI. There are checks to provide the default sound for edge cases provided by the custom sound box on the UI. The UI may be changed in the future to not except these edge cases.

[MOB-5406]: https://iterable.atlassian.net/browse/MOB-5406?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableNotificationHelper.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableNotificationHelper.java
@@ -455,6 +455,9 @@ class IterableNotificationHelper {
         String[] soundFile = soundName.split("\\.");
         soundName = soundFile[0];
 
+        IterableLogger.d(IterableNotificationBuilder.TAG, "sound file name: " + soundName);
+        IterableLogger.d(IterableNotificationBuilder.TAG, "sound file type: " + soundFile[1]);
+
         if (!soundName.equalsIgnoreCase(IterableConstants.DEFAULT_SOUND)) {
             int soundID = context.getResources().getIdentifier(soundName, IterableConstants.SOUND_FOLDER_IDENTIFIER, context.getPackageName());
             soundUri = Uri.parse(IterableConstants.ANDROID_RESOURCE_PATH + context.getPackageName() + "/" + soundID);
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableNotificationHelper.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableNotificationHelper.java
@@ -122,7 +122,7 @@ class IterableNotificationHelper {
             notificationBody = extras.getString(IterableConstants.ITERABLE_DATA_BODY);
             soundName = extras.getString(IterableConstants.ITERABLE_DATA_SOUND);
 
-            if (soundName.contains(".") && soundName.charAt(0) != '.') {
+            if (soundName != null && soundName.contains(".") && soundName.charAt(0) != '.') {
                 String[] soundFile = soundName.split("\\.");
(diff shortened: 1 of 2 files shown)


iterableapi/src/main/java/com/iterable/iterableapi/IterableNotificationHelper.java
```diff
@@ -116,12 +117,20 @@ public IterableNotificationBuilder createNotification(Context context, Bundle ex
                 return null;
             }
 
-            IterableNotificationBuilder notificationBuilder = new IterableNotificationBuilder(context, getChannelId(context));
             JSONObject iterableJson = null;
             title = extras.getString(IterableConstants.ITERABLE_DATA_TITLE, applicationName);
             notificationBody = extras.getString(IterableConstants.ITERABLE_DATA_BODY);
             soundName = extras.getString(IterableConstants.ITERABLE_DATA_SOUND);
 
+            if (soundName != null && soundName.contains(".") && soundName.charAt(0) != '.') {
```
Ayyanchira, line 125: The last portion - `&& soundName.charAt(0) != '.' `

Why to check one special use case where we check just the first letter being '.' ? Is it okay if first character is ','?

iterableapi/src/main/java/com/iterable/iterableapi/IterableNotificationHelper.java
```diff
@@ -116,12 +117,20 @@ public IterableNotificationBuilder createNotification(Context context, Bundle ex
...
+                String[] soundFile = soundName.split("\\.");
+                soundName = soundFile[0];
+                soundResourceId = context.getResources().getIdentifier(soundFile[0], "raw", context.getPackageName());
```
Ayyanchira, line 128: Always assume to have null conditions. The file may not be present at all. Should have the mechanism to fallback to default sound.

Ayyanchira: Above this, a null check has been done for soundName to be present in the json. Another null check if it actually resides in the disk will make it more robust.

iterableapi/src/main/java/com/iterable/iterableapi/IterableNotificationHelper.java
evangreer91: The special case is for taking care of edge cases such as '......'. Without the check, the application crashes due to the split call on line 126.

iterableapi/src/main/java/com/iterable/iterableapi/IterableNotificationHelper.java
evangreer91: This is what the soundResourceId is for. The soundResourceId is zero if the sound file is not present in the resources directory. Then, the soundResourceId is passed down and checked on line 441. If the soundResourceId is zero, the default sound uri is returned.

Ayyanchira: Also it will be great to log an error to say that sound was passed from iterable but SDK could not locate the file in res> folder.

Ayyanchira (approved):
Have some conditions & null checks to be added. Approving as the base is not master and WIP.
Should be taken before it gets merged to master.

PR #503, opened 2023-02-19
Mob 5277 enforced encryption

Ayyanchira:
## 🔹 Jira Ticket(s) if any

* [MOB-5277](https://iterable.atlassian.net/browse/MOB-5277)

## ✏️ Description

> Implementation of enforced encryption with fallback to unencrypted shared preference in case the initialization of EncryptedSharedPreference throws exception.
Introduced enforcedEncryption Key in config. This will enable app to enforce encryption. This may crash the app due to some devices reported to not work well with Android Crypto library. Unless required, enforced encryption is not recommended.


iterableapi/build.gradle
```diff
@@ -42,7 +42,7 @@ dependencies {
     api 'androidx.appcompat:appcompat:1.0.0'
     api 'androidx.annotation:annotation:1.0.0'
     api 'com.google.firebase:firebase-messaging:20.3.0'
-    implementation "androidx.security:security-crypto:1.0.0"
+    implementation "androidx.security:security-crypto:1.1.0-alpha04"
```
Ayyanchira, line 45: Updating the version which has newer method and replaces a deprecated method.

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -135,7 +134,11 @@ IterableAuthManager getAuthManager() {
     @NonNull
     IterableKeychain getKeychain() {
         if (keychain == null) {
-            keychain = new IterableKeychain(getMainActivityContext());
+            try {
```
Ayyanchira, line 137: One keychain class taking care of both either `SharedPreference` or `EncryptedSharedPreference` for storing PII.
Based on flag `enforcedEncryption` the constructor of this class can throw an exception if creation of encryptedSharedPreference fails and encryption cannot be compromised.

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -341,83 +344,23 @@ private String getDeviceId() {
...
+        _userId = getKeychain().getUserId();
+        _authToken = getKeychain().getAuthToken();
 
         if (_authToken != null) {
             getAuthManager().queueExpirationRefresh(_authToken);
         }
     }
 
     private void updateSDKVersion() {
-        if (hasEncryptionDependency()) {
```
Ayyanchira, line 379: updateSDKVersion is now blank. The condition of checking minSDKversion and migration is taken care of in IterableKeychain's constructor

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -341,83 +344,23 @@ private String getDeviceId() {
...
-            migrateAuthDataFromSharedPrefsToKeychain();
-        }
-    }
-
-    private void migrateAuthDataFromSharedPrefsToKeychain() {
```
Ayyanchira, line 384: This function is moved to `IterableKeychain` class

iterableapi/src/main/java/com/iterable/iterableapi/IterableConfig.java
```diff
@@ -77,6 +77,8 @@ public class IterableConfig {
      */
     final boolean useInMemoryStorageForInApps;
 
+    final boolean encryptionEnforced;
```
Ayyanchira, line 80: This is the flag that will be checked to see during a fallback to normal sharedPreference if it can be done or not. If this flag is true, the fallback will be skipped and a Runtimeexception will be thrown.

iterableapi/src/main/java/com/iterable/iterableapi/IterableKeychain.kt
```diff
@@ -2,10 +2,14 @@ package com.iterable.iterableapi
 
 import android.content.Context
 import android.content.SharedPreferences
+import android.os.Build
+import androidx.annotation.RequiresApi
 import androidx.security.crypto.EncryptedSharedPreferences
-import androidx.security.crypto.MasterKeys
+import androidx.security.crypto.MasterKey
```
Ayyanchira, line 8: This is the new module from new alpha version used to create masterkeys.

iterableapi/src/main/java/com/iterable/iterableapi/IterableKeychain.kt
```diff
@@ -14,15 +18,67 @@ class IterableKeychain {
     private val userIdKey = "iterable-user-id"
     private val authTokenKey = "iterable-auth-token"
 
-    constructor(context: Context) {
-        val masterKeyAlias = MasterKeys.getOrCreate(MasterKeys.AES256_GCM_SPEC)
+    private var encryptionEnabled = false
```
Ayyanchira, line 21: default value is `false` allowing fallback to normal sharedPreference in case creation of Encrypted one fails.

iterableapi/src/main/java/com/iterable/iterableapi/IterableKeychain.kt
```diff
@@ -14,15 +18,67 @@ class IterableKeychain {
...
+                    EncryptedSharedPreferences.PrefKeyEncryptionScheme.AES256_SIV,
+                    EncryptedSharedPreferences.PrefValueEncryptionScheme.AES256_GCM)
+                encryptionEnabled = true
+            } catch (e: Exception) {
+                if (encryptionEnforced) {
+                    //TODO: In memory Or similar solution needs to be implemented in future.
+                    IterableLogger.e(TAG, "Error creating EncryptedSharedPreferences", e)
+                    throw e.fillInStackTrace()
+                } else {
+                    sharedPrefs = context.getSharedPreferences(
```
Ayyanchira, line 50: This is where normal sharedPreference is created after checking for the flag.

iterableapi/src/main/java/com/iterable/iterableapi/IterableKeychain.kt
```diff
@@ -2,10 +2,14 @@ package com.iterable.iterableapi
...
+import android.os.Build
+import androidx.annotation.RequiresApi
 import androidx.security.crypto.EncryptedSharedPreferences
-import androidx.security.crypto.MasterKeys
+import androidx.security.crypto.MasterKey
 
 class IterableKeychain {
+
+    private val TAG = "IterableKeychain"
     private var sharedPrefs: SharedPreferences
```
Ayyanchira, line 13: a global sharedPreference which the class will refer to when setting and getting PII. Its generic at this point but can be Encrypted or non encrypted based on conditions

iterableapi/src/main/java/com/iterable/iterableapi/IterableKeychain.kt
Ayyanchira, line 21: a private flag for the class to understand if the encryption was enabled at the end or not. This is also used to proceed for migration from normal to encrypted shared preference if everything before was successful.

iterableapi/src/main/java/com/iterable/iterableapi/IterableKeychain.kt
```diff
@@ -54,4 +110,43 @@ class IterableKeychain {
             .putString(authTokenKey, authToken)
             .apply()
     }
+
+    @RequiresApi(api = Build.VERSION_CODES.M)
+    private fun migrateAuthDataFromSharedPrefsToKeychain(context: Context) {
```
Ayyanchira, line 115: migration code from IterableApi > updateSDKVersion part

roninopf (approved):

iterableapi/build.gradle
davidtruong: Is this only because the alpha release has what the available methods?

davidtruong: Just note the dependent functions in the comments in case something changes in the stable branch when they release 1.1.0. so we can easily identify the changes to modify if needed in the future.

Ayyanchira: > Is this only because the alpha release has what the available methods?

https://developer.android.com/reference/androidx/security/crypto/MasterKey
To use this library which has a wrapper for master key.
The old one used was already showing as deprecated.

davidtruong (approved):
LFTM

PR #529, opened 2023-03-09
[MOB-5785] - [Manager 2] - Sync remote embedded messages

Ayyanchira:
## 🔹 Jira Ticket(s) if any

* [MOB-5785](https://iterable.atlassian.net/browse/MOB-5785)

## ✏️ Description

1. update the local list of messages with remote messages instead of directly assigning remote message to local list.
2. Keep placeholder places for places to call delivery and deletion where new message is identified and old message is identified gotten deleted

[MOB-5785]: https://iterable.atlassian.net/browse/MOB-5785?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableEmbeddedManager.kt
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableEmbeddedManager.kt
@@ -180,10 +180,10 @@ public class IterableEmbeddedManager: IterableActivityMonitor.AppStateCallback{
 
         // Check for new messages and add them to the local list
         remoteMessageList.forEach {
-            if(!localMessageMap.containsKey(it.metadata.id)) {
-                //TODO: Make a call to the updateHandler to notify that the message has been added
+            if (!localMessageMap.containsKey(it.metadata.id)) {
                 localMessageList.add(it)
             }
+            //TODO: Make a call to the updateHandler to notify that the message has been added
         }
 
         // Check for messages in the local list that are not in the remote list and remove them
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableEmbeddedManager.kt
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableEmbeddedManager.kt
@@ -233,6 +233,7 @@ public class IterableEmbeddedManager: IterableActivityMonitor.AppStateCallback{
                 IterableLogger.v(TAG, "setting isSyncScheduled to true")
             } else {
                 IterableLogger.v(TAG, "Not scheduling a sync.. App is in background")
+                lastSync = autoFetchDuration.toLong()
             }
         }
     }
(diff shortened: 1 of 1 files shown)


iterableapi/src/main/java/com/iterable/iterableapi/IterableEmbeddedManager.kt
```diff
@@ -168,6 +168,44 @@ public class IterableEmbeddedManager: IterableActivityMonitor.AppStateCallback{
         scheduleSync()
     }
 
+    fun updateLocalMessages(remoteMessageList: List<IterableEmbeddedMessage>) {
+        IterableLogger.printInfo()
+
+        // Get local messages in a mutable list
+        val localMessageList = getEmbeddedMessages().toMutableList()
```
Ayyanchira, line 175: One thing to note is that the app does not stores the messages permanently. It will be in memory. In app delivery if tracked, might be misleading as every new app session has potential to track delivery of all messages again.

iterableapi/src/main/java/com/iterable/iterableapi/IterableEmbeddedManager.kt
```diff
@@ -168,6 +168,44 @@ public class IterableEmbeddedManager: IterableActivityMonitor.AppStateCallback{
...
+        val localMessageMap = mutableMapOf<String, IterableEmbeddedMessage>()
+        localMessageList.forEach {
+            localMessageMap[it.metadata.id] = it
+        }
+
+        // Check for new messages and add them to the local list
+        remoteMessageList.forEach {
+            if(!localMessageMap.containsKey(it.metadata.id)) {
+                //TODO: Make a call to the updateHandler to notify that the message has been added
+                localMessageList.add(it)
```
Ayyanchira, line 185: This is where we tally local list with remote list and add the new ones in local list one by one. This will allow us to call any track delivery calls along with other update handler calls here if we need do

iterableapi/src/main/java/com/iterable/iterableapi/IterableEmbeddedManager.kt
```diff
@@ -168,6 +168,44 @@ public class IterableEmbeddedManager: IterableActivityMonitor.AppStateCallback{
...
+
+        // Check for messages in the local list that are not in the remote list and remove them
+        val remoteMessageMap = mutableMapOf<String, IterableEmbeddedMessage>()
+        remoteMessageList.forEach {
+            remoteMessageMap[it.metadata.id] = it
+        }
+        val messagesToRemove = mutableListOf<IterableEmbeddedMessage>()
+        localMessageList.forEach {
+            if(!remoteMessageMap.containsKey(it.metadata.id)) {
+                messagesToRemove.add(it)
```
Ayyanchira, line 197: this is where local messages no longer valid are removed from local queue. Again like the above one, this can be a good place to call consume/ deleted/ update handlers for removed messages

iterableapi/src/main/java/com/iterable/iterableapi/IterableEmbeddedManager.kt
```diff
@@ -168,6 +168,44 @@ public class IterableEmbeddedManager: IterableActivityMonitor.AppStateCallback{
...
+        val localMessageList = getEmbeddedMessages().toMutableList()
+        val localMessageMap = mutableMapOf<String, IterableEmbeddedMessage>()
+        localMessageList.forEach {
+            localMessageMap[it.metadata.id] = it
+        }
+
+        // Check for new messages and add them to the local list
+        remoteMessageList.forEach {
+            if(!localMessageMap.containsKey(it.metadata.id)) {
+                //TODO: Make a call to the updateHandler to notify that the message has been added
```
roninopf, line 184: I was thinking we'd call the update handler after ALL the changes are done, so, right after line 205, where it's copied over to messages.

Ayyanchira: Correct, that makes sense.. thank you..

iterableapi/src/main/java/com/iterable/iterableapi/IterableEmbeddedManager.kt
roninopf: Nice catch, I missed that possibility too. I think that's an acceptable "known issue" for now.

roninopf (approved):

iterableapi/src/main/java/com/iterable/iterableapi/IterableEmbeddedManager.kt
evangreer91: LGTM

evangreer91 (approved):

PR #536, opened 2023-03-23
Prepare for 3.4.11

Ayyanchira:
## 🔹 Jira Ticket(s) if any

* [MOB-5894](https://iterable.atlassian.net/browse/MOB-5894)

## ✏️ Description

> Prepare for 3.4.11


bradumbaugh (approved):
@Ayyanchira My only question here is whether or not we want the hard wrap at 80 columns or not.. At least on the releases page in the repository, GitHub seems to render things better if you let the linds go long

PR #545, opened 2023-04-04
[MOB-5698] adds data center to config and associated unit tests

evantk91:
## 🔹 Jira Ticket(s) if any

* [MOB-5698](https://iterable.atlassian.net/browse/MOB-5698)

## ✏️ Description

- creates the IterableDataRegion enum which contains the associated endpoint
- adds the dataRegion value to the IterableConfig class and builder
- adds unit tests for testing the IterableConfig builder

[MOB-5698]: https://iterable.atlassian.net/browse/MOB-5698?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableDataRegion.kt
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableDataRegion.kt
@@ -1,0 +1,6 @@
+package com.iterable.iterableapi
+
+enum class IterableDataRegion(val endpoint: String) {
+    US("https://api.iterable.com/api/"),
+    EU("https://api.eu.iterable.com/api/")
+}
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableConfig.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableConfig.java
@@ -71,6 +71,11 @@ public class IterableConfig {
      */
     final String[] allowedProtocols;
 
+    /**
+     * Data region determining which data center and endpoints are used by the SDK.
+     */
+    final IterableDataRegion dataRegion;
+
     private IterableConfig(Builder builder) {
         pushIntegrationName = builder.pushIntegrationName;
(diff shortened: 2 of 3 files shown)


iterableapi/src/main/java/com/iterable/iterableapi/IterableDataRegion.kt
```diff
@@ -0,0 +1,6 @@
+package com.iterable.iterableapi
+
+enum class IterableDataRegion(val endpoint: String) {
```
Ayyanchira, line 3: Thinking if it will be better to have this enum within `IterableConstants` or `IterableRequestTask`.

Ayyanchira (commented):
Looking great overall.
Choice of string valued associated Enums are also great. Added some comments for discussion.
Really liked that first EUDC file changes came with its test methods! 🙌

iterableapi/build.gradle
```diff
@@ -91,6 +91,8 @@ if(hasProperty("mavenPublishEnabled")) {
 task javadoc(type: Javadoc) {
     source = android.sourceSets.main.java.srcDirs
     classpath += project.files(android.getBootClasspath().join(File.pathSeparator))
+
+    exclude '**/*.kt'
```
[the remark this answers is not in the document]
evantk91: Sounds good just added to pass the check

iterableapi/src/main/java/com/iterable/iterableapi/IterableDataRegion.kt
evantk91: I think it should go into IterableConstants and be used by IterableRequestTask

Ayyanchira (commented):
Changes look good Evan. Only the jacoco.exec file is unnecessary. Can as well add it in .gitignore file so its not added next time :)

PR #552, opened 2023-04-15
[MOB-5848] auto fetch boundary

evantk91:
## 🔹 Jira Ticket(s) if any

* [MOB-5848](https://iterable.atlassian.net/browse/MOB-5848)

## ✏️ Description

This pull request adds the boundary to the auto fetching interval to allow users to turn off auto fetching.

[MOB-5848]: https://iterable.atlassian.net/browse/MOB-5848?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


iterableapi/src/main/java/com/iterable/iterableapi/IterableEmbeddedManager.kt
```diff
@@ -200,31 +200,35 @@ public class IterableEmbeddedManager: IterableActivityMonitor.AppStateCallback{
 
     // region auto fetch functionality
 
-    fun scheduleSync() {
+    fun scheduleSync(autoFetchDuration: Double) {
```
Ayyanchira, line 203: `autoFetchDuration` is globally available in the file. Adding parameter to method signature would be a added cognitive overhead for readers.

Ayyanchira (commented):
Change of method signature might not be needed. Added comments

iterableapi/src/main/java/com/iterable/iterableapi/IterableEmbeddedManager.kt
evantk91: I agree

Issue #470, opened 2022-09-15
Custom Action Handler not invoked for specific flows


FDavidse20:
We have set up push notifications with Custom Action (for the Open action). 
We've implemented the necessary methods and followed the steps from the documentation for iOS and Android. 

For iOS it's working flawlessly, (test) push notifications can be received and interacted with. 
For Android it is working sort of. 
Push notifications can be received and interacted with. 

That means that the action handler is invoked when pressing a push notification. 
And the performSomeAction() is called.

`val config = IterableConfig.Builder()
            .setAutoPushRegistration(false)
            .setPushIntegrationName(pushIntegrationName)
            .setCustomActionHandler { action, _ ->
                action.type?.let {
                   //do something, open a detail page in the app based on the data send in the action
                  performSomeAction()
                }
                true
            }
            .build()
        IterableApi.initialize(
            application,
            apiKey,
            config
        )`

However, there is a very specific flow in which that function is no longer called.

Steps to reproduce:
- Run app
- log in (push notifications are enabled, email is set, etc.)
- receive push notification (doesn't matter if user is interacting with the app, or if the app is in the background)
- open push notification
- app opens (if not already opened) and custom action handler is executed and performSomeAction() called
- hard close the app
- receive push notification
- open push notification
- app opens and custom action handler is executed and performSomeAction() called
- Everything still going well, as expected, but now:
- receive push notification
- open push notification
- app opens (if not already opened) and custom action handler is NOT executed and performSomeAction() NOT called

in other words, after hard closing the app, the flow will only work perfectly once more, after that something is broken. 

I've tried all sorts of things debugging, but frankly haven't got a clue. 

I hope you can help me, thank you.
Filip

Ayyanchira:
Thanks for reporting this with detailed steps to repro @FDavidse20. I will check if I can reproduce this and will get back on this one...

FDavidse20:
> Thanks for reporting this with detailed steps to repro @FDavidse20. I will check if I can reproduce this and will get back on this one...

Hi @Ayyanchira, any updates on this issue?

dmitry-funda:
Hi @Ayyanchira , could you please address this issue?

tnortman-jabra:
Just commenting here to say I have the same issue +1

PR #434, opened 2022-03-08
[MOB - 4026] Fix - App not visible in recent apps

Ayyanchira:
## 🔹 Jira Ticket(s) if any

* [MOB-4026](https://iterable.atlassian.net/browse/MOB-4026)

## ✏️ Description

> 1. `singleInstance` launchmode of TrampolineActivity did not allow for parent activity to show in recent apps after destorying itself. Setting it to `singleTop` allows it.

PR #553, opened 2023-04-17
Ensure IterableTrampolineActivity always launched

tnortman-jabra:
## ✏️ Description

> Please provide a brief description of what this pull request does.

- Ensure `IterableTrampolineActivity` always launched to handle all notifications
- Reverts back to `launchMode="singleInstance"` to ensure that the activity is always launched
- Use `excludeFromRecents="true"` to ensure that the activity is not shown in the recent apps list

This originates from the current `launchModes` and `Intent` flags for `IterableTrampolineActivity`

When receiving a notification, the Iterable Android SDK builds up an `Intent` and stuffs it inside the notification. This intent is executed upon the user tapping the notification. This intent encapsulates the action to create a new Activity, `IterableTrampolineActivity`.

Without this code change, `IterableTrampolineActivity` will occasionally not be created as described in the linked issues. The `IterableTrampolineActivity` is the activity which handles passing off deep links to our application.

The `IterableTrampolineActivity` has a `launchMode="singleTop"`, which dictates to Android OS to only create a new activity if the activity doesn't exist on the top of our current back stack / task for our app. 

We can ensure `IterableTrampolineActivity` is created every time and thus links handled by having it's `launchMode` set to `singleInstance`. `singleInstance` tells Android OS to create a new activity, in a new task every time. 

This was actually how [Iterable Android SDK had functioned in the past, until someone filed a bug](https://github.com/Iterable/iterable-android-sdk/pull/434/files) that their app would show up twice in the recent apps. 

So to alleviate this issue as well, we simply add the `<activity>` attribute `excludeFromRecents` to tell Android OS to not show this activity in the list of recent apps ([your app tray](https://developer.android.com/guide/components/activities/recents))

https://github.com/Iterable/iterable-android-sdk/issues/470
https://github.com/Iterable/react-native-sdk/issues/394

### DIFF
--- a/iterableapi/src/main/AndroidManifest.xml
+++ b/iterableapi/src/main/AndroidManifest.xml
@@ -25,7 +25,8 @@
         <activity
             android:name=".IterableTrampolineActivity"
             android:exported="false"
-            android:launchMode="singleTop"
+            android:launchMode="singleInstance"
+            android:excludeFromRecents="true"
             android:theme="@style/TrampolineActivity.Transparent"/>
     </application>
 


nicopir:
Thanks for the fix!
When will it be merged? I'm having the same problem than the linked issue.

oelburk:
Looks awesome, any time frame on when it will be merged? Thanks!

joelbrostrom:
We need this in order to rewrite our customer's native apps into a cross platform flutter application.
Please give this some attention 🙏

Ayyanchira (approved):
Verified that this change adds a great value. 
Does not resolve all of the deep-linking issue in the SDK (still in progress) but it resolves situations of resuming the app at the last state and increased robust behavior when dealing with push notification.

PR #558, opened 2023-04-27
[MOB-5398] embedded delivery event tracking using custom track event

evantk91:
## 🔹 Jira Ticket(s) if any

* [MOB-5398](https://iterable.atlassian.net/browse/MOB-5398)

## ✏️ Description

This pull request implements embedded message delivery tracking using the custom track event. After the endpoint is ready, this will be connected to the embedded delivery endpoint.

[MOB-5398]: https://iterable.atlassian.net/browse/MOB-5398?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
@@ -238,13 +238,13 @@ public class IterableApi {
         apiClient.trackInAppDelivery(message);
     }
 
-    void trackEmbeddedDelivery(@NonNull IterableEmbeddedMessage message) {
+    void trackEmbeddedMessageReceived(@NonNull IterableEmbeddedMessage message) {
         if(!checkSDKInitialization()) {
             return;
         }
 
         if (message == null) {
-            IterableLogger.e(TAG, "trackEmbeddedDelivery: message is null");
+            IterableLogger.e(TAG, "trackEmbeddedMessageReceived: message is null");
             return;
         }
 
@@ -256,7 +256,7 @@ public class IterableApi {
             throw new RuntimeException(e);
         }
 
-        track("embedded delivery", dataFields);
+        track("embedded message received", dataFields);
 //        apiClient.trackEmbeddedDelivery();
     }
 
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
@@ -250,13 +250,13 @@ public class IterableApi {
 
         JSONObject dataFields = new JSONObject();
         try {
-            dataFields.put(IterableConstants.ITERABLE_EMBEDDED_MESSAGE_ID, message.getMetadata().getId());
-            dataFields.put(IterableConstants.ITERABLE_EMBEDDED_MESSAGE_TITLE, message.getElements().getTitle());
+            dataFields.put("name", "received");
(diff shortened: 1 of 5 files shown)


iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -238,6 +238,28 @@ void trackInAppDelivery(@NonNull IterableInAppMessage message) {
         apiClient.trackInAppDelivery(message);
     }
 
+    void trackEmbeddedDelivery(@NonNull IterableEmbeddedMessage message) {
```
roninopf, line 241: Let's call it `trackEmbeddedMessageReceived` for consistency across everywhere. It makes more sense than a "delivery" in how we use that term.

evantk91: I will make the change. Thanks!

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -238,6 +238,28 @@ void trackInAppDelivery(@NonNull IterableInAppMessage message) {
...
+
+        if (message == null) {
+            IterableLogger.e(TAG, "trackEmbeddedMessageReceived: message is null");
+            return;
+        }
+
+        JSONObject dataFields = new JSONObject();
+        try {
+            dataFields.put(IterableConstants.ITERABLE_EMBEDDED_MESSAGE_ID, message.getMetadata().getId());
+            dataFields.put(IterableConstants.ITERABLE_EMBEDDED_MESSAGE_TITLE, message.getElements().getTitle());
```
roninopf, line 254: I don't think we'll need the title here. The ID should be fine.

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -238,6 +238,28 @@ void trackInAppDelivery(@NonNull IterableInAppMessage message) {
...
+        } catch (JSONException e) {
+            throw new RuntimeException(e);
+        }
+
+        track("embedded message received", dataFields);
```
roninopf, line 259: Let's match up the data on this call, see [this](https://github.com/Iterable/swift-sdk/blob/ec1568d488318852a964ff6dee9021bad227ef8a/swift-sdk/Internal/EmbeddedMessagingManager.swift#L138). But replace "delivery" with "received" instead, I changed that in a currently open PR.

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
evantk91: Sounds good. I will go ahead with the change. I am just gonna use a string "name" instead of creating a constant since this is just a placeholder.

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
evantk91: Sounds good.

evantk91: Changing event name to "embedded-messaging" as well

roninopf (approved):

PR #562, opened 2023-05-03
[MOB-6135] - Check SDK initialization before API call

Ayyanchira:
## 🔹 Jira Ticket(s) if any

* [MOB-6135](https://iterable.atlassian.net/browse/MOB-6135)

## ✏️ Description

> Similar implementation to getInAppMessages where SDK initialization is checked before making a call to API

[MOB-6135]: https://iterable.atlassian.net/browse/MOB-6135?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
@@ -223,6 +223,20 @@ public class IterableApi {
     }
 
     /**
+     * Gets a list of Embedded Messages from Iterable; passes the result to the callback.
+     * To get list of messages as a list of EmbeddedMessages, use
+     * {@link IterableEmbeddedManager#getEmbeddedMessages()} instead
+     *
+     * @param onCallback
+     */
+    void getEmbeddedMessages(@NonNull IterableHelper.IterableActionHandler onCallback) {
+        if (!checkSDKInitialization()) {
+            return;
+        }
+        apiClient.getEmbeddedMessages(onCallback);
+    }
+
+    /**
      * Tracks in-app delivery events (per in-app)
      * @param message the in-app message to be tracked as delivered */
     void trackInAppDelivery(@NonNull IterableInAppMessage message) {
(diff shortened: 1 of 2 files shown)


iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -222,6 +222,20 @@ void getInAppMessages(int count, @NonNull IterableHelper.IterableActionHandler o
         apiClient.getInAppMessages(count, onCallback);
     }
 
+    /**
+     * Gets a list of Embedded Messages from Iterable; passes the result to the callback.
+     * To get list of messages as a list of EmbeddedMessages, use
+     * {@link IterableEmbeddedManager#getEmbeddedMessages()} instead
```
evantk91, line 228: Should note that this method is package private.

evantk91 (commented):
Left a comment

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
Ayyanchira: Yes. This way developers will get embedded messages from EmbeddedManger instead of IterableAPI class directly. This will keep the functionality similar to getInAppMessages from InAppManager instead of directly from the IterableApi class

evantk91: I am saying we should provide in the comments that this is package private for clarity.

Ayyanchira: Oh that might not be needed as long as its a void function. The above function has that probably because it was a public method previously.. 
But no harm in having it written for development purpose :) Will have it updated.

evantk91: Oh gotcha. Yeah I will leave it up to you. Approved.

evantk91 (approved):

Ayyanchira:
Checkstyle errors are gone. Other test is still failing. Need to check that in another PR.

PR #577, opened 2023-06-04
[MOB-6217] implement impression tracking functions

evantk91:
## 🔹 Jira Ticket(s) if any

* [MOB-6217](https://iterable.atlassian.net/browse/MOB-6217)

## ✏️ Description

- sets up impression tracking functions and connects them to start and end session functions

[MOB-6217]: https://iterable.atlassian.net/browse/MOB-6217?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/EmbeddedImpressionData.kt
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/EmbeddedImpressionData.kt
@@ -13,17 +13,4 @@ data class EmbeddedImpressionData(
     constructor(
         messageId: String
     ) : this(messageId, 0, 0.0f, null)
-
-    fun startImpression() {
-        this.start = Date()
-    }
-
-    fun endImpression() {
-        if(this.start != null) {
-            this.displayCount = this.displayCount?.plus(1)
-            this.duration =
-                this.duration?.plus((Date().time - this.start!!.time) / 1000.0).toFloat()
-            this.start = null
-        }
-    }
 }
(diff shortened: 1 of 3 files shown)


iterableapi/src/main/java/com/iterable/iterableapi/EmbeddedImpressionData.kt
```diff
@@ -0,0 +1,29 @@
...
+
+    fun endImpression() {
+        if(this.start != null) {
+            this.displayCount = this.displayCount?.plus(1)
+            this.duration =
+                this.duration?.plus((Date().time - this.start!!.time) / 1000.0).toFloat()
+            this.start = null
+        }
+    }
+}
```
Ayyanchira, line 29: Both session manager and impressionData having these methods look confusing. Lets use Impression Data as Data class and only use to store the parameters instead of performing a function inside.

evantk91: So get rid of the refactor and just put it in the session manager? I feel like it makes sense to have the functions that modify the data as members of the class.

evantk91: We can rename maybe? Since the startImpression in ImpressionData actually does the data manipulation.

evantk91: I added the impression data manipulation to the session manager and refactored accordingly

Ayyanchira:
LGTM

PR #578, opened 2023-06-06
[MOB - 6115] EM feature flagging

Ayyanchira:
## 🔹 Jira Ticket(s) if any

* [MOB-6115](https://iterable.atlassian.net/browse/MOB-6115)

## ✏️ Description

> Feature flagging Embedded messaging

[MOB-6115]: https://iterable.atlassian.net/browse/MOB-6115?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableApiClient.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableApiClient.java
@@ -232,6 +232,23 @@ class IterableApiClient {
         }
     }
 
+    void getEmbeddedMessages(@NonNull IterableHelper.SuccessHandler onSuccess, @NonNull IterableHelper.FailureHandler onFailure) {
+        JSONObject requestJSON = new JSONObject();
+
+        try {
+            addUserKeyToJson(requestJSON);
+            requestJSON.put(IterableConstants.KEY_PLATFORM, IterableConstants.ITBL_PLATFORM_ANDROID);
+            requestJSON.put(IterableConstants.ITBL_KEY_SDK_VERSION, IterableConstants.ITBL_KEY_SDK_VERSION_NUMBER);
+            requestJSON.put(IterableConstants.ITBL_SYSTEM_VERSION, Build.VERSION.RELEASE);
+            requestJSON.put(IterableConstants.KEY_PACKAGE_NAME, authProvider.getContext().getPackageName());
+            requestJSON.put("placementId", "0");
+
+            sendGetRequest(IterableConstants.ENDPOINT_GET_EMBEDDED_MESSAGES, requestJSON, onSuccess, onFailure);
+        } catch (JSONException e) {
+            e.printStackTrace();
+        }
+    }
+
     public void trackInAppOpen(@NonNull String messageId) {
         JSONObject requestJSON = new JSONObject();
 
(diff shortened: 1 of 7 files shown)


iterableapi/src/main/java/com/iterable/iterableapi/IterableEmbeddedManager.kt
```diff
@@ -115,36 +115,48 @@ public class IterableEmbeddedManager: IterableActivityMonitor.AppStateCallback{
...
+
+            } catch (e: JSONException) {
+                IterableLogger.e(TAG, e.toString())
             }
             lastSync = IterableUtil.currentTimeMillis()
             scheduleSync()
-        }
+        }, object : IterableHelper.FailureHandler {
+            override fun onFailure(reason: String, data: JSONObject?) {
+                if(reason.equals("SUBSCRIPTION_INACTIVE", ignoreCase = true)) {
```
Ayyanchira, line 147: @mprew97 , Will need your review here to see if backend is sending the exact same failure reason.

iterableapi/src/main/java/com/iterable/iterableapi/IterableApiClient.java
```diff
@@ -232,6 +232,23 @@ void getEmbeddedMessages(@NonNull IterableHelper.IterableActionHandler onCallbac
         }
     }
 
+    void getEmbeddedMessages(@NonNull IterableHelper.SuccessHandler onSuccess, @NonNull IterableHelper.FailureHandler onFailure) {
```
evantk91, line 235: Hm should this call also have the action handler? Seems like it is a requirement.

iterableapi/src/main/java/com/iterable/iterableapi/IterableApiClient.java
```diff
@@ -658,6 +675,10 @@ void sendGetRequest(@NonNull String resourcePath, @NonNull JSONObject json, @Nul
         getRequestProcessor().processGetRequest(authProvider.getApiKey(), resourcePath, json, authProvider.getAuthToken(), onCallback);
     }
 
+    void sendGetRequest(@NonNull String resourcePath, @NonNull JSONObject json, @NonNull IterableHelper.SuccessHandler onSuccess, @NonNull IterableHelper.FailureHandler onFailure) {
+        getRequestProcessor().processGetRequest(authProvider.getApiKey(), resourcePath, json, authProvider.getAuthToken(), onSuccess, onFailure);
```
evantk91, line 679: Here too. Should we pass onCallback?

evantk91 (commented):
Left a few comments.

iterableapi/src/main/java/com/iterable/iterableapi/IterableApiClient.java
Ayyanchira: I think its more like passing onSuccess and failure callback instead of one onCallback handler.

Ayyanchira: Only post requests had explicit success and failure callback handlers before.

iterableapi/src/main/java/com/iterable/iterableapi/IterableApiClient.java
Ayyanchira: What would be the use case for having actionHandler? Would success callback not capture that scenario?

evantk91: I see so we would leave the action handler call in case something with a callback is implemented in the future.

evantk91 (approved):

PR #591, opened 2023-07-01
[MOB-6443] remove polling logic and adds callbacks to embedded session manager

evantk91:
## 🔹 Jira Ticket(s) if any

* [MOB-6443](https://iterable.atlassian.net/browse/MOB-6443)

## ✏️ Description

- removes polling logic and fetching timer
- makes syncMessages function public
- adds onSwitchToForeground and onSwitchToBackground callbacks to the embedded session manager
- removes setting auto fetching interval from config

[MOB-6443]: https://iterable.atlassian.net/browse/MOB-6443?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/EmbeddedSessionManager.kt
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/EmbeddedSessionManager.kt
@@ -4,12 +4,11 @@ import androidx.annotation.RestrictTo
 import java.util.Date
 
 @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
-public class EmbeddedSessionManager() {
+public class EmbeddedSessionManager {
 
     private val TAG = "EmbeddedSessionManager"
 
     private var impressions: MutableMap<String, EmbeddedImpressionData> = mutableMapOf()
-    private var activityMonitor: IterableActivityMonitor? = null
 
     var session: IterableEmbeddedSession = IterableEmbeddedSession(
         null,
--- a/iterableapi/src/main/java/com/iterable/iterableapi/EmbeddedSessionManager.kt
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/EmbeddedSessionManager.kt
@@ -126,11 +126,10 @@ public class EmbeddedSessionManager(): IterableActivityMonitor.AppStateCallback 
     }
 
     override fun onSwitchToForeground() {
-        IterableLogger.d(TAG, "switch to foreground!!")
+        IterableLogger.printInfo()
     }
 
     override fun onSwitchToBackground() {
-        IterableLogger.d(TAG, "switch to background!!")
         endSession()
     }
 }
--- a/iterableapi/src/main/java/com/iterable/iterableapi/EmbeddedSessionManager.kt
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/EmbeddedSessionManager.kt
@@ -4,17 +4,12 @@ import androidx.annotation.RestrictTo
 import java.util.Date
 
 @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
-public class EmbeddedSessionManager(): IterableActivityMonitor.AppStateCallback {
+public class EmbeddedSessionManager() {
 
     private val TAG = "EmbeddedSessionManager"
 
     private var impressions: MutableMap<String, EmbeddedImpressionData> = mutableMapOf()
     private var activityMonitor: IterableActivityMonitor? = null
-
-    init{
-        activityMonitor = IterableActivityMonitor.getInstance()
-        activityMonitor?.addCallback(this)
-    }
 
     var session: IterableEmbeddedSession = IterableEmbeddedSession(
         null,
@@ -123,13 +118,5 @@ public class EmbeddedSessionManager(): IterableActivityMonitor.AppStateCallback 
             impressionData.start = null
         }
         return impressionData
-    }
-
-    override fun onSwitchToForeground() {
-        IterableLogger.printInfo()
-    }
-
-    override fun onSwitchToBackground() {
-        endSession()
     }
 }
--- a/iterableapi/src/main/java/com/iterable/iterableapi/EmbeddedSessionManager.kt
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/EmbeddedSessionManager.kt
@@ -4,11 +4,17 @@ import androidx.annotation.RestrictTo
 import java.util.Date
 
 @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
-public class EmbeddedSessionManager {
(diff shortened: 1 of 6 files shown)


iterableapi/src/main/java/com/iterable/iterableapi/IterableEmbeddedManager.kt
```diff
@@ -92,27 +77,13 @@ public class IterableEmbeddedManager: IterableActivityMonitor.AppStateCallback{
...
-    }
-
     //Gets the list of embedded messages in memory without syncing
     fun getEmbeddedMessages(): List<IterableEmbeddedMessage> {
         return localMessages
     }
 
     //Network call to get the embedded messages
-    private fun syncMessages() {
+    public fun syncMessages() {
```
anlinguist, line 86: From an app perspective, it would be helpful to know when syncMessages() finishes. I suggest adding a callback or  coroutine to syncMessages() so that the calling app can know when it finishes, and, if needed, update UIs.

Ayyanchira: Thats a great idea @anlinguist !

iterableapi/src/main/java/com/iterable/iterableapi/EmbeddedSessionManager.kt
```diff
@@ -4,11 +4,17 @@ import androidx.annotation.RestrictTo
 import java.util.Date
 
 @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
-public class EmbeddedSessionManager {
+public class EmbeddedSessionManager(): IterableActivityMonitor.AppStateCallback {
```
Ayyanchira, line 7: I think only one module - EmbeddedManager having a AppState callback could have been sufficient. Submodules and parents having access to same level of lifecycle callbacks can lead to many source of truths for what happens when app comes to foreground. 

May be we should try having manager class invoking session related methods when it sees app coming to foreground and background

iterableapi/src/main/java/com/iterable/iterableapi/IterableEmbeddedManager.kt
```diff
@@ -216,76 +182,15 @@ public class IterableEmbeddedManager: IterableActivityMonitor.AppStateCallback{
...
-
-    private fun canSyncEmbeddedMessages(): Boolean {
-        return getSecondsSinceLastFetch() >= autoFetchDuration
-    }
-
     // endregion
 
     // region IterableActivityMonitor.AppStateCallback
     override fun onSwitchToForeground() {
         IterableLogger.printInfo()
```
Ayyanchira, line 189:
```suggestion
        IterableLogger.printInfo()
        syncMessages()
```

Ayyanchira: We removed the code of autofetch, but the app should atleast sync consistently when it comes to foreground

iterableapi/src/main/java/com/iterable/iterableapi/EmbeddedSessionManager.kt
```diff
@@ -4,11 +4,17 @@ import androidx.annotation.RestrictTo
...
 
     private val TAG = "EmbeddedSessionManager"
 
     private var impressions: MutableMap<String, EmbeddedImpressionData> = mutableMapOf()
+    private var activityMonitor: IterableActivityMonitor? = null
+
+    init{
+        activityMonitor = IterableActivityMonitor.getInstance()
+        activityMonitor?.addCallback(this)
+    }
```
Ayyanchira, line 17: We can remove this portion if we decided to have app state callbacks being handled by EM Manager

iterableapi/src/main/java/com/iterable/iterableapi/EmbeddedSessionManager.kt
```diff
@@ -116,4 +124,12 @@ public class EmbeddedSessionManager {
...
         return impressionData
     }
+
+    override fun onSwitchToForeground() {
+        IterableLogger.printInfo()
+    }
+
+    override fun onSwitchToBackground() {
+        endSession()
+    }
```
Ayyanchira, line 134: Same like above

iterableapi/src/main/java/com/iterable/iterableapi/IterableEmbeddedManager.kt
```diff
@@ -21,39 +19,26 @@ public class IterableEmbeddedManager: IterableActivityMonitor.AppStateCallback{
...
-        autoFetchDuration = autoFetchInterval
         activityMonitor = IterableActivityMonitor.getInstance()
         activityMonitor?.addCallback(this)
-
-        postConstruction()
-    }
-
-    fun postConstruction() {
-        scheduleSync()
     }
```
Ayyanchira, line 41: We will need a call to sync message when we initialize the class as a first instance of app attempting to get the message for the user. 
This can be avoided if the app state callbacks are confirmed running on the first launch of the app. Sometimes, the AppState callbacks gets attached at a later point and the app could miss the first instance of `onForeground`

iterableapi/src/main/java/com/iterable/iterableapi/IterableEmbeddedManager.kt
```diff
@@ -92,27 +77,13 @@ public class IterableEmbeddedManager: IterableActivityMonitor.AppStateCallback{
...
-
-        IterableLogger.v(TAG, "Going to sync messages")
-
-        syncMessages()
-
-        IterableLogger.v(TAG, "Returning messages")
-
-        return localMessages
-    }
-
```
Ayyanchira, line 108: This is good. Felt like redundant and not exactly clear what it was doing. Thanks for removing :)

Ayyanchira (commented):
Shared some thoughts

iterableapi/src/main/java/com/iterable/iterableapi/EmbeddedSessionManager.kt
evantk91: I will update accordingly.

iterableapi/src/main/java/com/iterable/iterableapi/IterableEmbeddedManager.kt
evantk91: We should address this on a separate PR. Possibly post beta release.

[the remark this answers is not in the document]
evantk91: I shall remove.

Ayyanchira (approved):
LGTM

PR #601, opened 2023-07-19
[MOB-6490] toggle off impression tracking if no embedded messages

evantk91:
## 🔹 Jira Ticket(s) if any

* [MOB-6490](https://iterable.atlassian.net/browse/MOB-6490)
* [MOB-6610](https://iterable.atlassian.net/browse/MOB-6610)
* [MOB-6605](https://iterable.atlassian.net/browse/MOB-6605)

## ✏️ Description

This pull request includes the following:
- toggles off impression tracking call if there are no impressions to track
- updates message updated callback to align with iOS SDK
- updates naming of several items to align with iOS SDK
- renames getEmbeddedMessages to getMessages to align with iOS SDK

[MOB-6490]: https://iterable.atlassian.net/browse/MOB-6490?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[MOB-6610]: https://iterable.atlassian.net/browse/MOB-6610?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[MOB-6605]: https://iterable.atlassian.net/browse/MOB-6605?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/iterableapi/build.gradle
+++ b/iterableapi/build.gradle
@@ -61,6 +61,7 @@ dependencies {
     androidTestImplementation 'androidx.test.uiautomator:uiautomator:2.2.0'
     androidTestImplementation 'com.squareup.okhttp3:mockwebserver:4.2.2'
     androidTestImplementation 'org.mockito:mockito-android:3.0.0'
+    androidTestImplementation project(path: ':iterableapi')
 }
 
 ext {
--- a/iterableapi/build.gradle
+++ b/iterableapi/build.gradle
@@ -61,7 +61,6 @@ dependencies {
     androidTestImplementation 'androidx.test.uiautomator:uiautomator:2.2.0'
     androidTestImplementation 'com.squareup.okhttp3:mockwebserver:4.2.2'
     androidTestImplementation 'org.mockito:mockito-android:3.0.0'
-    androidTestImplementation project(path: ':iterableapi')
 }
 
 ext {
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableEmbeddedManager.kt
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableEmbeddedManager.kt
@@ -54,7 +54,6 @@ public class IterableEmbeddedManager : IterableActivityMonitor.AppStateCallback 
     //Add updateHandler to the list
     public fun addUpdateHandler(updateHandler: EmbeddedMessageUpdateHandler) {
         updateHandleListeners.add(updateHandler)
-        embeddedSessionManager.startSession(localMessages)
     }
 
     //Remove actionHandler from the list
@@ -85,6 +84,11 @@ public class IterableEmbeddedManager : IterableActivityMonitor.AppStateCallback 
     //Gets the list of embedded messages in memory without syncing
     fun getEmbeddedMessages(): List<IterableEmbeddedMessage> {
         return localMessages
+    }
+
+    fun reset() {
+        val emptyMessages = listOf<IterableEmbeddedMessage>()
+        updateLocalMessages(emptyMessages)
     }
 
     //Network call to get the embedded messages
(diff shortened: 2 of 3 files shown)


iterableapi/src/main/java/com/iterable/iterableapi/IterableEmbeddedManager.kt
```diff
@@ -54,7 +54,10 @@ public class IterableEmbeddedManager : IterableActivityMonitor.AppStateCallback
     //Add updateHandler to the list
     public fun addUpdateHandler(updateHandler: EmbeddedMessageUpdateHandler) {
         updateHandleListeners.add(updateHandler)
-        embeddedSessionManager.startSession()
+        if(localMessages.isNotEmpty()) {
+            embeddedSessionManager.startSession()
+            IterableLogger.d(TAG, "switch to foreground - start session")
+        }
```
Ayyanchira, line 60: AddupdateHandler is where the activities and fragment tell the manager that it is eager to listen to the events..
By design, it might not be the best place to start a session. Or may be it work out best. However, this part should be separate from a session starting checking if the local message is empty or not. 
The check of whether to start a session based on localmessages should be a part of startSession method.

iterableapi/src/main/java/com/iterable/iterableapi/IterableEmbeddedManager.kt
```diff
@@ -65,7 +68,10 @@ public class IterableEmbeddedManager : IterableActivityMonitor.AppStateCallback
     //Remove updateHandler from the list
     public fun removeUpdateHandler(updateHandler: EmbeddedMessageUpdateHandler) {
         updateHandleListeners.remove(updateHandler)
-        embeddedSessionManager.endSession()
+        if(localMessages.isNotEmpty()) {
+            embeddedSessionManager.endSession()
+            IterableLogger.d(TAG, "switch to background - end session")
+        }
```
Ayyanchira, line 74: Similar to above comments, endsession should get called regardless of localmessage queue. Messages could get deleted while the session is active. So it is important to not have external checks like message in memory check before calling endSession(). However, endSession can internally have its own checks like if the session had already started or not, if there were any impressions collected or not. And only then make the api request

Ayyanchira (changes requested):
Added some comments

iterableapi/src/main/java/com/iterable/iterableapi/IterableEmbeddedManager.kt
evantk91: Good point. I will restructure.

iterableapi/src/main/java/com/iterable/iterableapi/IterableEmbeddedManager.kt
evantk91: If we add the check to the start session internally we could need to pass in localMessages, do you think this is the best approach?

Ayyanchira: May be for a session to start we wont need to look at local message queue. We can start session even with no messages. Its just that endSession will not end up making calls as there will be no impressions accumulated

Ayyanchira (approved):
LGTM

iterableapi/build.gradle
```diff
@@ -61,6 +61,7 @@ dependencies {
     androidTestImplementation 'androidx.test.uiautomator:uiautomator:2.2.0'
     androidTestImplementation 'com.squareup.okhttp3:mockwebserver:4.2.2'
     androidTestImplementation 'org.mockito:mockito-android:3.0.0'
+    androidTestImplementation project(path: ':iterableapi')
```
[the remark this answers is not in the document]
evantk91: Yeah prob making test fail :) I will remove

PR #609, opened 2023-07-28
removes extra dependency

evantk91:
## 🔹 Jira Ticket(s) if any

## ✏️ Description

Removes extra dependency that may be causing tests to fail.


iterableapi/build.gradle
Ayyanchira, line 1: Yeah even I was a bit skeptical about this line. When we are locally testing, it asks to import this line sometimes so I let it merge in the master during the eudc merging I guess.

Ayyanchira (approved):
Should work even without this line.. just like before.

PR #612, opened 2023-08-01
Fix CI

vbabenkoru:
## 🔹 Jira Ticket(s) if any

## ✏️ Description

Fix CI on the encryption branch.


Ayyanchira (approved):
This fixes the CI error and also fixes a possible crash scenario when master key is built

PR #614, opened 2023-08-07
MOB-6129

devcsomnicg:
## 🔹 Jira Ticket(s) if any

* [MOB-MOB](https://iterable.atlassian.net/browse/MOB-6129)

## ✏️ Description

> Fixes for authhandler callback on setEmail/updateEmail without an auth token as a secondary parameter


devcsomnicg:
here is the video for the fix: https://drive.google.com/file/d/1OgV322cFJAp-p2-1nGnYuNmtw4Km3qkU/view?usp=sharing

Ayyanchira:
This PR is not needed as we set `pendingAuth` to false in the main flow as soon as we receive any kind of authtoken back from the app. Hence closing

PR #618, opened 2023-08-29
Ensure IterableTrampolineActivity always launched (#553)

Ayyanchira:
## 🔹 Jira Ticket(s) if any

#553 

## ✏️ Description

- Ensure IterableTrampolineActivity always launched to handle all notifications
- Reverts back to launchMode="singleInstance" to ensure that the activity is always launched
- Use excludeFromRecents="true" to ensure that the activity is not shown in the recent apps list


evantk91 (approved):
LGTM

PR #637, opened 2023-09-27
Catching Error along with Exception

Ayyanchira:
## 🔹 Jira Ticket(s) if any

Continuation on 6865. Add on changes

## ✏️ Description

1. Checking for Throwable which will collect both Exception and Error if at all Security library is crashing with a Error and not exception.
2. Updating the dependency to use latest alpha version. Alpha version 5 also seems to fix race condition during creation process. Reference - https://developer.android.com/jetpack/androidx/releases/security#1.1.0-alpha05
3. Added some warning logs if this situation arrives so developers can take some action if it occurs during debugging


davidtruong (approved):
LFTM

PR #643, opened 2023-10-16
[MOB-7047] update impression to include placement

evantk91:
## 🔹 Jira Ticket(s) if any

* [MOB-7047](https://iterable.atlassian.net/browse/MOB-7047)

## ✏️ Description

> This pull request updates the impression payload to include the placement id in the impression vs. the session.

[MOB-7047]: https://iterable.atlassian.net/browse/MOB-7047?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/EmbeddedImpressionData.kt
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/EmbeddedImpressionData.kt
@@ -13,5 +13,5 @@ data class EmbeddedImpressionData(
 ) {
     constructor(
         messageId: String
-    ) : this(messageId, "0",0, 0.0f, null)
+    ) : this(messageId, "",0, 0.0f, null)
 }
--- a/iterableapi/src/main/java/com/iterable/iterableapi/EmbeddedImpressionData.kt
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/EmbeddedImpressionData.kt
@@ -6,11 +6,12 @@ import java.util.Date
 @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
 data class EmbeddedImpressionData(
     val messageId: String,
+    val placementId: String,
     var displayCount: Int = 0,
     var duration: Float = 0.0f,
     var start: Date? = null
 ) {
     constructor(
         messageId: String
-    ) : this(messageId, 0, 0.0f, null)
+    ) : this(messageId, "0",0, 0.0f, null)
 }
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableConstants.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableConstants.java
@@ -44,7 +44,6 @@ public final class IterableConstants {
     public static final String KEY_TEMPLATE_ID          = "templateId";
     public static final String KEY_MESSAGE_CONTEXT      = "messageContext";
     public static final String KEY_MESSAGE_ID           = "messageId";
-    public static final String KEY_PLACEMENT_ID           = "placementId";
     public static final String KEY_TOKEN                = "token";
     public static final String KEY_TOTAL                = "total";
     public static final String KEY_UNSUB_CHANNEL        = "unsubscribedChannelIds";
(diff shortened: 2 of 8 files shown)


iterableapi/src/main/java/com/iterable/iterableapi/EmbeddedImpressionData.kt
```diff
@@ -6,11 +6,12 @@ import java.util.Date
...
     val messageId: String,
+    val placementId: String,
     var displayCount: Int = 0,
     var duration: Float = 0.0f,
     var start: Date? = null
 ) {
     constructor(
         messageId: String
-    ) : this(messageId, 0, 0.0f, null)
+    ) : this(messageId, "0",0, 0.0f, null)
```
Ayyanchira, line 16: Why 0? why not null?

iterableapi/src/main/java/com/iterable/iterableapi/IterableConstants.java
```diff
@@ -51,7 +51,6 @@ public final class IterableConstants {
     public static final String KEY_SUB_MESSAGE          = "subscribedMessageTypeIds";
     public static final String KEY_USER_ID              = "userId";
     public static final String KEY_USER                 = "user";
-    public static final String KEY_USER_KEY             = "userKey";
```
Ayyanchira, line 54: email/ userId again for now?

iterableapi/src/main/java/com/iterable/iterableapi/IterableEmbeddedManager.kt
```diff
@@ -87,10 +87,6 @@ public class IterableEmbeddedManager : IterableActivityMonitor.AppStateCallback
         IterableLogger.d(TAG, "$localMessages")
     }
 
-    fun reset() {
```
Ayyanchira, line 90: Oh looks like the implementation of reset was wrong intially. 
However, what happens when user logs out? We will need a reset functionality in EmbeddedManager which clears the in memory list of embedded messages and placementIds and also notifiy the listeners about with onUpdateMessages call

iterableapi/src/main/java/com/iterable/iterableapi/IterableEmbeddedPlacement.kt
```diff
@@ -86,8 +84,7 @@ data class IterableEmbeddedMessage (
 
 class EmbeddedMessageMetadata(
     var messageId: String,
-    //TODO: Remove this once the placementIDs are implemented in the backend
-    val placementId: String? = "",
+    val placementId: String = "",
```
Ayyanchira, line 87: Can the backend return `null` here ? In that case, SDK should keep this nullable.

Ayyanchira (commented):
Added some comments. Lets try to address those

iterableapi/src/main/java/com/iterable/iterableapi/IterableConstants.java
evantk91: This constant was not being used.

iterableapi/src/main/java/com/iterable/iterableapi/IterableEmbeddedManager.kt
evantk91: This is implemented correctly on the embedded branch. The listeners will be triggered when the local storage is set to an empty list.

iterableapi/src/main/java/com/iterable/iterableapi/EmbeddedImpressionData.kt
evantk91: I made it "0" because it is required. If we make it optional, we can make it nullable.

iterableapi/src/main/java/com/iterable/iterableapi/IterableEmbeddedPlacement.kt
evantk91: We can bring this up to mitch tomorrow.

iterableapi/src/main/java/com/iterable/iterableapi/EmbeddedImpressionData.kt
evantk91: Made it an empty string instead

iterableapi/src/main/java/com/iterable/iterableapi/IterableEmbeddedManager.kt
evantk91: I will update here.

iterableapi/src/main/java/com/iterable/iterableapi/IterableEmbeddedPlacement.kt
evantk91: Placement id will always be passed along from the backend

PR #644, opened 2023-10-17
[MOB-7179] adds comments to the getPlacementIds testing function

evantk91:
## 🔹 Jira Ticket(s) if any

* [MOB-7179](https://iterable.atlassian.net/browse/MOB-7179)

## ✏️ Description

> This pull request includes creating a getPlacementIds function that allows for users to get a list of all available placement ids for testing purposes.

[MOB-7179]: https://iterable.atlassian.net/browse/MOB-7179?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableEmbeddedManager.kt
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableEmbeddedManager.kt
@@ -14,6 +14,8 @@ public class IterableEmbeddedManager : IterableActivityMonitor.AppStateCallback 
 
     // region variables
     private var localMessages = mutableMapOf<String, List<IterableEmbeddedMessage>>()
+    private var placementIds = mutableListOf<String>()
+
     private var actionHandler: EmbeddedMessageActionHandler? = null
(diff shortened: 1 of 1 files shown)


iterableapi/src/main/java/com/iterable/iterableapi/IterableEmbeddedManager.kt
```diff
@@ -14,6 +14,8 @@ public class IterableEmbeddedManager : IterableActivityMonitor.AppStateCallback
 
     // region variables
     private var localMessages = mutableMapOf<String, List<IterableEmbeddedMessage>>()
+    private var placementIds = mutableListOf<String>()
```
Ayyanchira, line 17: We should discuss if having `placements` as object in this class. Can make sense to have just one list of placements which is a full fledged data class which maps to response of getMessages as is

Ayyanchira (commented):
Added some comments

PR #653, opened 2023-11-03
[MOB-7246] sets up click handler for implementing external/deep linking

evantk91:
## 🔹 Jira Ticket(s) if any

* [MOB-7246](https://iterable.atlassian.net/browse/MOB-7246)

## ✏️ Description

* sets up the `handleEmbeddedClick` function that can be called on the app side by a click listener
* the external links are handled by the action handler
* adds `EMBEDDED` to `IterableActionSource`
* updates IterableApi to provide context to external linking function

[MOB-7246]: https://iterable.atlassian.net/browse/MOB-7246?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
@@ -555,8 +555,8 @@ public class IterableApi {
         if (sharedInstance.embeddedManager == null) {
             sharedInstance.embeddedManager = new IterableEmbeddedManager(
                     sharedInstance,
-                    null,
-                    null);
+                    null
+            );
         }
 
         loadLastSavedConfiguration(context);
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
@@ -554,8 +554,7 @@ public class IterableApi {
 
         if (sharedInstance.embeddedManager == null) {
             sharedInstance.embeddedManager = new IterableEmbeddedManager(
-                    sharedInstance,
-                    null
+                    sharedInstance
             );
         }
 
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
@@ -555,6 +555,7 @@ public class IterableApi {
         if (sharedInstance.embeddedManager == null) {
             sharedInstance.embeddedManager = new IterableEmbeddedManager(
                     sharedInstance,
+                    sharedInstance.config.embeddedBaseUrl,
                     null,
                     null);
         }
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
@@ -553,7 +553,10 @@ public class IterableApi {
         }
 
         if (sharedInstance.embeddedManager == null) {
-            sharedInstance.embeddedManager = new IterableEmbeddedManager(null, null);
+            sharedInstance.embeddedManager = new IterableEmbeddedManager(
+                    sharedInstance,
+                    null,
+                    null);
         }
 
         loadLastSavedConfiguration(context);
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
@@ -555,7 +555,6 @@ public class IterableApi {
         if (sharedInstance.embeddedManager == null) {
             sharedInstance.embeddedManager = new IterableEmbeddedManager(
                     sharedInstance,
-                    sharedInstance.config.embeddedBaseUrl,
                     null,
                     null);
         }
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableConfig.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableConfig.java
@@ -112,6 +112,7 @@ public class IterableConfig {
         private String[] allowedProtocols = new String[0];
         private IterableDataRegion dataRegion = IterableDataRegion.US;
         private boolean useInMemoryStorageForInApps = false;
+
         public Builder() {}
 
         /**
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableConfig.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableConfig.java
@@ -82,6 +82,8 @@ public class IterableConfig {
      */
     final boolean useInMemoryStorageForInApps;
 
+    final String embeddedBaseUrl;
+
     private IterableConfig(Builder builder) {
         pushIntegrationName = builder.pushIntegrationName;
         urlHandler = builder.urlHandler;
@@ -96,6 +98,7 @@ public class IterableConfig {
         allowedProtocols = builder.allowedProtocols;
         dataRegion = builder.dataRegion;
         useInMemoryStorageForInApps = builder.useInMemoryStorageForInApps;
+        embeddedBaseUrl = builder.embeddedBaseUrl;
     }
 
     public static class Builder {
@@ -112,7 +115,7 @@ public class IterableConfig {
         private String[] allowedProtocols = new String[0];
         private IterableDataRegion dataRegion = IterableDataRegion.US;
         private boolean useInMemoryStorageForInApps = false;
-
+        private String embeddedBaseUrl = "";
         public Builder() {}
 
         /**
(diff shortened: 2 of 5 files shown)


iterableapi/src/main/java/com/iterable/iterableapi/IterableConfig.java
```diff
@@ -112,7 +115,7 @@ public static class Builder {
         private String[] allowedProtocols = new String[0];
         private IterableDataRegion dataRegion = IterableDataRegion.US;
         private boolean useInMemoryStorageForInApps = false;
-
+        private String embeddedBaseUrl = "";
```
Ayyanchira, line 118: Whats the purpose of baseUrl specially for Embedded?

evantk91: Yeah... Working with @vbabenkoru  on this. I wasn't sure how to prevent hard coding the URL in the code the handler is looking for when deep linking

evantk91: Probably a better approach

evantk91: Removed this.

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -553,7 +553,10 @@ public static void initialize(@NonNull Context context, @NonNull String apiKey,
         }
 
         if (sharedInstance.embeddedManager == null) {
-            sharedInstance.embeddedManager = new IterableEmbeddedManager(null, null);
+            sharedInstance.embeddedManager = new IterableEmbeddedManager(
+                    sharedInstance,
+                    null,
+                    null);
```
bradumbaugh, line 559: @evantk91 On the iOS side, I think this is where we add the config's `urlDelegate` to the embedded manager. Do we need to do that here?

iterableapi/src/main/java/com/iterable/iterableapi/IterableEmbeddedManager.kt
```diff
@@ -144,6 +125,11 @@ public class IterableEmbeddedManager : IterableActivityMonitor.AppStateCallback
         })
     }
 
+    fun handleEmbeddedClick(message: IterableEmbeddedMessage, buttonIdentifier: String?, clickedUrl: String?) {
+        IterableActionRunner.executeAction(context, IterableAction.actionOpenUrl(clickedUrl), IterableActionSource.EMBEDDED)
+        IterableApi.getInstance().trackEmbeddedClick(message, buttonIdentifier, clickedUrl)
```
bradumbaugh, line 130: @evantk91 I think we're removing the click handling right?

bradumbaugh (commented):
Couple of comments inline

iterableapi/src/main/java/com/iterable/iterableapi/IterableEmbeddedManager.kt
pauljung14: We do want to support open url handling, just not custom action handling afaik

evantk91: We are handling clicks just not tracking them here. @pauljung14 We are supporting only open url which the SDK treats as a type of `IterableAction`. The other custom actions will be supported after beta open.

iterableapi/src/main/java/com/iterable/iterableapi/IterableEmbeddedManager.kt
```diff
@@ -144,6 +125,10 @@ public class IterableEmbeddedManager : IterableActivityMonitor.AppStateCallback
         })
     }
 
+    fun handleEmbeddedClick(message: IterableEmbeddedMessage, buttonIdentifier: String?, clickedUrl: String?) {
+        IterableActionRunner.executeAction(context, IterableAction.actionOpenUrl(clickedUrl), IterableActionSource.EMBEDDED)
```
vbabenkoru, line 129: I'd also add support for `action://` URLs that get routed to the custom action handler, this part from the in-app manager code:
image

iterableapi/src/main/java/com/iterable/iterableapi/IterableEmbeddedManager.kt
```diff
@@ -13,16 +12,11 @@ public class IterableEmbeddedManager : IterableActivityMonitor.AppStateCallback
...
-    //TODO: See if coalescing all the messages into one list making one source of truth for local messages can be done.
-    private var _messages = MutableLiveData<List<IterableEmbeddedMessage>>()
-    val messages: LiveData<List<IterableEmbeddedMessage>>
-        get() = _messages
-
     private var localMessages: List<IterableEmbeddedMessage> = ArrayList()
-    private var actionHandler: EmbeddedMessageActionHandler? = null
     private var updateHandler: EmbeddedMessageUpdateHandler? = null
-    private var actionHandleListeners = mutableListOf<EmbeddedMessageActionHandler>()
     private var updateHandleListeners = mutableListOf<EmbeddedMessageUpdateHandler>()
```
vbabenkoru, line 17: Why do you need both `updateHandler` and `updateHandleListeners`?

iterableapi/src/main/java/com/iterable/iterableapi/IterableEmbeddedManager.kt
```diff
@@ -34,44 +28,31 @@ public class IterableEmbeddedManager : IterableActivityMonitor.AppStateCallback
 
     //Constructor of this class with actionHandler and updateHandler
     public constructor(
-        actionHandler: EmbeddedMessageActionHandler?,
+        iterableApi: IterableApi,
         updateHandler: EmbeddedMessageUpdateHandler?
```
vbabenkoru, line 32: Looks like this is always null? What's its purpose?

iterableapi/src/main/java/com/iterable/iterableapi/IterableEmbeddedManager.kt
evantk91: Custom action support will be added after beta open. We will be hiding the custom action field on the front end.

iterableapi/src/main/java/com/iterable/iterableapi/IterableEmbeddedManager.kt
evantk91: Yeah I was looking at that. The updateHandler's purpose is to let the embedded views know that messages were updated which triggers the view to get the most updated messages from the local storage. I kinda think we should rremove the `updateHandler` from the constructor. Akshay added this and I think the idea was to allow users to override this but maybe we should remove for now.

iterableapi/src/main/java/com/iterable/iterableapi/IterableEmbeddedManager.kt
evantk91: I think we should just keep `updateHandleListeners`. This is a list of interfaces that register views to the `embeddedManager`. When `onMessagesUpdated` in the `EmbeddedMessageUpdateHandler` is called, the views listed under `updateHandleListeners` are notified that messages are changed.

iterableapi/src/main/java/com/iterable/iterableapi/IterableEmbeddedManager.kt
vbabenkoru: idk I would add support for it now while we're at it, so that there is one less thing to do and less risk that we'll forget about it

iterableapi/src/main/java/com/iterable/iterableapi/IterableEmbeddedManager.kt
vbabenkoru: Yeah, up to you, if this code will evolve over time

iterableapi/src/main/java/com/iterable/iterableapi/IterableEmbeddedManager.kt
evantk91: I will remove since it is always null and not pulled from the config.

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
evantk91: No I dont think we do. it is handled differently in Android. `IterableUrlHandler` extends `MainActivity` and that is where we define the logic in the `handleIterableURL` method.

vbabenkoru:
Are there tests for the `IterableEmbeddedManager`? If not, are there any plans to add those?

evantk91:
> Are there tests for the `IterableEmbeddedManager`? If not, are there any plans to add those?

Yes I am currently working on adding tests in a future deliverable

PR #680, opened 2024-01-10
sync event if user is not created

hani-iterable:
## ✏️ Description

> We have implemented a call to the sync event method if a known user is not created.


hani-iterable:
We can close this PR as it includes all the changes related to the issue. You can find the updated changes in [PR](https://github.com/Iterable/iterable-android-sdk/pull/685).

PR #682, opened 2024-01-12
[MOB-7646] clean up of processing logic for new diff payload

evantk91:
## 🔹 Jira Ticket(s) if any

* [MOB-7646](https://iterable.atlassian.net/browse/MOB-7646)

## ✏️ Description

> This pull request adds logic to properly modify the local message list stored in the map based on the new payload. It also stores the current messages in a local list that is updated when messages are removed. The local message id list is also reset when the placements payload is empty. The use of shared preferences was removed since we want the full payload to be processed on every new app session. This was causing issues. 

I will need to add associated unit tests for the added logic which I will add on a separate PR due to time constraints.

[MOB-7646]: https://iterable.atlassian.net/browse/MOB-7646?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableApiClient.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableApiClient.java
@@ -248,7 +248,7 @@ class IterableApiClient {
         }
     }
 
-    void getEmbeddedMessages(@Nullable List<String> messageIds, @Nullable List<Long> placementIds, @NonNull IterableHelper.SuccessHandler onSuccess, @NonNull IterableHelper.FailureHandler onFailure) {
+    void getEmbeddedMessages(@Nullable List<String> currentMessageIds, @Nullable List<Long> placementIds, @NonNull IterableHelper.SuccessHandler onSuccess, @NonNull IterableHelper.FailureHandler onFailure) {
         JSONObject requestJSON = new JSONObject();
 
         try {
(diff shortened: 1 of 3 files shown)


iterableapi/src/main/java/com/iterable/iterableapi/IterableApiClient.java
```diff
@@ -258,9 +258,10 @@ void getEmbeddedMessages(@Nullable List<String> messageIds, @Nullable List<Long>
             requestJSON.put(IterableConstants.ITBL_SYSTEM_VERSION, Build.VERSION.RELEASE);
             requestJSON.put(IterableConstants.KEY_PACKAGE_NAME, authProvider.getContext().getPackageName());
 
-            SharedPreferences sharedPreferences = IterableApi.sharedInstance.getMainActivityContext().getSharedPreferences(IterableConstants.SHARED_PREFS_FILE, Context.MODE_PRIVATE);
-            Set<String> retrievedIdsSet = sharedPreferences.getStringSet(IterableConstants.SHARED_PREFS_CURRENT_EMBEDDED_MSGS, new HashSet<>());
-            String[] currentMessageIds = retrievedIdsSet.toArray(new String[0]);
+//            SharedPreferences sharedPreferences = IterableApi.sharedInstance.getMainActivityContext().getSharedPreferences(IterableConstants.SHARED_PREFS_FILE, Context.MODE_PRIVATE);
```
davidtruong, line 261: Why are these commented out? Shouldn't we just remove this?

evantk91: Yes I removed all the commented out lines.

PR #700, opened 2024-01-29
Full qualified name for style resource

Ayyanchira:
## 🔹 Jira Ticket(s) if any

* [MOB-?](https://iterable.atlassian.net/browse/MOB-?)

## ✏️ Description

> Fiterable app was throwing error as resource not found. Making the resource with full name is making it work. Probably would be helpful in all projects having such conflicts in finding the resource. Also bumping up the minSdkversion in api-ui


evantk91 (approved):

davidtruong (approved):
LFTM

PR #709, opened 2024-03-06
[MOB-6832] - AuthDecode crash fix

Ayyanchira:
## 🔹 Jira Ticket(s) if any

* [MOB-6832](https://iterable.atlassian.net/browse/MOB-6832)

## ✏️ Description

> Prevents crash by checking if the three components of auth token are present before decoding the expiration time. If not the SDK will throw an Exception continuing with retry mechanism and invoking failure callbacks.

[MOB-6832]: https://iterable.atlassian.net/browse/MOB-6832?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


evantk91 (approved):
LGTM. Take a look at code climate fix might be worth doing.

PR #745, opened 2024-06-25
[MOB-8948] sets anonymous user

evantk91:
## 🔹 Jira Ticket(s) if any

* [MOB-8948](https://iterable.atlassian.net/browse/MOB-8948)

## ✏️ Description

> This pull request introduces a flag that indicates if a user is anonymous which determines if stored user id is cleared or not .

[MOB-8946]: https://iterable.atlassian.net/browse/MOB-8946?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[MOB-8948]: https://iterable.atlassian.net/browse/MOB-8948?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
@@ -780,22 +780,23 @@ public class IterableApi {
 
     public void setAnonUser(@Nullable String userId) {
         _userIdAnon = userId;
+        setUserId(userId, null, null, null, true);
         storeAuthData();
     }
 
     public void setUserId(@Nullable String userId) {
-        setUserId(userId, null, null, null);
+        setUserId(userId, null, null, null, false);
     }
 
     public void setUserId(@Nullable String userId, @Nullable IterableHelper.SuccessHandler successHandler, @Nullable IterableHelper.FailureHandler failureHandler) {
-        setUserId(userId, null, successHandler, failureHandler);
+        setUserId(userId, null, successHandler, failureHandler, false);
     }
 
     public void setUserId(@Nullable String userId, @Nullable String authToken) {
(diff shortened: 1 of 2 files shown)


iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -811,7 +812,11 @@ public void setUserId(@Nullable String userId, @Nullable String authToken, @Null
...
                     logoutPreviousUser();
-                    _userIdAnon = null;
+
+                    if (!isAnon) {
+                        _userIdAnon = null;
+                    }
+
                     _email = null;
                     _userId = userId;
                     anonymousUserManager.syncEvents();
```
Ayyanchira, line 822: anonymousUserManager.syncEvents() will get called everytime even if feature ius enabled or disabled or used. Will have to make sure that events are cleared in every scenario after a sync happens.

Ayyanchira (approved):
LGTM

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
evantk91: Yeah I had some thoughts on this. Seems like merge API is called everytime even when feature is disabled. We should wrap this in a check for the config flag. I will cut a ticket.

evantk91: Well syncEvents would only occur if the merge was successful. I will verify though.

PR #781, opened 2024-09-04
MOB-6575 - Android crash on press push notification

megha-iterable:
## Jira Ticket(s) if any

* [MOB-6575](https://iterable.atlassian.net/browse/MOB-6575)

## Description

> Android crash on press push notification

[MOB-6575]: https://iterable.atlassian.net/browse/MOB-6575?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


evantk91 (approved):

marcorighini:
Hi @evantk91 @megha-iterable , 
we discovered that changing this has several implication on the already existing activities that are in the same affinity (the default affinity). These activities are destroyed in order to have the iterable trampoline activity as the root of the task. 

Shouldn't we set a different affinity for the iterable task, instead of having the default affinity? In this way it will have no consequences on the existing activities in the current default task

PR #822, opened 2024-10-23
[MOB-10031] user update array fix

evantk91:
## 🔹 Jira Ticket(s) if any

* [MOB-10031](https://iterable.atlassian.net/browse/MOB-10031)

## ✏️ Description

> This pull request updates the user update aggregation to replace nested arrays instead of appending to them.

[MOB-10031]: https://iterable.atlassian.net/browse/MOB-10031?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Ayyanchira (approved):
Good catch!

Ayyanchira (commented):
What happens if the object is array type?

evantk91:
> What happens if the object is array type?

It is treated like any other value and is replaced rather than appended to.

PR #841, opened 2024-11-19
[MOB-9211] - InApp resizing fix

Ayyanchira:
## 🔹 Jira Ticket(s) if any

* [MOB-9211](https://iterable.atlassian.net/browse/MOB-9211)

## ✏️ Description

> Removing treeobserver as the html would keep calling this callback even after the page was finished loading. This method was used as an additional measure to make sure inapps are resized in situation when inapps were still not displayed after finished loading. This fix adds delay of 100ms which seems to resolve inapps on Pixel 6pro and still making it all work on Pixel 8 device, needing the SDK to not rely on tree observers.

[MOB-9211]: https://iterable.atlassian.net/browse/MOB-9211?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFragmentHTMLNotification.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFragmentHTMLNotification.java
@@ -22,7 +22,6 @@ import android.view.LayoutInflater;
 import android.view.OrientationEventListener;
 import android.view.View;
 import android.view.ViewGroup;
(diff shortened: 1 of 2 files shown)


evantk91 (approved):
Looks good. Tested with all positions with both pixel 6 pro and pixel 8 emulator on API 34.

PR #853, opened 2025-01-02
[MOB-9235] Fix variable IV length issue

sumeruchat:
Fix Variable IV Length Issue

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableDataEncryptor.kt
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableDataEncryptor.kt
@@ -23,7 +23,8 @@ class IterableDataEncryptor {
         private const val TRANSFORMATION_LEGACY = "AES/CBC/PKCS5Padding"
         private const val ITERABLE_KEY_ALIAS = "iterable_encryption_key"
         private const val GCM_TAG_LENGTH = 128
-        private const val IV_LENGTH = 16
+        private const val GCM_IV_LENGTH = 12
+        private const val CBC_IV_LENGTH = 16
         private val TEST_KEYSTORE_PASSWORD = "test_password".toCharArray()
 
         private val keyStore: KeyStore by lazy {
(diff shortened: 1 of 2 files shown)


iterableapi/src/main/java/com/iterable/iterableapi/IterableDataEncryptor.kt
```diff
@@ -132,11 +133,12 @@ class IterableDataEncryptor {
...
             }
 
-            // Combine isModern flag, IV, and encrypted data
-            val combined = ByteArray(1 + encryptedData.iv.size + encryptedData.data.size)
+            // Combine isModern flag, IV length, IV, and encrypted data
+            val combined = ByteArray(1 + 1 + encryptedData.iv.size + encryptedData.data.size)
             combined[0] = if (encryptedData.isModernEncryption) 1 else 0
-            System.arraycopy(encryptedData.iv, 0, combined, 1, encryptedData.iv.size)
-            System.arraycopy(encryptedData.data, 0, combined, 1 + encryptedData.iv.size, encryptedData.data.size)
+            combined[1] = encryptedData.iv.size.toByte()  // Store IV length
```
evantk91, line 139: I see IV length is pulled from data instead of being hardcoded.

iterableapi/src/main/java/com/iterable/iterableapi/IterableDataEncryptor.kt
```diff
@@ -187,16 +190,15 @@ class IterableDataEncryptor {
         }
         
         val cipher = Cipher.getInstance(TRANSFORMATION_MODERN)
-        val iv = generateIV()
-        val spec = GCMParameterSpec(GCM_TAG_LENGTH, iv)
-        cipher.init(Cipher.ENCRYPT_MODE, getKey(), spec)
```
evantk91, line 192: Why is generateIV removed here? Seems like it should follow encrypt legacy but with modern set to true

evantk91 (approved):
Added some clarifications

iterableapi/src/main/java/com/iterable/iterableapi/IterableDataEncryptor.kt
sumeruchat: in modern we dont generate instead we let android generate it

iterableapi/src/main/java/com/iterable/iterableapi/IterableDataEncryptor.kt
sumeruchat: right so we have variable iv lenghts and we encode it

sumeruchat: and then when we decode it we know what the iv length was

PR #864, opened 2025-01-16
[MOB-10627] removes disable device call

evantk91:
## 🔹 Jira Ticket(s) if any

* [MOB-10627](https://iterable.atlassian.net/browse/MOB-10627)

## ✏️ Description

> This pull request removes the disable device call from the enhanced notification tracking logic.

[MOB-10627]: https://iterable.atlassian.net/browse/MOB-10627?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
@@ -411,11 +411,7 @@ public class IterableApi {
 
         if (sharedInstance.isInitialized()) {
             if (sharedInstance.config.autoPushRegistration && hasStoredPermission && (isNotificationEnabled != systemNotificationEnabled)) {
-                if (!systemNotificationEnabled) {
-                    sharedInstance.disablePush();
-                } else {
-                    sharedInstance.registerForPush();
-                }
+                sharedInstance.registerForPush();
             }
 
             SharedPreferences.Editor editor = sharedPref.edit();


sumeruchat (approved):

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -411,11 +411,7 @@ private void onForeground() {
 
         if (sharedInstance.isInitialized()) {
             if (sharedInstance.config.autoPushRegistration && hasStoredPermission && (isNotificationEnabled != systemNotificationEnabled)) {
-                if (!systemNotificationEnabled) {
-                    sharedInstance.disablePush();
-                } else {
-                    sharedInstance.registerForPush();
-                }
+                sharedInstance.registerForPush();
```
sumeruchat, line 414: same as iOS when notifications are disabled will android give us a token? If not maybe we have to use the previously registered token

sumeruchat (commented):
lets test the disable and enable case together before merging

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
davidtruong: You shouldn't need to OS permissions turned on to get the push token, you just request it from the device. It is given regardless of that setting, the OS setting simply just gates whether or not the device will display the notifications.

sumeruchat: @davidtruong Okay sounds good then. lets test it thoroughly

PR #884, opened 2025-02-19
MOB-10857 fix encryption crash

sumeruchat:
## 🔹 Jira Ticket(s) if any

## ✏️ Description

Added plaintext fallback for data storage when encryption fails:
- Modified `IterableKeychain` to handle encryption failures gracefully
- Data is stored in plaintext with a flag when encryption fails
- Added unit tests to verify fallback behavior
- Maintains backward compatibility with existing encrypted data

Key changes:
1. Added plaintext flag to track unencrypted data
2. Implemented fallback mechanism in `secureSave`
3. Updated `secureGet` to handle plaintext data
4. Added comprehensive test coverage

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableKeychain.kt
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableKeychain.kt
@@ -9,6 +9,7 @@ class IterableKeychain {
         const val KEY_EMAIL = "iterable-email"
 		const val KEY_USER_ID = "iterable-user-id"
 		const val KEY_AUTH_TOKEN = "iterable-auth-token"
+        private const val PLAINTEXT_SUFFIX = "_plaintext"
     }
 
     private var sharedPrefs: SharedPreferences
@@ -74,6 +75,11 @@ class IterableKeychain {
     }
 
     private fun secureGet(key: String): String? {
+        // First check if it's stored in plaintext
+        if (sharedPrefs.getBoolean(key + PLAINTEXT_SUFFIX, false)) {
+            return sharedPrefs.getString(key, null)
+        }
+        
         return try {
             sharedPrefs.getString(key, null)?.let { encryptor.decrypt(it) }
         } catch (e: Exception) {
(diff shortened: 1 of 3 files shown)


iterableapi/src/main/java/com/iterable/iterableapi/IterableKeychain.kt
```diff
@@ -83,9 +89,22 @@ class IterableKeychain {
...
+        }
+
+        try {
+            editor.putString(key, encryptor.encrypt(value))
+                .remove(key + PLAINTEXT_SUFFIX)
+                .apply()
+        } catch (e: Exception) {
+            handleDecryptionError(e)
+            editor.putString(key, value)
+                .putBoolean(key + PLAINTEXT_SUFFIX, true)
```
evantk91, line 105: So my understanding of this is we only add the suffix if there is a problem encrypting the key

evantk91 (approved):
LGTM. Just need some clarification.

iterableapi/src/main/java/com/iterable/iterableapi/IterableKeychain.kt
sumeruchat: yes if there is a crash we save the data in plaintext

evantk91: Is there no removal for key without suffix?

sumeruchat: well with the key with a suffix we are just saving boolean. The original key still gets used for saving the actual value

iterableapi/src/main/java/com/iterable/iterableapi/IterableKeychain.kt
```diff
@@ -83,9 +89,22 @@ class IterableKeychain {
     }
 
     private fun secureSave(key: String, value: String?) {
-        sharedPrefs.edit()
-            .putString(key, value?.let { encryptor.encrypt(it) })
-            .apply()
+        val editor = sharedPrefs.edit()
+        if (value == null) {
+            editor.remove(key).remove(key + PLAINTEXT_SUFFIX).apply()
```
evantk91, line 94: I see the removal is here.

Ayyanchira (approved):
LGTM

PR #893, opened 2025-03-27
[MOB-9340]  Fix for removing recalled campaigns from device memory

sumeruchat:
Ticket: https://iterable.atlassian.net/browse/MOB-9340

ISSUE: In-App Messages Remain Visible After Server Recall

PROBLEM DESCRIPTION:
In-app messages continue to display on Android devices even after they've been recalled server-side. The SDK recognizes the recall event but fails to mark these messages as consumed locally, causing them to remain visible to users.

ROOT CAUSE:
During synchronization, the SDK correctly identifies messages that have been removed from the server (missing from server response but present in local storage)

FIX IMPLEMENTED:
Modified the syncWithRemoteQueue() method in IterableInAppManager.java to mark messages as consumed before removing them from storage when they're no longer present in the server response.

The changes are minimal and targeted, similar to the iOS fix. Added unit test to verify the fix works correctly.

### DIFF
--- a/iterableapi/src/test/java/com/iterable/iterableapi/IterableInAppManagerSyncTest.java
+++ b/iterableapi/src/test/java/com/iterable/iterableapi/IterableInAppManagerSyncTest.java
@@ -90,10 +90,10 @@ public class IterableInAppManagerSyncTest extends BaseTest {
         // Create a test message in local storage
         IterableInAppMessage testMessage = InAppTestUtils.getTestInboxInAppWithId("test-message-1");
         doReturn(testMessage).when(storageMock).getMessage("test-message-1");
-        
+
         // Create a storage with only this message
         doReturn(Arrays.asList(testMessage)).when(storageMock).getMessages();
-        
+
         // Setup the API to return empty message list (simulating recall)
         doAnswer(new Answer() {
             @Override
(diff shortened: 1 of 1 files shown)


app/build.gradle
```diff
@@ -21,6 +21,10 @@ android {
         multiDexEnabled true
     }
 
+    buildFeatures {
+        buildConfig true
+    }
+
```
Ayyanchira, line 27: Lets remove all the build and embedded related changes from this PR

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppManager.java
```diff
@@ -365,7 +367,10 @@ private void syncWithRemoteQueue(List<IterableInAppMessage> remoteQueue) {
 
         for (IterableInAppMessage localMessage : storage.getMessages()) {
             if (!remoteQueueMap.containsKey(localMessage.getMessageId())) {
+                // Mark message as consumed before removing it to prevent it from being displayed
+                localMessage.setConsumed(true);
                 storage.removeMessage(localMessage);
+                api.inAppConsume(localMessage, null, null, null, null);
```
Ayyanchira, line 373: Lets get a confirmation from product to see if recalled message should be consumed. As it could affect the analytics. 
Also referencing `localMessage` after its removal could cause problems.

Ayyanchira (commented):
Added suggestions

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppManager.java
```diff
@@ -321,6 +321,8 @@ public void handleInAppClick(@NonNull IterableInAppMessage message, @Nullable Ur
     synchronized void removeMessage(String messageId) {
         IterableInAppMessage message = storage.getMessage(messageId);
         if (message != null) {
+            // Mark message as consumed before removing it to prevent it from being displayed
+            message.setConsumed(true);
```
Ayyanchira: Not sure if we need this

PR #899, opened 2025-04-04
[MOB-11167] Add timeout for crypto operations to solve ANRs and config to disable encryption altogether

sumeruchat:
## 🔹 Jira Ticket(s) if any

* [MOB-11167](https://iterable.atlassian.net/browse//MOB-11167)

## ✏️ Description

This PR fixes an ANR (Application Not Responding) issue that occurs during app initialization when the Android KeyStore operations block the main thread.

Key changes:

Added a background thread executor for all crypto operations in IterableKeychain

Implemented a 500ms timeout mechanism for all crypto operations

Added error handling for timeout scenarios with appropriate fallbacks

Ensured resetKeys operation also respects the timeout limit

This approach minimizes code changes while ensuring that crypto operations never block the main thread during app startup, eliminating the ANR 

Added config flag to disable encryption altogether.

[MOB-11167]: https://iterable.atlassian.net/browse/MOB-11167?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableConfig.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableConfig.java
@@ -97,7 +97,7 @@ public class IterableConfig {
      * When set to true, disables encryption for keychain storage.
      * By default, encryption is enabled for storing sensitive user data.
      */
-    final boolean disableKeychainEncryption;
+    final boolean keychainEncryption;
 
     /**
      * Handler for decryption failures of PII information.
@@ -127,7 +127,7 @@ public class IterableConfig {
         dataRegion = builder.dataRegion;
         useInMemoryStorageForInApps = builder.useInMemoryStorageForInApps;
         enableEmbeddedMessaging = builder.enableEmbeddedMessaging;
-        disableKeychainEncryption = builder.disableKeychainEncryption;
+        keychainEncryption = builder.keychainEncryption;
         decryptionFailureHandler = builder.decryptionFailureHandler;
         mobileFrameworkInfo = builder.mobileFrameworkInfo;
     }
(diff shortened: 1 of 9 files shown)


iterableapi/src/main/java/com/iterable/iterableapi/IterableConfig.java
```diff
@@ -93,6 +93,12 @@ public class IterableConfig {
      */
     final boolean enableEmbeddedMessaging;
 
+    /**
+     * When set to true, disables encryption for keychain storage.
+     * By default, encryption is enabled for storing sensitive user data.
+     */
+    final boolean disableKeychainEncryption;
```
Ayyanchira, line 100: Lets reword this to just `encryption`. `Disable` prefix also adds complexity when one wants to enable encryption and wants to use this config. They will have to disable the disableKeychain Encryption.

Ayyanchira (commented):
Added a nit pick

iterableapi/src/main/java/com/iterable/iterableapi/IterableConfig.java
evantk91: Well when set to true it disables encryption. So the default value is false. Unless we want to rename to enableKeychainEncryption and default to true.

evantk91 (approved):

iterableapi/src/main/java/com/iterable/iterableapi/IterableConfig.java
Ayyanchira: Went with just `encryption` to keep it clean and easy to understand. Adding prefixes made it more confusing than helpful.

PR #929, opened 2025-07-24
[MOB-11764] Fix JWT refresh timer race condition on app background 

sumeruchat:
## Problem

Customer reported JWT refresh timer not being cleared when app goes to background after null JWT token is returned, potentially causing hanging timers and unnecessary retry attempts.

## Root Cause  

Race condition between:
- `scheduleAuthTokenRefresh()` (executor thread) - schedules retry timer when JWT returns null
- `clearRefreshTimer()` (activity lifecycle thread) - clears timer on app background

If app backgrounds immediately after null JWT response, timer creation and clearing could race, leaving uncanceled timers.

## Solution

- **Synchronized** `scheduleAuthTokenRefresh()` and `clearRefreshTimer()` methods
- Ensures timer operations are **atomic** and **thread-safe**
- Made `isTimerScheduled` visible for testing with `@VisibleForTesting`

## Testing

✅ **New test**: `testNullJWTRefreshTimerRaceConditionOnBackground`  
✅ **All auth tests pass** (24/24)  
✅ **Full test suite passes** (200+ tests)  
✅ **No regressions**

## Manual Test Case

1. **Setup**: Configure auth handler to return null JWT
2. **Trigger**: Set user email to start auth flow  
3. **Background**: Immediately background app during auth retry
4. **Verify**: No hanging timers continue firing after app backgrounds
5. **Foreground**: Bring app to foreground - should cleanly restart auth

## Impact

- **Fixes**: Hanging JWT refresh timers on app background
- **Improves**: App lifecycle management during auth failures  
- **Minimal**: Thread-safe synchronization, no functional changes

Fixes [MOB-11764](https://iterable.atlassian.net/browse/MOB-11764)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -59,6 +59,9 @@ freeline_project_description.json
 # MacOS generated files
 .DS_Store
 
+# AI assistant configuration
+.claude/
+
 # [Maven] ========================
 target/
 pom.xml.tag
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,7 @@
+# CLAUDE.md
+
+🤖 **AI Agent Instructions**
+
+Please read the `agent/AGENT_README.md` file for comprehensive project information, development workflow, and testing procedures.
+
+All the information you need to work on this Iterable Android SDK project is documented there. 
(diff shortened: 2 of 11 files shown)


Ayyanchira:
@sumeruchat , is this branch and PR also for solving the same JWT in background issue? Should we close this one?

Ayyanchira:
Should we close this as this is duplicate @sumeruchat ?

PR #952, opened 2025-10-14
SDK-113 Fix WebView CORS issue with self-hosted custom fonts

sumeruchat:
## 🐛 Problem

Wolt reported that their self-hosted custom fonts aren't rendering in Iterable in-app messages on Android. The same fonts work perfectly on iOS and Web, but Android users see fallback fonts instead of Wolt's branded typography.

## 🔍 Root Cause

The issue was in `IterableWebView.loadDataWithBaseURL()` - it was passing an empty string as the baseURL parameter. This caused the WebView to use `about:blank` as the origin instead of a proper web origin.

When the HTML tried to load Wolt's custom fonts from `https://webfonts.wolt.com/`, the browser's CORS policy blocked the requests because:
- Wolt's server CORS is configured to allow `https://app.eu.iterable.com`
- But the WebView was requesting from `about:blank` origin
- CORS mismatch = blocked fonts = sad customers 😢

## ✅ Solution

**Added a configurable WebView base URL parameter** that customers can set to match their CORS configuration:

- **New config option**: `setWebViewBaseUrl()` in `IterableConfig.Builder`
- **Backward compatible**: Defaults to empty string (original behavior) if not set
- **Flexible**: Works for any customer domain, not just Iterable's
- **Internal implementation**: Added `IterableUtil.getWebViewBaseUrl()` helper method

## 🚀 Usage

**For Wolt (fixes the CORS issue):**
```java
IterableConfig config = new IterableConfig.Builder()
    .setWebViewBaseUrl("https://app.eu.iterable.com")  // Matches their CORS config
    .build();
IterableApi.initialize(context, apiKey, config);
```

**For existing customers (no change needed):**
```java
IterableApi.initialize(context, apiKey);  // Still works exactly the same
```

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableUtil.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableUtil.java
@@ -89,7 +89,7 @@ public class IterableUtil {
         return instance.isUrlOpenAllowed(url);
     }
 
-    static String getWebViewBaseUrl() {
+    public static String getWebViewBaseUrl() {
         return instance.getWebViewBaseUrl();
     }
 }


Ayyanchira (approved):
Looking great Sumeru!

PR #960, opened 2025-11-03
BCIT - EM Local

Ayyanchira:
## 🔹 Jira Ticket(s) if any

* [SDK-16](https://iterable.atlassian.net/browse/SDK-16)

### Introduced new Test like for inapp.
1. Instead of campaign trigger, this test updates the user with `isPremium:true` value. Based on this, a live campaign (assumed to be running endlessly on server) will have this user eligible.
2. Hardcoded PlacementID value is used.
3. EmbeddedTestActivity and its corresponding xml is now updated to be able to manually test the functionality as well.

[SDK-16]: https://iterable.atlassian.net/browse/SDK-16?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/integration-tests/src/androidTest/java/com/iterable/integration/tests/EmbeddedMessageIntegrationTest.kt
+++ b/integration-tests/src/androidTest/java/com/iterable/integration/tests/EmbeddedMessageIntegrationTest.kt
@@ -35,7 +35,6 @@ class EmbeddedMessageIntegrationTest : BaseIntegrationTest() {
     
     private lateinit var uiDevice: UiDevice
     private lateinit var mainActivityScenario: ActivityScenario<MainActivity>
-    private lateinit var embeddedActivityScenario: ActivityScenario<EmbeddedMessageTestActivity>
     
     @Before
     override fun setUp() {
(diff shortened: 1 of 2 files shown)


integration-tests/src/androidTest/java/com/iterable/integration/tests/EmbeddedMessageIntegrationTest.kt
```diff
@@ -0,0 +1,268 @@
...
+                val state = mainActivityScenario.state
+                Log.d(TAG, "🔧 MainActivity state: $state")
+                state == Lifecycle.State.RESUMED
+            }
+        
+        Log.d(TAG, "🔧 MainActivity is ready!")
+        
+        // Step 2: Click the "Embedded Messages" button to navigate to EmbeddedMessageTestActivity
+        Log.d(TAG, "🔧 Step 2: Clicking 'Embedded Messages' button...")
+        val embeddedButton = uiDevice.findObject(UiSelector().resourceId("com.iterable.integration.tests:id/btnEmbeddedMessages"))
```
joaodordio: Not sure if that's possible here but is there a way to add this to a constants files or something similar?

PR #978, opened 2026-01-21
Double Callback fixes proposal

franco-zalamena-iterable:
## 🔹 Jira Ticket(s) if any

* [SDK-289](https://iterable.atlassian.net/browse/SDK-289)

## ✏️ Description

After discussion, the "problems" with multiple callback calls was discovered to be business logic, there are improvements to be made, but two successes was expected. The problem is that one of the successes is a local success and the other is a remote success.
The proposal for the fix is to have it differentiated so when implementing the callback it is possible to choose which call is relevant

[SDK-289]: https://iterable.atlassian.net/browse/SDK-289?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
@@ -12,13 +12,11 @@ import androidx.annotation.RestrictTo;
 import androidx.annotation.VisibleForTesting;
 import androidx.core.app.NotificationManagerCompat;
 
-import com.iterable.iterableapi.response.IterableAuthResponseObject;
 import com.iterable.iterableapi.response.IterableResponseObject;
 import com.iterable.iterableapi.response.handlers.IterableCallbackHandlers;
 import com.iterable.iterableapi.response.handlers.auth.IterableAuthCallbackHandlers;
 import com.iterable.iterableapi.util.DeviceInfoUtils;
 
-import org.jetbrains.annotations.NotNull;
 import org.json.JSONException;
 import org.json.JSONObject;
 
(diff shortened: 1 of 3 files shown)


iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -449,8 +449,9 @@ private void completeUserLogin(@Nullable String email, @Nullable String userId,
         }
 
         if (config.autoPushRegistration) {
-            registerForPush();
-        } else if (_setUserSuccessCallbackHandler != null) {
+            registerForPush(); //TODO: FIX, THE LOGIN NEVER CALLS THE CALLBACK IF THE AUTOPUSH REGISTRATION IS TRUE
+        }
+        if (_setUserSuccessCallbackHandler != null) { // todo: why is there an else if it can be both true
             _setUserSuccessCallbackHandler.onSuccess(new JSONObject()); // passing blank json object here as onSuccess is @Nonnull
```
Ayyanchira: Problem here is we need the registerPush() to be passing a completion handler so that we can pass the correct onSuccess Object

Ayyanchira: Thanks for asking that question. I didnt do a good job up there. 
What I meant was, because the registerForPush() is itself a void method, it is a fire and forget method. The next line of code will not wait for success of failure from that. Therefore, calling a successCallbackHandler right after registerForPush() may give incorrect result for the parent method `setEmail(email, successahandler, failureHandler)`
To mitigate that, we might have to make registerDevice async as well where we get result back and decide to call success or failure callback. 
Perhaps in setEmail case, we directly call the callback from api level and registerForPush() doesnt need callback. Free to hear more

PR #985, opened 2026-02-03
[SDK-289] Fixed double callback problem for setEmail with auto push registration

franco-zalamena-iterable:
## 🔹 Jira Ticket(s) if any

* [SDK-289](https://iterable.atlassian.net/browse/SDK-289)

## ✏️ Description

This fix is meant to be a more straight forward fix for the double callback issue some clients have encountered when autoPushRegistration is set to false.

Root cause:
When autoPushRegistration is false and we call setEmail, no remote call is made to actually update the current user in the iterable app, so we wait for the push registration to be made to consider it completely done.
Because some users were using the callbackHandler within a couroutine, when the registerDeviceToken was called we would wrap the setEmail callback into the registerDeviceToken call, so we could notify when the setEmail flow was actually completed, but that caused the callback to fire twice, and if the coroutine was dependant of a success to continue that would cause the app to crash.

The Solution:
To clear the success handler when it is called on autoPushRegistration = false, so when the user calls for registering the device we won't cause the same success handler to fire twice.
Also some better responses are provided for better understanding of the process and why things are happening this way.

[SDK-289]: https://iterable.atlassian.net/browse/SDK-289?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableHelper.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableHelper.java
@@ -38,7 +38,7 @@ public class IterableHelper {
 }
 
 class IterableResponse {
-    static JSONObject setEmailLocalSuccessResponse = new JSONObject(Map.of("message", "setEmail was completed locally, but still requires the register device call. If possible, use autoPushRegistration."));
+    static final JSONObject setEmailLocalSuccessResponse = new JSONObject(Map.of("message", "setEmail was completed locally."));
 
-    static JSONObject setReadLocalSuccessResponse = new JSONObject(Map.of("message", "setRead was completed locally, no remote call was done yet."));
+    static final JSONObject setReadLocalSuccessResponse = new JSONObject(Map.of("message", "setRead was completed locally."));
 }


iterableapi/src/main/java/com/iterable/iterableapi/IterableHelper.java
```diff
@@ -33,4 +35,10 @@ public interface FailureHandler {
     public interface SuccessAuthHandler {
         void onSuccess(@NonNull String authToken);
     }
+}
+
+class IterableResponse {
+    static JSONObject setEmailLocalSuccessResponse = new JSONObject(Map.of("message", "setEmail was completed locally, but still requires the register device call. If possible, use autoPushRegistration."));
+
+    static JSONObject setReadLocalSuccessResponse = new JSONObject(Map.of("message", "setRead was completed locally, no remote call was done yet."));
```
Ayyanchira: Any reason why these are static?

Ayyanchira: Hmm I think for it to be retain its value throughout the flow ?

PR #987, opened 2026-02-08
[SDK-342] - AuthRetry logic

Ayyanchira:
## 🔹 Jira Ticket(s) if any

* [SDK-342](https://iterable.atlassian.net/browse/SDK-342)

## ✏️ Description

Adds automatic JWT retry for offline-queued requests. When a stored offline task receives a 401 JWT error, instead of retrying inline, the task is retained in the DB and task processing pauses until a fresh token is obtained.

### Changes by file

**IterableApi.java**
- Reads `autoRetry` flag from remote configuration and persists it in SharedPreferences.
- Exposes package-private `isAutoRetryOnJwtFailure()` for internal use only (not customer-facing).

**IterableAuthManager.java**
- Introduces `AuthState` enum (`VALID`, `INVALID`, `UNKNOWN`) to represent JWT token health.
- Adds `AuthTokenReadyListener` pub/sub so components (e.g., `IterableTaskRunner`) can react when a new token is available.
- `handleAuthTokenSuccess` sets state to `UNKNOWN` (new token, not yet verified); `setIsLastAuthTokenValid(true)` transitions to `VALID`.
- State transitions from `INVALID` → non-`INVALID` notify listeners to resume processing.

**IterableTaskRunner.java**
- Implements `AuthTokenReadyListener`. On JWT 401, marks auth `INVALID`, pauses processing, and returns `RETRY` without deleting the task.
- On `onAuthTokenReady()` callback, resumes task processing with the fresh token.
- Always uses the current live auth token (via `authTokenOverride` in `fromJSON`) instead of the stale one stored in the DB at queue time.

**IterableRequestTask.java**
- When `autoRetry` is enabled and the request is offline, skips inline retry and schedules a delayed auth token refresh instead — `IterableTaskRunner` handles the retry.
- Defensive null-check on `getErrorStream()` and `responseCode >= 0` guard against network inspector interference returning -1.
- Heuristic: if `responseCode == -1` but response body contains JWT error codes, infers 401.
- `fromJSON` overload accepts optional `authTokenOverride` for injecting fresh tokens into deserialized offline tasks.

**OfflineRequestProcessor.java**
- Registers `taskRunner` as `AuthTokenReadyListener` on construction.
- `dispose()` method unregisters the listener to prevent stale accumulation when offline mode is toggled.

**IterableApiClient.java**
- Calls `dispose()` on the old `OfflineRequestProcessor` before replacing it in `setOfflineProcessingEnabled()`.

### Resolved concerns from earlier draft
- ~~`IterableConfig.autoRetryOnJwtFailure` should not be customer-facing~~ → Removed from `IterableConfig` (commit be05ccb0). Flag is now remote-config only.
- ~~`fromJSON` authTokenOverride purpose unclear~~ → Used by `TaskRunner.processTask()` to inject fresh token into stale DB tasks. Correct architectural choice (task-management concern belongs in TaskRunner, not RequestTask).
- ~~`isJwtFailure` method redundant?~~ → Kept as a small, well-named abstraction. In offline context, any 401 is treated as JWT failure since the API key was valid at queue time.
- ~~responseCode -1 handling~~ → Defensive heuristic kept. Null errorStream check is valid per Android HttpURLConnection docs.

### ⚠️ Known issue discovered during testing (pre-existing, not introduced by this PR)

**`registerDeviceToken` is called repeatedly on every JWT rotation.**

`setAuthToken()` calls `completeUserLogin()` whenever the token string changes. `completeUserLogin()` unconditionally calls `registerForPush()`, `syncInApp()`, and `syncMessages()`. This means every JWT refresh (e.g. on expiration) triggers a full re-registration cycle even though the user identity (email/userId) and push token (FCM) haven't changed.

`completeUserLogin` was designed for user identity changes (`setEmail`/`setUserId`), where re-registering makes sense. But `setAuthToken` also triggers it on routine JWT rotation, causing unnecessary API calls. `setAuthToken` should distinguish between "new user login" and "credential refresh" to avoid redundant work.

**Needs a separate ticket to address.**

[SDK-342]: https://iterable.atlassian.net/browse/SDK-342?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/OfflineRequestProcessor.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/OfflineRequestProcessor.java
@@ -30,8 +30,11 @@ class OfflineRequestProcessor implements RequestProcessor {
             IterableConstants.ENDPOINT_TRACK_INAPP_CLOSE,
             IterableConstants.ENDPOINT_TRACK_INBOX_SESSION,
             IterableConstants.ENDPOINT_TRACK_INAPP_DELIVERY,
-            IterableConstants.ENDPOINT_GET_EMBEDDED_MESSAGES,
             IterableConstants.ENDPOINT_INAPP_CONSUME,
+            IterableConstants.ENDPOINT_UPDATE_CART,
+            IterableConstants.ENDPOINT_TRACK_EMBEDDED_RECEIVED,
+            IterableConstants.ENDPOINT_TRACK_EMBEDDED_CLICK,
+            IterableConstants.ENDPOINT_TRACK_EMBEDDED_SESSION,
             IterableConstants.ENDPOINT_DISABLE_DEVICE));
 
     OfflineRequestProcessor(Context context) {


iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -194,6 +203,22 @@ static void loadLastSavedConfiguration(Context context) {
...
+            sharedInstance._autoRetryOnJwtFailure = sharedInstance.config.autoRetryOnJwtFailure;
+        }
+    }
+
+    /**
+     * Returns whether auto-retry on JWT failure is enabled.
+     * The remote configuration flag takes precedence when present;
+     * otherwise the local {@link IterableConfig#autoRetryOnJwtFailure} value is used.
+     */
+    boolean isAutoRetryOnJwtFailure() {
```
franco-zalamena-iterable: is there a reason for it not to be private or public?

iterableapi/src/main/java/com/iterable/iterableapi/IterableRequestTask.java
```diff
@@ -186,13 +194,36 @@ static IterableApiResponse executeApiRequest(IterableApiRequest iterableApiReque
                     jsonError = e.getMessage();
                 }
 
+                // If getResponseCode() returned -1 (e.g. due to network inspector
+                // interference) but the response body contains JWT error codes,
+                // we can infer the actual response was a 401.
+                if (responseCode == -1 && matchesJWTErrorCodes(jsonResponse)) {
+                    responseCode = 401;
+                }
```
Ayyanchira: -1's can be discarded. Dont have to categorize into 401 flow.

iterableapi/src/main/java/com/iterable/iterableapi/IterableRequestTask.java
```diff
@@ -186,13 +194,36 @@ static IterableApiResponse executeApiRequest(IterableApiRequest iterableApiReque
...
                         apiResponse = IterableApiResponse.failure(responseCode, requestResult, jsonResponse, "JWT Authorization header error");
                         IterableApi.getInstance().getAuthManager().handleAuthFailure(iterableApiRequest.authToken, getMappedErrorCodeForMessage(jsonResponse));
-                        // We handle the JWT Retry for both online and offline here rather than handling online request in onPostExecute
-                        requestNewAuthTokenAndRetry(iterableApiRequest);
+
+                        // [F] When autoRetry is enabled and this is an offline task, skip the inline
+                        // retry. The task stays in the DB and IterableTaskRunner will retry it once
+                        // a valid JWT is obtained via the AuthTokenReadyListener callback.
+                        // For online requests or when autoRetry is disabled, use the existing inline retry.
+                        boolean autoRetry = IterableApi.getInstance().isAutoRetryOnJwtFailure();
```
Ayyanchira: Also thinking, it should not be a public variable that should be accessible for developers like this. Can cause confusion.

iterableapi/src/main/java/com/iterable/iterableapi/IterableRequestTask.java
```diff
@@ -186,13 +194,36 @@ static IterableApiResponse executeApiRequest(IterableApiRequest iterableApiReque
...
+                        if (autoRetry && iterableApiRequest.getProcessorType() == IterableApiRequest.ProcessorType.OFFLINE) {
+                            // Schedule a delayed token refresh (respects retry policy).
+                            // Do NOT retry the request inline -- IterableTaskRunner will handle
+                            // the retry after the AuthTokenReadyListener callback fires.
+                            IterableAuthManager authManager = IterableApi.getInstance().getAuthManager();
+                            authManager.setIsLastAuthTokenValid(false);
+                            long retryInterval = authManager.getNextRetryInterval();
+                            authManager.scheduleAuthTokenRefresh(retryInterval, false, null);
```
Ayyanchira: This 4 line can be combined into one if authmanager has `scheduleAuthTOkenRefresh` which figures out internally whats the nextRetryInterval. Too much back and forth.

iterableapi/src/main/java/com/iterable/iterableapi/IterableTaskRunner.java
```diff
@@ -127,30 +144,51 @@ private void processTasks() {
...
             IterableApiResponse response = null;
             TaskResult result = TaskResult.FAILURE;
             try {
-                IterableApiRequest request = IterableApiRequest.fromJSON(getTaskDataWithDate(task), null, null);
+                // Use the current live auth token instead of the stale one stored in the DB.
+                // The token in the DB was captured at queue time and may have since expired.
+                String currentAuthToken = IterableApi.getInstance().getAuthToken();
+                IterableApiRequest request = IterableApiRequest.fromJSON(getTaskDataWithDate(task), currentAuthToken, null, null);
                 request.setProcessorType(IterableApiRequest.ProcessorType.OFFLINE);
                 response = IterableRequestTask.executeApiRequest(request);
```
Ayyanchira: authToken should be applied during from within IterableRequestTask.execute. I dont think we should handle and add auth token here. But can be acceptable.

iterableapi/src/main/java/com/iterable/iterableapi/IterableTaskRunner.java
```diff
@@ -127,30 +144,51 @@ private void processTasks() {
...
             } catch (Exception e) {
                 IterableLogger.e(TAG, "Error while processing request task", e);
                 healthMonitor.onDBError();
             }
 
+            boolean autoRetry = IterableApi.getInstance().isAutoRetryOnJwtFailure();
```
Ayyanchira: This is repeated. Probably just one time class level variable that gets value from AuthManager should be good.

PR #988, opened 2026-02-10
[SDK-337] Fix for status bar overlap on in-app messages

franco-zalamena-iterable:
## 🔹 Jira Ticket(s) if any

* [SDK-337](https://iterable.atlassian.net/browse/SDK-337)

## ✏️ Description

We had a problem with fullscreen in-app messages, some clients were getting the close button behind the status bar and losing the ability to close the message.
The flag `FLAG_FULLSCREEN` was just making the content appear under the status bar but blocking clicks and obstructing the view. 

The fix is to have new flags:
- FORCE_RESPECT_BOUNDS;
- FORCE_EDGE_TO_EDGE;
- FOLLOW_APP_LAYOUT;
- FORCE_FULLSCREEN.

These are meant to control the way the in-app will be drawn in relation to the status bar of the app. The default behavior will be FORCE_EDGE_TO_EDGE so clients implementing it will not have to migrate or opt out

[SDK-337]: https://iterable.atlassian.net/browse/SDK-337?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFragmentHTMLNotification.java
```diff
@@ -507,11 +497,7 @@ public void run() {
             }
         };
 
-        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN) {
```
franco-zalamena-iterable: We do not support JELLY_BEAN anymore, so this is not doing anything

sumeruchat (changes requested):
Great work! Overall looks good to me but there is a docs issue we should resolve before merging :) 

The enum's javadoc on `FOLLOW_APP_LAYOUT` says *"Default."*, but the actual Builder default and `resolveDisplayMode()` fallback are both `FORCE_EDGE_TO_EDGE`. The CHANGELOG also documents `FORCE_EDGE_TO_EDGE` as the default.

This is also inconsistent with the original PR description, which said the default would be `FOLLOW_APP_LAYOUT` so "clients implementing it will not have to think about it."

Can you clarify which was intended?

1. **Opt-in (current behavior)** — existing clients keep hitting the obscured close-button bug until they explicitly set a different mode. If this is right, just move the "Default." note from `FOLLOW_APP_LAYOUT` to `FORCE_EDGE_TO_EDGE` in `IterableInAppDisplayMode.java`.
2. **Auto-fix (per original description)** — change the Builder default and `resolveDisplayMode()` fallback to `FOLLOW_APP_LAYOUT`, and update the CHANGELOG.

This is a product call but it needs to be consistent across javadoc, Builder default, fallback, and CHANGELOG before merge.

sumeruchat:
One more thing — given this touches system-bar handling across a range of API levels (pre-R legacy flags, R+ WindowInsetsController, API 35+ default-transparent bars) and four display modes, please do thorough manual testing before merge. Specifically:

- All four `IterableInAppDisplayMode` values on an edge-to-edge host app and a non-edge-to-edge host app.
- API levels: pre-R (legacy flag path), R–34, and 35+ (default-transparent bars path).
- Fullscreen in-apps, top/bottom/center layouts — verify the close button is reachable and not behind the status bar in every combination.
- Orientation changes while an in-app is displayed.

Worth looping QA in for a pass on this one. cc @jena-chakour

franco-zalamena-iterable:
The default should remain what we had before, force edge to edge, i am fixing the javadoc to reflect that

iterableapi/src/main/java/com/iterable/iterableapi/IterableConfig.java
```diff
@@ -140,6 +140,12 @@ public class IterableConfig {
     @Nullable
     final IterableAPIMobileFrameworkInfo mobileFrameworkInfo;
 
+    /**
+     * Controls how in-app messages interact with the system bars (status bar, navigation bar).
+     * Defaults to {@link IterableInAppDisplayMode#FORCE_EDGE_TO_EDGE}.
+     */
+    final IterableInAppDisplayMode inAppDisplayMode;
```
joaodordio, line 147: Should we add the `@NonNull` annotation here? 

```suggestion
     */
    @NonNull 
    final IterableInAppDisplayMode inAppDisplayMode;
```

joaodordio (approved):

iterableapi/src/main/java/com/iterable/iterableapi/IterableConfig.java
franco-zalamena-iterable: We are not using this annotation on other fields, i think this is not really necessary since we are only configuring it from the builder method

sumeruchat (approved):
Approved but lets test on different screen types before merging :) Lmk if i can help with that

PR #1003, opened 2026-03-13
[SDK-369] no push reregister on token refresh

franco-zalamena-iterable:
## 🔹 Jira Ticket(s) if any

* [SDK-369](https://iterable.atlassian.net/browse/SDK-369)

## ✏️ Description

Fixed a problem with completeuserlogin retrigger

Decoupling the completeUserLogin from the SetAuthToken flow. Those live independently now.

### Summary
setAuthToken replaced with updateAuthToken. This only handles the update of the auth token without the side effects of login. If the user expects a login flow, the setEmail should be called. This prevents unnecessary calls to the backend on token refreshes

[SDK-369]: https://iterable.atlassian.net/browse/SDK-369?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -1076,6 +1077,9 @@ public void setEmail(@Nullable String email, @Nullable String authToken, @Nullab
 
         if (_email != null && _email.equals(email)) {
             checkAndUpdateAuthToken(authToken);
+            _setUserSuccessCallbackHandler = successHandler;
+            _setUserFailureCallbackHandler = failureHandler;
+            onLogin(authToken, email, true, merge, replay, false, failureHandler);
```
joaodordio: The `checkAndUpdateAuthToken(authToken)` call is now redundant on the `setEmail` and `setUserId` methods because the `onLogin()` will always do `updateAuthToken(authToken)`

iterableapi/src/main/java/com/iterable/iterableapi/IterableAuthManager.java
```diff
@@ -95,6 +96,21 @@ void setAuthTokenInvalid() {
         setAuthState(AuthState.INVALID);
     }
 
+    /**
+     * Handles a server-side JWT rejection (401). Invalidates the current token,
+     * clears any pending refresh, and schedules a new token request using the retry policy.
+     * When the new token arrives, AuthTokenReadyListeners are notified via the
+     * INVALID → UNKNOWN state transition.
+     */
+    void handleAuthTokenRejection() {
```
joaodordio: This called from `IterableTaskRunner.processTask()` on a worker thread. 
Multiple concurrent 401s could race through the method. 
The `scheduleAuthTokenRefresh` is internally sync'd on `timerLock`, so it won't double-schedule, but the `resetFailedAuth()` + `getNextRetryInterval()` sequence could yield different retry intervals under contention

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
franco-zalamena-iterable: good catch, latest commit addresses this

iterableapi/src/main/java/com/iterable/iterableapi/IterableAuthManager.java
franco-zalamena-iterable: i reverted this to what we had on master, will keep this refactor for a different ticket to reduce complexity and possible errors

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -679,19 +683,16 @@
...
-                // The completion handler receives exact values stored to keychain, preventing TOCTOU
-                // attacks where keychain could be modified between storage and completeUserLogin execution.
-                storeAuthData((email, userId, token) -> completeUserLogin(email, userId, token));
-            } else if (bypassAuth) {
-                // SECURITY: Pass current credentials directly to completeUserLogin.
-                // completeUserLogin will validate authToken presence when JWT auth is enabled.
-                completeUserLogin(_email, _userId, _authToken);
-            }
+            _authToken = authToken;
+            storeAuthData();
```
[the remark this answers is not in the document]
franco-zalamena-iterable: This condition is necessary for this code path, the user-controlled value is from a test on the emulator.

sumeruchat (changes requested):
AI Review

  1. Behavior change for setEmail(sameEmail) is a compat break. Apps that were calling it repeatedly as an idempotent no-op will now see extra syncs and backend
   calls. I don't think it's in the CHANGELOG — worth mentioning there since it's an observable runtime change.
  2. Deprecation warning log fires on every setAuthToken call. For apps that haven't migrated, that's potentially thousands of log lines per session. Consider
  IterableLogger.w once-per-session (static AtomicBoolean).
  3. updateAuthToken(null) silently stores null with no validation. Previously the old setAuthToken(null, true) path went through completeUserLogin which
  rejected null under JWT. This is fine given the split, but means updateAuthToken is a footgun — a caller on the public API could clear the token in a state
  where the SDK then makes unauthenticated requests. Maybe worth a @NonNull or at least no-op on null.
  4. Bugbash gate: since Franco asked to wait for bugbash, probably worth approving conditionally or holding review approval until QA signs off — especially
  given #1 is a behavioral change.

My review

This look a little bit dangerous to me as its behavior change and since no one is asking for this do we need to do this? @franco-zalamena-iterable Basically i see only potential risks of breaking existing behaviour with no upside.

franco-zalamena-iterable:
Meetiong arranged to discuss this

PR #1007, opened 2026-03-24
[SDK-406] Sdk fix flaky tests

franco-zalamena-iterable:
## 🔹 Jira Ticket(s) if any

* [SDK-406](https://iterable.atlassian.net/browse/SDK-406)

## ✏️ Description

Removing timeouts in some tests that can be causing flakiness. Moved some unit tests that were running as instrumental tests, added better description into ignored tests

[SDK-406]: https://iterable.atlassian.net/browse/SDK-406?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sumeruchat (changes requested):
@franco-zalamena-iterable Can you please explain the changes in background executor.

sumeruchat:
The suggested fix is to stop shutdownBackgroundExecutorAsync() from looking up the global backgroundExecutor at shutdown time.

  Right now it does this in iterableapi/src/main/java/com/iterable/iterableapi/IterableBackgroundInitializer.java:340:

  - grab the current static backgroundExecutor
  - spawn a thread
  - shut that executor down later

  That is the race. If test reset swaps in a new executor before a straggler calls this helper, the helper can shut down the new executor instead of the old
  one.

  The concrete change is:

  1. Change shutdownBackgroundExecutorAsync() to accept an ExecutorService argument.
  2. Pass the executor instance that actually ran the work.
  3. In reset, keep a local reference to the old executor, install the new one, then shut down only the old reference.

sumeruchat (approved):
LGTM

PR #1009, opened 2026-03-31
[SDK-386] embedded card improvements

franco-zalamena-iterable:
## 🔹 Jira Ticket(s) if any

* [SDK-386](https://iterable.atlassian.net/browse/SDK-386)

## ✏️ Description

Improvements to the Card Embedded View Image
*NEW* image scale type config option for the embedded view

The Card Embedded view currently has display issues for the buttons, with them showing up with a start padding that made them unaligned. To match the campaign preview **we need to check how to deal with this**

The image from the card embedded now has an aspect ratio field that can be configured by the new image scale type field

closes #997 

[SDK-386]: https://iterable.atlassian.net/browse/SDK-386?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sumeruchat (dismissed):
• The UI/layout changes look reasonable, but the public IterableEmbeddedViewConfig constructor change introduces a source-compatibility regression for
  existing Java consumers of the SDK. That makes the patch incorrect as a library change until compatibility is preserved.

  Review comment:

  - [P2] Preserve Java source compatibility for IterableEmbeddedViewConfig — /Users/sumeru.chatterjee/MyProjects/iterable-android-sdk/iterableapi-ui/src/main/
    java/com/iterable/iterableapi/ui/embedded/IterableEmbeddedViewConfig.kt:5-16
    If an app integrates this SDK from Java and already calls new IterableEmbeddedViewConfig(...), this change breaks compilation on upgrade because Kotlin
    default parameters do not create a shorter Java overload. Adding imageScaleType changes the only public constructor from 10 args to 11 args, so every
    existing Java call site now fails until it is updated manually.

iterableapi-ui/src/main/java/com/iterable/iterableapi/ui/embedded/IterableEmbeddedView.kt
```diff
@@ -179,6 +179,7 @@ class IterableEmbeddedView() : Fragment() {
             if(message.elements?.mediaURL?.isEmpty() == true) {
                 embeddedMessageImageView.visibility = View.GONE
             } else {
+                config?.imageScaleType?.let { embeddedMessageImageView.scaleType = it }
```
joaodordio: `imageScaleType` gets applied inside the `viewType != NOTIFICATION` block, so it affects both CARD and BANNER views. 

But banners use a fixed 80x80dp image, not the 16:9 container cards use. If we set  `FIT_CENTER` on a small thumbnail it will look weird no?


```suggestion
                if (viewType == IterableEmbeddedViewType.CARD) {
                    config?.imageScaleType?.let { embeddedMessageImageView.scaleType = it }
                }
```

joaodordio (changes requested):

iterableapi-ui/src/main/java/com/iterable/iterableapi/ui/embedded/IterableEmbeddedView.kt
franco-zalamena-iterable: We need to decide what will be the best approach, if we just apply to card maybe that can be confusing to clients since they are passing a imageScaleType to the configuration and it won't be applied if it is not a card, even though the config is not specific about cards.

franco-zalamena-iterable: I added some clarification in the latest commit

joaodordio: I like the DEFAULT_IMAGE_SCALE_TYPE idea!

joaodordio (approved):

PR #1015, opened 2026-04-06
[SDK-100] compose support

### LINKED ISSUE #461: compose compatibility
Hello, 
We have been developing a new mobile app using jetpack compose. I implemented the iterable SDK. However, when showing the in-app message, the popup is flicking.

franco-zalamena-iterable:
## 🔹 Jira Ticket(s) if any

* [SDK-100](https://iterable.atlassian.net/browse/SDK-100)

## ✏️ Description

- Add Dialog-based in-app message display for Compose apps using ComponentActivity (existing Fragment path unchanged)
- Extract shared logic into service classes (layout, animation, tracking, orientation, webview)
- Fix activity recreation state handling to prevent duplicate trackInAppOpen calls

### Why?
The SDK's in-app display requires FragmentActivity, which pure Compose apps don't use. This adds a Dialog-based fallback that activates automatically when the host activity is a ComponentActivity.

[SDK-100]: https://iterable.atlassian.net/browse/SDK-100?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sumeruchat (changes requested):
Thanks Franco — appreciate the work pushing toward Compose support. The dispatcher order (`FragmentActivity` first) is correct, and the new unit tests pass locally. That said, I think there are real correctness issues that need to land before this is mergeable. Four blockers below, plus some non-blocking items.

One framing note up front: the PR description says shared logic was extracted into services that **both** Fragment and Dialog consume, but `IterableInAppFragmentHTMLNotification.java` has zero diff vs `origin/master` in this branch — the new service classes are only consumed by the Dialog path. Worth updating the description (and maybe a follow-up to actually share them), because right now we have two parallel implementations and the Dialog one diverges from Fragment behavior in several places (see #2, #4, #5).

---

### 🚫 Blockers (P1)

**1. Dialog has no lifecycle binding → leaks + blocks future in-apps**

`IterableInAppDialogNotification` is a plain `Dialog(activity)` that holds a strong `Activity` reference and is stashed in a static singleton (`IterableInAppDialogNotification.kt:25-40,53-63`). There's no `LifecycleObserver` or `DefaultLifecycleObserver` on the host activity, and no fallback dismiss when the activity is destroyed.

On rotation / `finish()` while showing:
- Best case: in-app silently disappears with no close/removal callback fired.
- Worst case: the static `notification` field keeps the destroyed Activity + its window alive, and `isShowingInApp()` keeps returning true so all future in-apps are suppressed until process death.

`DialogFragment` solves this by being lifecycle-aware; a raw `Dialog` doesn't. Either wrap as a `DialogFragment`, attach a `LifecycleObserver` that calls `dismiss()` + clears the singleton on `ON_DESTROY`, or use `androidx.lifecycle` to listen to the host.

**2. Duplicate `trackInAppOpen` fix doesn't actually hold for the Dialog path**

The Fragment path persists `IN_APP_OPEN_TRACKED` in `onSaveInstanceState` (`IterableInAppFragmentHTMLNotification.java:284-315`) and skips tracking on recreated views. The Dialog path uses an instance boolean `inAppOpenTracked` only (`IterableInAppDialogNotification.kt:45,160-163`) and the "Drop ineffective `onSaveInstanceState` dance" commit removed the bridge. On rotation, the new Activity creates a new Dialog with `inAppOpenTracked = false`, so the message either disappears (data loss) or re-shows and re-tracks open (duplicate event). This is the exact bug the PR description says is fixed.

**3. Two in-apps can display simultaneously (split guards + non-atomic create)**

`isShowingInApp()` correctly ORs both singletons (`IterableInAppDisplayer.java:18-21`), but the static display helpers only check their own:
- `showIterableFragmentNotificationHTML` checks only `IterableInAppFragmentHTMLNotification.getInstance()` (`:68-79`)
- `showIterableDialogNotificationHTML` checks only `IterableInAppDialogNotification.getInstance()` (`:113-124`)

If a Dialog is showing and the current activity changes to a `FragmentActivity` (or any caller invokes the static helper directly), a Fragment in-app can stack on top of the Dialog. Reverse also true.

Separately, `IterableInAppDialogNotification.createInstance()` is a check-then-set on a non-atomic field (`:80-109`) — two concurrent callers can both pass the null check and create dialogs. The singleton churn in the commit history (add volatile → revert → "tighten guard") suggests this hasn't been fully nailed down. Recommend a single `synchronized` (or `AtomicReference.compareAndSet`) guard that checks both singletons together.

**4. Dialog resize is not functionally equivalent to Fragment**

Fragment debounces `runResizeScript()`, reads `webView.getContentHeight()`, and resizes the native WebView and window (`IterableInAppFragmentHTMLNotification.java:523-646`). The Dialog implementation only calls `webView.evaluateJavascript("window.resize()")` and has a `TODO` saying native resize isn't ported (`IterableInAppDialogNotification.kt:287-295`, `InAppWebViewService.kt:94-101`).

Top/bottom/center in-apps (i.e. anything non-fullscreen, and especially dynamic-height content) will render at wrong sizes for Compose hosts. This is a customer-visible regression vs. the Fragment path on the same content.

---

### Non-blocking (P2)

**5. Back-press skips the exit animation on the Dialog path**

Fragment back press calls `onBackPressed()` → `hideWebView()` (`IterableInAppFragmentHTMLNotification.java:159-165`), which runs the layout-aware exit animation, background fade, and delayed dismiss (`:466-508`). Dialog back press tracks/removes and calls `dismiss()` immediately (`IterableInAppDialogNotification.kt:185-203`), skipping `hideWebView()` entirely. Contradicts the "Port layout-aware enter/exit animations to Dialog for Fragment parity" commit for the BACK close path.

**6. Delayed animation runnables aren't cancelled / can fire after activity finishes**

`prepareToShowWebView()` and `hideWebView()` post delayed runnables on the WebView (`IterableInAppDialogNotification.kt:251-259, 320-335`) but `dismiss()` doesn't remove pending callbacks. The `context != null` guard inside them is meaningless — `Dialog.context` is non-null by construction (Kotlin compiler emits a warning for this). If the Activity finishes mid-animation, the runnable still touches the Dialog window/WebView. Recommend `webView.removeCallbacks(...)` on dismiss, or a lifecycle-bound cancel.

**7. Unintended public API surface expansion**

`IterableInAppDialogNotification` is declared as a public Kotlin class (its static `createInstance` / `getInstance` and constructor appear as public bytecode). The new `internal` service classes also compile to public JVM classes/methods that Java consumers can call (`javap` confirms). If these are implementation details, please hide them (`@RestrictTo`, package-private wrappers, or move to an internal package + `@JvmSynthetic` where applicable) before release.

---

### P3

**8. Misleading log on legacy helper**

`showIterableFragmentNotificationHTML(...)` now logs "Attempting dialog-based display" for non-`FragmentActivity` contexts (`IterableInAppDisplayer.java:81-84`), but it still just returns `false` and never attempts the Dialog path. The dispatcher in `showMessage()` already routes correctly, so this is harmless externally (the helper is package-private), but the log lies.

---

### Test gaps to consider for the fix-up

The four new test files mostly cover happy paths. None of these scenarios are exercised, and they're all in scope of the blockers above:

- Activity recreation during in-app display (both paths) and `trackInAppOpen` once-only semantics
- Fragment-vs-Dialog singleton collision when current activity changes mid-show
- Concurrent `createInstance` callers
- Activity `finish()` during enter/exit animation
- Back-press exit-animation parity

---

Static review + ran the new unit-test subset locally (`testDebugUnitTest --tests IterableInAppDialogNotificationTest --tests InAppTrackingServiceTest --tests InAppLayoutServiceTest --tests InAppOrientationServiceTest`) — all pass. Did not run instrumentation tests.

franco-zalamena-iterable:
I also added side to side comparisons to in-apps in the same app but one with fragments (using the fragment dialog logic) and the new with compose (using the new implementation)

sumeruchat (changes requested):
Thanks for the substantial rework, Franco — real progress here. The resize port (#4) is solid, the simultaneous-display guard (#3) is fixed for the normal main-thread case, back-press now plays the exit animation, and `dismiss()` is idempotent with proper runnable cancellation. But two of the original P1s aren't truly cleared yet, and there's a red test. Re-requesting changes — details below.

### 🚫 Blockers

**1. (P1) Rotation consumes the in-app message — `onDestroy` over-fires on configuration changes.**

`hostLifecycleObserver.onDestroy()` calls `processMessageRemoval()` + `dismiss()` on **every** host Activity destroy, including configuration-change recreation (rotation). For the normal `IN_APP` flow this is destructive:

- `IterableInAppManager.showMessage(... location=IN_APP)` marks the message for deletion after display
- `InAppTrackingService.removeMessage()` calls `inAppManager.removeMessage(message)` when `markedForDeletion && !consumed`
- `IterableInAppManager.removeMessage()` sets `consumed=true` and calls `api.inAppConsume(...)`

So **rotating while an in-app is open consumes the message locally _and_ server-side, the in-app disappears, and no close is tracked.** That's a worse failure mode than the original leak, because it corrupts message state.

The Fragment path deliberately avoids this: `IterableInAppFragmentHTMLNotification.onDestroy()` removes pending resize callbacks, then returns early when `getActivity().isChangingConfigurations()`, preserving state and using `savedInstanceState` to avoid duplicate `trackInAppOpen`. The Dialog path needs the same `isChangingConfigurations()` guard so a rotation tears down the window without consuming/removing the message.

This also means **#2 (duplicate `trackInAppOpen`) isn't genuinely fixed** — there's no duplicate only because the dialog is dropped/consumed instead of recreated. A plain `Dialog` can't restore across config change like `DialogFragment`, so you'll likely need to either (a) re-show + restore `inAppOpenTracked` after recreation, or (b) explicitly accept "dialog in-apps don't survive rotation" but at minimum not *consume* the message when they don't.

Note: the new test `hostDestroy_shouldCallProcessMessageRemoval_beforeDismiss()` currently codifies the destructive behavior and doesn't distinguish finish vs. configuration change — it'll need to assert the config-change case does NOT remove/consume.

**2. (P1) `iterableapi` unit suite is red.**

Full run is `657 tests, 1 failed`: `IterableInAppManagerTest.testJsonOnlyMessageDisplay` — expected `getMessages().size() == 1`, got `0`. (Focused dialog tests — 22 — all pass.) Please confirm whether this is fallout from the `master` merge (`1fdb066a`) or a branch issue, and get the suite green before merge. If it's a pre-existing master breakage, link the tracking issue so we know it's not from this PR.

### Non-blocking (P2 / P3)

**3. (P2) Display-mode parity drift from the master merge.**
After the merge, the Fragment path gained `IterableInAppDisplayMode` support (`resolveDisplayMode` / `configureSystemBarsForMode` / `shouldApplySystemBarInsets`). The Dialog path still only calls `InAppLayoutService.configureWindowFlags()` and never reads `IterableConfig.inAppDisplayMode`, so Compose/Dialog hosts won't honor `FORCE_EDGE_TO_EDGE` / `FOLLOW_APP_LAYOUT` / `FORCE_FULLSCREEN` / `FORCE_RESPECT_BOUNDS` the way Fragment hosts now do. Worth closing before we claim full Fragment/Dialog parity (follow-up is fine if scoped).

**4. (P2) `createInstance()` is still non-atomic.**
Both static helpers now correctly use `isAnyInAppShowing()` (good — the normal Fragment-vs-Dialog stacking bug is fixed). But `createInstance()` is still check-then-set on `notificationRef`: two concurrent callers can both observe null, create separate Dialogs + observers, and both show, with `notificationRef` pointing at only the last. If display is guaranteed main-thread this is minor — but the public `showMessage` APIs don't enforce/annotate main-thread-only. A `synchronized`/CAS around the check-and-publish would close it.

**5. (P3) `@RestrictTo` doesn't hide ABI.**
`javap` still shows `IterableInAppDialogNotification`, the service classes, and `InAppPadding` as public bytecode (Kotlin `internal` ≠ Java package-private; `@RestrictTo` is lint/advisory only). Acceptable if the goal is "not a supported public API" — but if you wanted true ABI hiding, this doesn't achieve it.

### Verified genuinely improved
- WeakReference + lifecycle observer means a *real* destroyed `ComponentActivity` no longer pins memory or permanently blocks future in-apps; `dismiss()` removes the observer and clears statics.
- `#4` resize mirrors `IterableInAppFragmentHTMLNotification.resize(float)` (main-looper debounce, contentHeight validation, 1dp epsilon, density scaling, CENTER/TOP/BOTTOM gravity rules, MATCH_PARENT window). `runResizeScript()` schedules on `Handler(Looper.getMainLooper())`, `resize()` marshals via `runOnUiThread` and bails if dismissed/not-showing, and `dismiss()` removes `pendingResizeRunnable` before nulling the handler.
- Back-press tracks click+close, calls `processMessageRemoval()`, then `hideWebView()` (exit animation), which ends in the now-idempotent `dismiss()`.
- The two new TODOs (tap-outside cancel not firing close/removal; orientation runnable not cancelled) accurately mirror existing Fragment-path gaps — fine to defer and fix together.

Static review + Codex ran the unit suite locally (focused dialog tests green; full suite 1 red as above).

sumeruchat:
@franco-zalamena-iterable — on the rotation-consumes-message blocker, here are two ship-able fixes with the tradeoff laid out. Pick whichever fits the timeline; both fully resolve the consume-on-rotation bug **without** reintroducing the three earlier P1s (Activity leak, permanent `isShowingInApp()` block, duplicate `trackInAppOpen`).

### First, an important correction

I'd originally assumed that if we just stop consuming on `onDestroy`, the in-app manager would re-display the unconsumed message on the recreated activity (with a brief flicker). **That's not what happens.** After the first display the message is `processed=true` + `read=true`, and `processMessages()` only shows messages matching `!processed && !consumed && IMMEDIATE && !read`. Rotation also doesn't reliably trigger a foreground re-sync. So on its own, removing the consume call makes the in-app **silently vanish on rotation**, leaving a `read`/`processed`/`markedForDeletion`-but-`unconsumed` local message.

Also confirmed: **not consuming on a real (non-config) destroy matches the Fragment path** — `IterableInAppFragmentHTMLNotification.onDestroy()` only clears statics; it never calls `processMessageRemoval()`. Consumption is tied to user close actions (back / link / delete), not lifecycle teardown. So whichever option you choose, `processMessageRemoval()` should come **out** of the lifecycle `onDestroy`.

---

## Option A — Minimal (≈3 lines)

`onDestroy` stops consuming and just tears down:

```kotlin
override fun onDestroy(owner: LifecycleOwner) {
    // Lifecycle teardown is not a user close. Do NOT consume/remove here — that
    // belongs to back-press / link-click / tap-outside. We only release the window
    // + static singleton so we neither leak the Activity nor block future in-apps.
    dismiss()
}
```

- ✅ No wrongful consume, no leak, no permanent block, no duplicate open.
- ⚠️ **Cost:** the in-app disappears on rotation instead of staying on screen. It's a correct, non-destructive behavior (the message is not consumed and remains available for a future session), but it's a UX regression vs. the Fragment path. Would need a documented "dialog in-apps don't persist across configuration changes" limitation.

---

## Option B — Full Fragment parity (≈80 lines)

`onDestroy` distinguishes configuration change from real destroy. On a config change it stages a short-lived **pending-restore state that holds no Activity/Dialog/WebView references** (only the data needed to rebuild the plain Dialog), dismisses immediately to avoid the leak/block, then posts a main-thread retry that waits for `IterableActivityMonitor.getCurrentActivity()` to point at the recreated host and re-creates + shows the Dialog seeded with `inAppOpenTracked = true` (so open is tracked exactly once). It preserves the user-visible Fragment behavior (in-app stays through rotation).

Guards baked in:
- `messageRemovalProcessed` flag so a rotation *during* a user-close exit animation can't resurrect a message the user already closed.
- Pending state is **one-shot**, with a 5s expiry, and is cleared on any `createInstance`, on real destroy, and on explicit removal — so it can't wrongly suppress a legitimately-new impression of the same message later.

<details>
<summary>Concrete diff for Option B</summary>

```diff
diff --git a/iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppDialogNotification.kt b/iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppDialogNotification.kt
@@
 import android.os.Bundle
 import android.os.Handler
 import android.os.Looper
+import android.os.SystemClock
@@
     private var orientationListener: OrientationEventListener? = null
     private var inAppOpenTracked: Boolean = false
     private var dismissed: Boolean = false
+    private var messageRemovalProcessed: Boolean = false
@@
     private val hostLifecycleObserver = object : DefaultLifecycleObserver {
         override fun onDestroy(owner: LifecycleOwner) {
-            processMessageRemoval()
+            val host = owner as? Activity
+            if (host?.isChangingConfigurations == true && !messageRemovalProcessed) {
+                stageRestoreAfterConfigChange()
+            } else {
+                clearPendingConfigChangeRestore()
+            }
             dismiss()
         }
     }
@@
         private const val DISMISS_DELAY_MS = 400L
         private const val RESIZE_DEBOUNCE_DELAY_MS = 200L
         private const val RESIZE_HEIGHT_EPSILON = 1.0f
+        private const val CONFIG_RESTORE_TIMEOUT_MS = 5_000L
+        private const val CONFIG_RESTORE_RETRY_MS = 100L
+        private const val CONFIG_RESTORE_MAX_ATTEMPTS = 50
+
+        private val mainHandler = Handler(Looper.getMainLooper())
+        @Volatile private var pendingConfigChangeRestore: PendingConfigChangeRestore? = null
+        @Volatile private var nextRestoreToken: Long = 1L
+
+        private data class PendingConfigChangeRestore(
+            val token: Long,
+            val htmlString: String,
+            val callbackOnCancel: Boolean,
+            val urlCallback: IterableHelper.IterableUrlCallback,
+            val location: IterableInAppLocation,
+            val message: IterableInAppMessage,
+            val backgroundAlpha: Double,
+            val padding: Rect,
+            val shouldAnimate: Boolean,
+            val inAppBgColor: IterableInAppMessage.InAppBgColor,
+            val openAlreadyTracked: Boolean,
+            val expiresAtUptimeMillis: Long
+        )
@@
             val newInstance = IterableInAppDialogNotification(
@@
                 InAppServices.orientation
             )
+
+            if (consumePendingOpenTracking(message.messageId)) {
+                newInstance.inAppOpenTracked = true
+            }
 
             lifecycle.addObserver(newInstance.hostLifecycleObserver)
@@
             return newInstance
         }
+
+        private fun consumePendingOpenTracking(messageId: String): Boolean {
+            val pending = pendingConfigChangeRestore ?: return false
+            pendingConfigChangeRestore = null // one-shot, even on a different message
+            if (SystemClock.uptimeMillis() > pending.expiresAtUptimeMillis) return false
+            return pending.message.messageId == messageId && pending.openAlreadyTracked
+        }
+
+        private fun clearPendingConfigChangeRestore() {
+            pendingConfigChangeRestore = null
+        }
+
+        private fun restoreAfterConfigChange(token: Long, attempt: Int = 0) {
+            val pending = pendingConfigChangeRestore ?: return
+            if (pending.token != token) return
+            if (SystemClock.uptimeMillis() > pending.expiresAtUptimeMillis) {
+                pendingConfigChangeRestore = null
+                return
+            }
+            if (getInstance() != null || IterableInAppFragmentHTMLNotification.getInstance() != null) {
+                pendingConfigChangeRestore = null
+                return
+            }
+
+            val activity = IterableActivityMonitor.getInstance().currentActivity
+            val lifecycle = (activity as? LifecycleOwner)?.lifecycle
+            if (activity == null || lifecycle == null || lifecycle.currentState == Lifecycle.State.DESTROYED) {
+                if (attempt < CONFIG_RESTORE_MAX_ATTEMPTS) {
+                    mainHandler.postDelayed({ restoreAfterConfigChange(token, attempt + 1) }, CONFIG_RESTORE_RETRY_MS)
+                } else {
+                    pendingConfigChangeRestore = null
+                }
+                return
+            }
+
+            pendingConfigChangeRestore = null
+            val restored = createInstance(
+                activity,
+                pending.htmlString,
+                pending.callbackOnCancel,
+                pending.urlCallback,
+                pending.location,
+                pending.message,
+                pending.backgroundAlpha,
+                pending.padding,
+                pending.shouldAnimate,
+                pending.inAppBgColor
+            ) ?: return
+            restored.inAppOpenTracked = pending.openAlreadyTracked
+            restored.show()
+        }
@@
     private fun processMessageRemoval() {
+        messageRemovalProcessed = true
+        clearPendingConfigChangeRestore()
         trackingService.removeMessage(message)
     }
+
+    private fun stageRestoreAfterConfigChange() {
+        val callback = clickCallback ?: return
+        val loc = location ?: IterableInAppLocation.IN_APP
+        val token = nextRestoreToken++
+        pendingConfigChangeRestore = PendingConfigChangeRestore(
+            token = token,
+            htmlString = htmlString ?: "",
+            callbackOnCancel = callbackOnCancel,
+            urlCallback = callback,
+            location = loc,
+            message = message,
+            backgroundAlpha = backgroundAlpha,
+            padding = Rect(insetPadding),
+            shouldAnimate = shouldAnimate,
+            inAppBgColor = IterableInAppMessage.InAppBgColor(inAppBackgroundColor, inAppBackgroundAlpha),
+            openAlreadyTracked = inAppOpenTracked,
+            expiresAtUptimeMillis = SystemClock.uptimeMillis() + CONFIG_RESTORE_TIMEOUT_MS
+        )
+        mainHandler.post { restoreAfterConfigChange(token) }
+    }
 }
```

Notes:
- `currentActivity` is Kotlin property syntax for `getCurrentActivity()`.
- The restore state intentionally holds no Activity/Dialog/WebView. It does hold the url callback (matching the Fragment path's existing static callback behavior), but only with the short timeout.
- Field names like `inAppBackgroundColor` / `inAppBackgroundAlpha` are placeholders — wire them to whatever the constructor already stores for bg color/alpha.

</details>

---

### Recommendation
If we want true parity (in-app survives rotation), **Option B** is the clean solution. If timeline is tight, **Option A** is a correct, much smaller fix as long as we document the "doesn't persist across rotation" limitation. Either way, also worth: get the red `IterableInAppManagerTest.testJsonOnlyMessageDisplay` green (confirm whether it's from the master merge), and a regression test asserting **config-change destroy does NOT consume** while **real destroy still clears the singleton**.

Happy to pair on whichever you pick.

franco-zalamena-iterable:
I think just removing the in-app without re show is the safest approach now, the new dialog is already a big change and we need to have real life testing before committing to add more things to it.

I will add a changelog entry for it stating the difference, there will already be a planned ticket to convert the fragment dialog to the new services approach (makes both dialogs have the same exact behavior via same code)

will link the new ticket when it's created also in this comment

franco-zalamena-iterable:
FORCE_FULLSCREEN flag in apps

| Display mode | Fragment | Compose |
| --- | --- | --- |
| Top      | fragment_top_ffs | compose_top_ffs |
| Center   | fragment_center_ffs | compose_center_ffs |
| Bottom   | fragment_bottom_ffs | compose_bottom_ffs |
| Full     | fragment_full_ffs | compose_full_ffs |


FORCE_RESPECT_BOUNDS flag in apps

| Display mode | Fragment | Compose |
| --- | --- | --- |
| Top      | fragment_top_frb | compose_top_frb |
| Center   | fragment_center_frb | compose_center_frb |
| Bottom   | fragment_bottom_frb | compose_bottom_frb |
| Full     | fragment_full_frb | compose_full_frb |


FORCE_EDGE_TO_EDGE flag in apps

| Display mode | Fragment | Compose |
| --- | --- | --- |
| Top      | fragment_top_fee | compose_top_fee |
| Center   | fragment_center_fee | compose_center_fee |
| Bottom   | fragment_bottom_fee | compose_bottom_fee |
| Full     | fragment_full_fee | compose_full_fee |


There are differences on the system bar colors for compose dialog because, since it is not a fragment we don't have access to the information necessary to make it the same as fragment

franco-zalamena-iterable:
FORCE_FULLSCREEN flag in apps

| Display mode | Master | PR |
| --- | --- | --- |
| Top      | master_top_ffs | pr_top_ffs |
| Center   | master_center_ffs | pr_center_ffs |
| Bottom   | master_bottom_ffs | pr_bottom_ffs |
| Full     | master_full_ffs | pr_full_ffs |


FORCE_RESPECT_BOUNDS flag in apps

| Display mode | Master | PR |
| --- | --- | --- |
| Top      | master_top_frb | pr_top_frb |
| Center   | master_center_frb | pr_center_frb |
| Bottom   | master_bottom_frb | pr_bottom_frb |
| Full     | master_full_frb | pr_full_frb |


FORCE_EDGE_TO_EDGE flag in apps

| Display mode | Master | PR |
| --- | --- | --- |
| Top      | master_top_fee | pr_top_fee |
| Center   | master_center_fee | pr_center_fee |
| Bottom   | master_bottom_fee | pr_bottom_fee |
| Full     | master_full_fee | pr_full_fee |

sumeruchat (approved):
Thanks Franco — the revert of `IterableInAppFragmentHTMLNotification.java` cleanly resolves the Fragment-path regression. Verified locally that `git diff eff6c4a2 HEAD -- iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFragmentHTMLNotification.java` is empty, so existing Fragment customers see no change from the display-mode work. Combined with the rotation fix and the now-green unit suite, the original blockers are all cleared. Approving.

### ✅ Cleared
- **Rotation no longer consumes the message.** `hostLifecycleObserver.onDestroy()` skips `processMessageRemoval()` on `isChangingConfigurations()`. Singleton, observer, callbacks, WeakReference all still cleaned up on every destroy — leak + permanent `isShowingInApp()` block remain fixed.
- **Unit suite green.** Full `:iterableapi:testDebugUnitTest` was 659/0/23 at `9fabf692`; the only change since is a single-file revert that cannot regress it.
- **Fragment path is byte-identical to pre-extraction.** No backward-compat risk for existing Fragment users.

### Non-blocking follow-ups (Dialog-only — fine to fold into the planned Fragment+Dialog services-unification ticket)

1. **`InAppDisplayModeService` is now a Dialog-internal helper rather than shared.** After the revert, only the Dialog consumes it. Worth a comment in the class KDoc that it's Dialog-only today, with a pointer to the follow-up ticket where Fragment will migrate onto it.

2. **`FORCE_RESPECT_BOUNDS` window flags differ from Fragment.** Dialog uses `setDecorFitsSystemWindows(window, false) + FLAG_LAYOUT_NO_LIMITS + manual padding`; Fragment uses `setDecorFitsSystemWindows(window, true)`. The visual result *should* be equivalent via the manual padding from `applyContentInsetsForMode`, but the flag semantics aren't identical. Easy follow-up: drop `FLAG_LAYOUT_NO_LIMITS` and use `setDecorFitsSystemWindows(true)` to match Fragment exactly.

3. **`restoreHostSystemBars()` on dismiss doesn't snapshot prior host state.** If a Compose host was intentionally immersive before the in-app showed, dismissing in `FORCE_FULLSCREEN` will force the system bars back. Capture the host's bar visibility on `onCreate` and restore exactly that on dismiss.

4. **Consume-on-real-destroy diverges from Fragment.** On a non-config destroy, Dialog calls `processMessageRemoval()` → `removeMessage()` → `consumed=true` + `inAppConsume(...)`; Fragment never consumes on lifecycle teardown (it only consumes on user close paths). This is a small but real telemetry/state divergence — a Compose-hosted in-app that gets backed out without user interaction will fire an `inAppConsume` that the Fragment equivalent wouldn't. Two clean options:
   - Match Fragment: also skip `processMessageRemoval()` on the non-config branch and accept the same read/processed/markedForDeletion/unconsumed zombie state the Fragment leaves.
   - Or deliberately change the impression semantics for *both* renderers in a separate ticket, with a CHANGELOG note and telemetry sign-off.

   Of these residuals, this is the one most worth handling sooner rather than later because it affects what gets sent to the server.

5. **Display-mode tests for the Dialog.** The new test additions cover singleton cleanup, host destroy/config-change non-consume, Fragment/Dialog guard, back-press telemetry, observer removal, and pending-resize cancellation, but they don't cover the four `IterableInAppDisplayMode` branches on the Dialog, Fragment/Dialog system-bar parity, or `restoreHostSystemBars` preserving a pre-existing immersive host. Targeted Robolectric tests around service window flags/insets would close the loop.

6. **Residuals from prior rounds (still open, low priority):**
   - `IterableInAppDialogNotification.createInstance()` is still a non-atomic check-then-set on `notificationRef` — main-thread-only in practice, but no annotation enforces it.
   - `@RestrictTo(LIBRARY_GROUP)` doesn't actually hide the new classes from public JVM bytecode (`javap` still shows them as public). Acceptable if the intent is "not a supported API" via lint, not if we wanted true ABI hiding.

### Reminder
The Option A behavior (Dialog in-app vanishes on rotation rather than being preserved) is the agreed compromise — please make sure the CHANGELOG entry documents the Fragment/Dialog difference, and that the planned Fragment+Dialog services-unification ticket is filed and linked from the CHANGELOG / PR description.

Nice work pulling this back into shape after a rough middle.

Issue #485, opened 2023-01-06
IterableInAppManager.java line 330 com.iterable.iterableapi.IterableInAppManager.syncWithRemoteQueue


ravisaini1990:
Fatal Exception: java.lang.NullPointerException: Attempt to invoke virtual method 'boolean com.iterable.iterableapi.IterableInAppMessage.isRead()' on a null object reference
       at com.iterable.iterableapi.IterableInAppManager.syncWithRemoteQueue(IterableInAppManager.java:330)
       at com.iterable.iterableapi.IterableInAppManager.access$000(IterableInAppManager.java:32)
       at com.iterable.iterableapi.IterableInAppManager$1.execute(IterableInAppManager.java:178)
       at com.iterable.iterableapi.IterableRequestTask.onPostExecute(IterableRequestTask.java:296)
       at com.iterable.iterableapi.IterableRequestTask.onPostExecute(IterableRequestTask.java:28)
       at android.os.AsyncTask.finish(AsyncTask.java:695)
       at android.os.AsyncTask.-wrap1()
       at android.os.AsyncTask$InternalHandler.handleMessage(AsyncTask.java:712)
       at android.os.Handler.dispatchMessage(Handler.java:105)
       at android.os.Looper.loop(Looper.java:164)
       at android.app.ActivityThread.main(ActivityThread.java:6797)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.android.internal.os.Zygote$MethodAndArgsCaller.run(Zygote.java:240)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:772)

Other Statistics of Crash From Firebase- 

Crash on Android Versions - 
50% Android 8
30% Android 9
10% Android 11
10% Android 13
-----
Crash on Devices - 
50% motorola
30% samsung
10% Sony
10% OnePlus
-----------------
60% In background
0% Rooted
100% Proximity on

joaodordio:
Hey @ravisaini1990 are you able to share more context on the issue you are having please?

PR #1017, opened 2026-04-07
#485 - Fix NPE in syncWithRemoteQueue when message removed during sync

franco-zalamena-iterable:
## Summary
- Adds a null check for `localMessage` after re-fetching from storage in `IterableInAppManager.syncWithRemoteQueue`
- Guards against a race condition where the message is removed between the existence check (`storage.getMessage() != null`) and the subsequent retrieval, which caused a `NullPointerException` when calling `isRead()` on a null reference

## Test plan
- [ ] Verify that `syncWithRemoteQueue` no longer throws NPE when a message is removed from storage between the existence check and re-fetch
- [ ] Verify that the normal sync flow (message exists and is updated) still works correctly
- [ ] Verify that read-state syncing from remote to local still functions as expected
- [ ] Run existing unit tests for `IterableInAppManager` to confirm no regressions

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppManager.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppManager.java
@@ -339,20 +339,20 @@ public class IterableInAppManager implements IterableActivityMonitor.AppStateCal
         Map<String, IterableInAppMessage> remoteQueueMap = new HashMap<>();
 
         for (IterableInAppMessage message : remoteQueue) {
+            if (message == null) {
+                continue;
+            }
+
             remoteQueueMap.put(message.getMessageId(), message);
 
-            boolean isInAppStored = storage.getMessage(message.getMessageId()) != null;
+            IterableInAppMessage localMessage = storage.getMessage(message.getMessageId());
 
-            if (!isInAppStored) {
+            if (localMessage == null) {
                 storage.addMessage(message);
                 onMessageAdded(message);
 
                 changed = true;
-            }
-
-            if (isInAppStored) {
-                IterableInAppMessage localMessage = storage.getMessage(message.getMessageId());
-
+            } else {
                 boolean shouldOverwriteInApp = !localMessage.isRead() && message.isRead();
 
                 if (shouldOverwriteInApp) {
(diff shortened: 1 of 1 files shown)


franco-zalamena-iterable:
## PR Analysis

**Problem:** `NullPointerException` at `IterableInAppManager.syncWithRemoteQueue` line 330 because `localMessage.isRead()` is called on a null reference. The original code calls `storage.getMessage()` twice — once for the existence check and once to retrieve the object — creating a TOCTOU (time-of-check-time-of-use) race where the message can be removed between the two calls (e.g., via `removeMessage()`, which is `synchronized` while `syncWithRemoteQueue` is not).

**Ideal fix plan:**
- Collapse the two `storage.getMessage()` calls into a single call and store the result, eliminating the TOCTOU race
- Use `if/else` instead of two separate `if` blocks with inverted conditions (dead-obvious code simplification)
- Add a null guard for `remoteQueue` entries as defensive programming (the caller already filters nulls, so this is belt-and-suspenders)
- Add a null guard for `storage.getMessages()` entries in the second loop for consistency
- Consider making `syncWithRemoteQueue` `synchronized` to fully close the race with `removeMessage` — without it, a concurrent `removeMessage` call could still modify storage mid-iteration
- Add a unit test that exercises the race condition (e.g., a storage mock that returns null on the second call)

**What the PR did:**
- Collapsed the double `storage.getMessage()` call into a single call with an `if/else` — directly fixing the TOCTOU bug
- Added a null guard for entries in the `remoteQueue` iteration
- Added a null guard for entries in the `storage.getMessages()` iteration
- Extracted `storage.getMessages()` into a local variable for the second loop (minor readability improvement)

**Assessment:**
- Root cause identified: **yes** — the double fetch from storage with a race window was correctly identified
- Fix correctness: **partially correct** — the TOCTOU gap is eliminated for the `isRead()` call, but `syncWithRemoteQueue` is still not `synchronized`, so a concurrent `removeMessage` can still mutate storage mid-iteration (e.g., ConcurrentModificationException on the `storage.getMessages()` loop, or a message disappearing between the `containsKey` check and `removeMessage` call). The proper fix should also add `synchronized` to this method to match `removeMessage`.
- Missed: (1) `syncWithRemoteQueue` should be `synchronized` to fully prevent concurrent modification from `removeMessage` or other synchronized methods. (2) No unit test was added to cover the null/race scenario.
- Wrong assessment: **none** — the PR description accurately describes the problem.
- Tests: **needed but missing** — a test with a mock storage that returns null on a second `getMessage()` call would validate the fix and prevent regression.

PR #1018, opened 2026-04-07
Add method to reset notification payload data

franco-zalamena-iterable:
## Summary
- Add clearPayloadData() method to IterableApi
- Allows users to reset notification payload after processing

## Test plan
- [ ] Set payload data via notification, call clearPayloadData(), verify getPayloadData() returns null
- [ ] Verify normal notification payload flow still works

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
@@ -1245,6 +1245,15 @@ public class IterableApi {
         return _payloadData;
     }
 
+    /**
+     * Resets the notification payload data by setting it to null.
+     * Call this method after you have finished processing the notification payload
+     * to prevent stale data from being returned by {@link #getPayloadData()}.
+     */
+    public void clearPayloadData() {
+        _payloadData = null;
+    }
+
     public void setDeviceAttribute(String key, String value) {
         if (key == null || value == null) {
             IterableLogger.e(TAG, "setDeviceAttribute: key and value must not be null");


franco-zalamena-iterable:
## PR Analysis

**Problem:** `IterableApi.getInstance().getPayloadData()` retains notification payload data indefinitely after a push notification is opened. There is no public API to reset it, so consumers cannot distinguish between "a notification was just opened" and "stale data from a previous notification" once they have finished processing it.

**Ideal fix plan:**
- Add a public method to clear `_payloadData` (set it to `null`) so consumers can reset it after processing
- Alternatively, the SDK could auto-clear payload data after it has been consumed (e.g., after the first `getPayloadData()` call or after `onResume` completes), but that would be a breaking behavioral change, so a manual clear method is the safer, non-breaking approach
- Add a unit test that sets payload data, calls the new clear method, and asserts `getPayloadData()` returns `null`

**What the PR did:**
- Added a public `clearPayloadData()` method that sets `_payloadData = null`
- Added Javadoc explaining when to call it
- No tests included (only a manual test plan in the PR description)

**Assessment:**
- Root cause identified: **yes** -- the field is set on notification open and never cleared, with no public API to reset it
- Fix correctness: **correct** -- a public nulling method is the minimal, non-breaking solution; naming (`clearPayloadData`) is consistent with common Android SDK conventions
- Missed: **unit tests** -- a simple test (set bundle via the package-private `setPayloadData`, call `clearPayloadData`, assert `getPayloadData()` is `null`) should be straightforward and would prevent regressions. The PR lists manual test steps but does not add automated tests.
- Wrong assessment: **none**

PR #1019, opened 2026-04-07
Improve notification icon fallback for standard Android conventions

franco-zalamena-iterable:
## Summary
- Update notification icon fallback to check Firebase, Expo, and standard Android icon conventions
- Prevents white square icons on Android 5.0+ when iterable_notification_icon is not configured

## Test plan
- [ ] Test with iterable_notification_icon configured (should use it)
- [ ] Test with Firebase default icon configured (should use it as fallback)
- [ ] Test with @drawable/notification_icon (Expo convention)
- [ ] Test with no icon configured (should fall back to launcher icon)

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableNotificationHelper.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableNotificationHelper.java
@@ -436,6 +436,43 @@ class IterableNotificationHelper {
                         context.getPackageName());
             }
 
+            //Check Firebase FCM default notification icon
+            if (iconId == 0) {
+                try {
+                    ApplicationInfo info = context.getPackageManager().getApplicationInfo(context.getPackageName(), PackageManager.GET_META_DATA);
+                    if (info.metaData != null) {
+                        iconId = info.metaData.getInt("com.google.firebase.messaging.default_notification_icon", 0);
+                        if (iconId != 0) {
+                            IterableLogger.d(IterableNotificationBuilder.TAG, "Using Firebase default notification icon");
+                        }
+                    }
+                } catch (PackageManager.NameNotFoundException e) {
+                    IterableLogger.w(IterableNotificationBuilder.TAG, e.toString());
+                }
+            }
+
+            //Check Expo/React Native convention: @drawable/notification_icon
+            if (iconId == 0) {
+                iconId = context.getResources().getIdentifier(
+                        "notification_icon",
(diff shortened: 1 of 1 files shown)


franco-zalamena-iterable:
## PR Analysis

**Problem:** When `iterable_notification_icon` is not configured, the SDK falls back to the app launcher icon, which produces white squares on Android 5.0+ because launcher icons are full-color rather than the required silhouette-style. Expo/Firebase/React Native apps already provide properly formatted notification icons through standard resource conventions, but the SDK ignores them.

**Ideal fix plan:**
- In `getIconId()`, after the existing checks for manifest metadata and code-configured icon, add fallback lookups before the launcher icon default
- Check `com.google.firebase.messaging.default_notification_icon` from manifest metadata
- Check `@drawable/notification_icon` (Expo/RN convention)
- Check `@drawable/ic_notification` (common Android convention)
- Preserve the existing launcher icon as the final fallback
- Add unit tests covering each fallback step in isolation (icon resolution is deterministic and testable with Robolectric)

**What the PR did:**
- Added the three fallback checks in the correct order between the existing code-configured icon check and the launcher icon fallback
- Used `getResources().getIdentifier()` for drawable lookups and `metaData.getInt()` for the Firebase manifest entry
- Added debug/warning logging for each fallback path
- Single file change, no tests

**Assessment:**
- Root cause identified: **yes** -- the launcher icon fallback is the problem, and conventional icon resources are available but not checked
- Fix correctness: **correct** -- the fallback chain is inserted at the right place, the lookup mechanisms are correct, and the order (Firebase manifest > Expo convention > common convention > launcher icon) is reasonable
- Missed:
  - **Unit tests are missing.** The existing `IterableNotificationTest.java` already uses Robolectric and sets `getApplicationInfo().icon` for testing. Tests should verify: (1) `iterable_notification_icon` still takes priority, (2) Firebase metadata icon is used when present, (3) `notification_icon` drawable is found, (4) `ic_notification` drawable is found, (5) launcher icon is still the final fallback when none of the above exist. These are straightforward to write with Robolectric shadow resources.
  - The Firebase metadata lookup creates a second `getApplicationInfo()` call with `GET_META_DATA` that duplicates the one already done a few lines above. Minor, but it could reuse the existing `info` variable if it were scoped appropriately (or the two metadata reads could be combined in a single try block).
- Wrong assessment: **none** -- the approach matches the issue request accurately
- Tests: **needed but missing** -- the fallback chain has multiple branches that should be verified; the project already has a Robolectric test infrastructure that makes this straightforward

PR #1022, opened 2026-04-07
Simplify IterableEmbeddedManager.handleEmbeddedClick API

franco-zalamena-iterable:
## Summary
- Add new simplified `handleEmbeddedClick(clickedUrl: String)` method that only requires the URL parameter
- Deprecate old `handleEmbeddedClick(message, buttonIdentifier, clickedUrl)` method with `@Deprecated` annotation and `ReplaceWith` hint
- Add KDoc documentation to both methods
- Old method delegates to new one internally

## Test plan
- [ ] Verify new method works correctly with action://, itbl://, and standard URLs
- [ ] Verify old deprecated method still works and delegates to new method
- [ ] Check deprecation warning appears in IDE when using old method

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableEmbeddedManager.kt
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableEmbeddedManager.kt
@@ -159,8 +159,16 @@ public class IterableEmbeddedManager : IterableActivityMonitor.AppStateCallback 
         }
     }
 
-    fun handleEmbeddedClick(message: IterableEmbeddedMessage, buttonIdentifier: String?, clickedUrl: String?) {
-        if ((clickedUrl != null) && clickedUrl.toString().isNotEmpty()) {
+    /**
+     * Handles an embedded message click by processing the given URL.
+     *
+     * Supports `action://`, `itbl://`, and standard URLs. Action and itbl scheme URLs
+     * are routed to the custom action handler, while other URLs are opened directly.
+     *
+     * @param clickedUrl The URL associated with the embedded message click.
+     */
+    fun handleEmbeddedClick(clickedUrl: String) {
+        if (clickedUrl.isNotEmpty()) {
             if (clickedUrl.startsWith(IterableConstants.URL_SCHEME_ACTION)) {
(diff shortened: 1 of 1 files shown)


franco-zalamena-iterable:
## PR Analysis

**Problem:** `IterableEmbeddedManager.handleEmbeddedClick(message, buttonIdentifier, clickedUrl)` requires `message` and `buttonIdentifier` parameters that are never used internally, making the API unnecessarily cumbersome. No documentation exists to clarify this.

**Ideal fix plan:**
- Add a new overload `handleEmbeddedClick(clickedUrl: String)` containing the actual logic
- Deprecate the old 3-parameter method and have it delegate to the new one
- Add KDoc to both methods
- Update internal callers (e.g., `IterableEmbeddedView`) to use the new method so they don't trigger deprecation warnings
- Add unit tests for the new method (at minimum: action:// URL, itbl:// URL, standard URL, empty string)

**What the PR did:**
- Added `handleEmbeddedClick(clickedUrl: String)` with the core logic
- Deprecated the old method with `@Deprecated` + `ReplaceWith` annotation
- Old method delegates to new one via `clickedUrl ?: ""`
- Added KDoc to both methods

**Assessment:**
- Root cause identified: **yes** -- the unused parameters are correctly identified
- Fix correctness: **partially correct** -- the new overload and deprecation are done well, but there is an omission (see below)
- Missed:
  - `IterableEmbeddedView.kt` (in `iterableapi-ui`) calls the old method in both `setDefaultAction` and `setButton`. These internal SDK callers should be updated to use the new simplified method. Shipping a deprecated method that your own SDK still calls is not ideal -- consumers will see the deprecated API in the SDK's own view code, and IDE warnings will fire inside the SDK itself.
  - No unit tests were added. The `handleEmbeddedClick` method has branching logic (action:// vs itbl:// vs standard URL vs empty string) that should be covered by tests. The PR's own test plan lists manual verification, but automated tests are appropriate here.
- Wrong assessment: **none** -- the analysis and approach are sound
- Tests: **needed but missing** -- branching URL-handling logic with three distinct code paths warrants unit test coverage

PR #1023, opened 2026-04-07
Replace RuntimeException with graceful handling when SDK not initialized

franco-zalamena-iterable:
## Summary
- Replace undocumented RuntimeExceptions with logged warnings and null returns
- SDK methods now fail gracefully when called before initialization

## Test plan
- [ ] Call getInAppManager() before initialization - should log warning, not crash
- [ ] Call getInAppManager() after initialization - should work normally
- [ ] Verify other similar methods also handle gracefully

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
@@ -998,20 +998,22 @@ public class IterableApi {
      * Make sure the Iterable API is initialized before calling this method.
      * @return {@link IterableInAppManager} instance
      */
-    @NonNull
+    @Nullable
     public IterableInAppManager getInAppManager() {
         if (inAppManager == null) {
-            throw new RuntimeException("IterableApi must be initialized before calling getInAppManager(). " +
+            IterableLogger.e(TAG, "IterableApi must be initialized before calling getInAppManager(). " +
                     "Make sure you call IterableApi#initialize() in Application#onCreate");
+            return null;
         }
         return inAppManager;
     }
 
-    @NonNull
+    @Nullable
     public IterableEmbeddedManager getEmbeddedManager() {
         if (embeddedManager == null) {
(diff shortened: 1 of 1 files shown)


franco-zalamena-iterable:
## PR Analysis

**Problem:** Calling `getInAppManager()` or `getEmbeddedManager()` before SDK initialization throws an undocumented `RuntimeException`, crashing the app. The user requests graceful failure with a log message instead.

**Ideal fix plan:**
- Replace `RuntimeException` with an error log in `getInAppManager()` and `getEmbeddedManager()`
- Return `@Nullable` and update annotation accordingly
- **Guard all internal call sites** that chain methods on the return value (e.g., `getInAppManager().syncInApp()`, `getInAppManager().getMessageById(...)`, `getEmbeddedManager().syncMessages()`) with null checks to prevent `NullPointerException`
- Add unit tests that call these methods before initialization and verify no crash occurs
- Consider the same pattern for any other public methods that throw undocumented `RuntimeException`

**What the PR did:**
- Replaced `RuntimeException` with `IterableLogger.e()` + `return null` in both `getInAppManager()` and `getEmbeddedManager()`
- Changed `@NonNull` to `@Nullable` on both methods

**Assessment:**
- Root cause identified: **yes** — the `RuntimeException` on uninitialized access is the root cause
- Fix correctness: **partially correct** — the two methods themselves are fixed, but the change introduces new `NullPointerException` crashes at every internal call site that chains on the result
- Missed:
  - **Critical:** At least 7 internal call sites now crash with NPE instead of RuntimeException. Examples: `getInAppManager().syncInApp()` (line ~479), `getEmbeddedManager().syncMessages()` (line ~480), `getInAppManager().getMessageById(messageId)` (lines ~1301, 1317, 1859, 1876, 1907). Each of these needs a null check or early return.
  - No tests were added. At minimum, tests should verify that calling `getInAppManager()` and `getEmbeddedManager()` before initialization returns null without crashing, and that downstream public methods like `inAppConsume()` also handle the null gracefully.
- Wrong assessment: none — the direction is correct, the execution is incomplete

Issue #477, opened 2022-10-24
Android Inbox Header Customisation


mkhurshid1:
Hi,

I want to know if there's any way we can customise the header/ action bar of android? For example I need the inbox header to be centre aligned and we need a close button on the header. Is that possible? I am unable to find any thing related to this in docs?

saintjab:
Any update on this please? I am also looking for such.

joaodordio:
@saintjab can you please submit this feature request directly to your assigned CSM?

PR #1029, opened 2026-04-07
Add inbox header customization options

franco-zalamena-iterable:
## Summary
- Add `IterableInboxCustomization` builder class for configuring inbox toolbar appearance
- Support center-aligned title and close button (with custom click handler) in `IterableInboxActivity`
- Configurable via programmatic API (`setInboxCustomization`) or intent extras (`EXTRA_TITLE_CENTER_ALIGNED`, `EXTRA_SHOW_CLOSE_BUTTON`)

## Test plan
- [ ] Test center-aligned inbox title via `IterableInboxCustomization.Builder().setTitleCenterAligned(true)`
- [ ] Test close button visibility and click handling
- [ ] Test intent extras for title alignment and close button
- [ ] Verify default behavior unchanged (toolbar shows with standard left-aligned title, no close button)

### DIFF
--- a/iterableapi-ui/src/main/AndroidManifest.xml
+++ b/iterableapi-ui/src/main/AndroidManifest.xml
@@ -5,7 +5,8 @@
         <activity
             android:name=".inbox.IterableInboxActivity"
             android:exported="false"
-            android:label="Inbox" />
+            android:label="Inbox"
+            android:theme="@style/IterableInboxActivityTheme" />
         <activity
             android:name=".inbox.IterableInboxMessageActivity"
             android:exported="false"/>
--- a/iterableapi-ui/src/main/java/com/iterable/iterableapi/ui/inbox/IterableInboxActivity.java
+++ b/iterableapi-ui/src/main/java/com/iterable/iterableapi/ui/inbox/IterableInboxActivity.java
@@ -2,8 +2,12 @@ package com.iterable.iterableapi.ui.inbox;
 
 import android.content.Intent;
 import android.os.Bundle;
+import android.view.Gravity;
+import android.widget.TextView;
+
 import androidx.annotation.Nullable;
 import androidx.appcompat.app.AppCompatActivity;
+import androidx.appcompat.widget.Toolbar;
 
 import com.iterable.iterableapi.IterableConstants;
 import com.iterable.iterableapi.IterableLogger;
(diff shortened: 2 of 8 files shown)


franco-zalamena-iterable:
## PR Analysis

**Problem:** Users have no way to customize the inbox activity header/toolbar -- specifically, there is no API to center-align the title or add a close button. This is a feature request (issue #477), not a bug.

**Ideal fix plan:**
- Add an explicit `Toolbar` to the inbox activity layout (replacing the default action bar) so it can be customized
- Create a configuration/customization class with a builder pattern for title text, title alignment, close button visibility, and close button behavior
- Expose both a programmatic API and intent extras so the customization works for different integration patterns
- Apply a `NoActionBar` theme to the activity to avoid conflicts between the custom toolbar and the default action bar
- Add unit tests covering: default behavior unchanged, center-aligned title, close button shown with default finish behavior, close button with custom listener, intent extras, and programmatic API taking precedence over extras
- Add a close icon drawable
- Update or add documentation

**What the PR did:**
- Replaced the default action bar with an explicit `Toolbar` in the layout, wrapped in a `LinearLayout`
- Created `IterableInboxCustomization` with a builder pattern covering title, center alignment, close button, and close button listener
- Exposed a static `setInboxCustomization()` method and intent extras (`EXTRA_TITLE_CENTER_ALIGNED`, `EXTRA_SHOW_CLOSE_BUTTON`)
- Applied a `NoActionBar` theme (`IterableInboxActivityTheme`) to avoid action bar conflicts
- Added a close icon vector drawable (`ic_close_black_24dp`)
- Declared custom styleable attributes in `attrs.xml` (though these are not actually read anywhere in the code)

**Assessment:**
- Root cause identified: yes -- there was no customization surface for the inbox toolbar
- Fix correctness: partially correct
- Missed:
  - **No tests were added.** The customization logic (`configureToolbar`, builder, intent extras vs. programmatic precedence) should have unit/instrumentation tests.
  - The `attrs.xml` declares `IterableInboxActivity` styleable attributes (`iterableInboxTitleCenterAligned`, `iterableInboxShowCloseButton`, `iterableInboxTitle`) but they are **never read via `obtainStyledAttributes`** anywhere. They are dead code -- either remove them or wire them up.
  - `pendingCustomization` is a **static mutable field** that acts as a one-shot communication channel between the caller and `onCreate`. This is fragile: it will not survive activity recreation (e.g., configuration change after the first creation consumes and nulls it), and it creates a subtle global state dependency. A more robust approach would be to serialize customization options into the intent extras (or a `Bundle`), so they survive the full activity lifecycle.
  - The hardcoded `IterableInboxActivityTheme` with a white `colorPrimary` **overrides the host app's theme**. If an app already has a custom theme with specific primary colors, this will break their look. It would be better to inherit from the app's theme or make the toolbar background configurable rather than forcing white.
  - The toolbar text color uses `?android:attr/textColorPrimary`, which may not contrast well against the hardcoded white background depending on the device/theme.
- Wrong assessment: none
- Tests: needed but missing

PR #1031, opened 2026-04-07
Fix in-app message WebView constant redraw issue

franco-zalamena-iterable:
## Summary
- Fix continuous layout redraws in in-app message WebView
- Prevent unnecessary requestLayout/invalidate cycles for static content

## Test plan
- [ ] Verify in-app messages display correctly
- [ ] Monitor logcat to confirm no excessive layout passes
- [ ] Test with static and animated in-app content

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFragmentHTMLNotification.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFragmentHTMLNotification.java
@@ -72,7 +72,11 @@ public class IterableInAppFragmentHTMLNotification extends DialogFragment implem
     private Handler resizeHandler;
     private Runnable pendingResizeRunnable;
     private float lastContentHeight = -1;
+    private boolean resizeComplete = false;
+    private int resizeAttemptCount = 0;
     private static final int RESIZE_DEBOUNCE_DELAY_MS = 200;
+    private static final int MAX_RESIZE_ATTEMPTS = 5;
+    private static final int STABILIZE_DELAY_MS = 500;
 
     private double backgroundAlpha; //TODO: remove in a future version
     private Rect insetPadding;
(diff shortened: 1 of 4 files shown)


franco-zalamena-iterable:
## PR Analysis

**Problem:** In-app message WebView enters a continuous layout/redraw loop (~10ms cycles) even for static content, caused by `requestLayout()` / `invalidate()` calls triggering new layout passes that in turn trigger more resize evaluations, creating an infinite feedback loop.

**Ideal fix plan:**
- Stop `onProgressChanged(100)` from firing the resize script more than once (it can report 100% multiple times)
- Remove the explicit `requestLayout()` calls after `setLayoutParams()` since `setLayoutParams` already triggers layout internally -- this eliminates a redundant pass that feeds the loop
- Cap the number of resize attempts so the resize-triggers-layout-triggers-resize cycle cannot run forever
- Once content height has stabilized, stop responding to further resize requests

**What the PR did:**
- Added `hasTriggeredResize` flag in `IterableWebChromeClient` so `onProgressChanged(100)` only fires resize once
- Removed explicit `requestLayout()` calls after `setLayoutParams()` (correct -- they were redundant and fed the loop)
- Added `resizeComplete` flag + `MAX_RESIZE_ATTEMPTS` cap in the fragment to stop the resize cycle
- Added `layoutStabilized` flag on `IterableWebView` that overrides `requestLayout()` and `invalidate()` to suppress internal WebView-initiated layout passes
- Resets stabilization on orientation change and before dismiss animations
- Updated existing `IterableWebChromeClientTest` to verify single-fire behavior

**Assessment:**
- Root cause identified: **yes** -- the loop comes from resize triggering layout which triggers resize, compounded by `onProgressChanged(100)` firing multiple times
- Fix correctness: **partially correct** -- the layered approach (single-fire progress, resize cap, layout suppression) is thorough, but overriding `invalidate()` is overly aggressive. Suppressing `requestLayout()` alone would stop the layout loop; suppressing `invalidate()` additionally prevents the WebView from repainting, which could break: (1) CSS-driven visual state changes (hover, focus, active), (2) scrollable content within the in-app message, (3) animated GIFs or CSS animations that the user might add in future campaigns, (4) any JavaScript-driven DOM updates post-load. Consider removing the `invalidate()` override and only suppressing `requestLayout()`.
- Missed:
  - No test coverage for the new `IterableWebView` stabilization logic (`setLayoutStabilized`, `requestLayout` suppression, `invalidate` suppression). These are the most critical behavioral changes and should have unit tests.
  - No test for `markResizeComplete()` or the `MAX_RESIZE_ATTEMPTS` cap logic in the fragment.
  - The `hasTriggeredResize` flag in `IterableWebChromeClient` is never reset. If the WebView is reused for a different in-app message (or reloaded), the resize will never fire again. Consider resetting it when a new page starts loading (e.g., in `onProgressChanged` when progress drops back below 100, or via a reset method).
- Wrong assessment: **none** -- the diagnosis is sound
- Tests: **needed but mostly missing** -- the `IterableWebChromeClientTest` update is good, but the core new logic (layout stabilization, resize capping, stabilization reset on orientation change) has no test coverage

Issue #976, opened 2026-01-19
Enhancement: Improve notification icon fallback to check standard Android notification icon conventions


dimitar-bold:
## ✍️ Issue Description

**Request: Improve default notification icon fallback to support standard Android notification icon conventions**

Currently, when `iterable_notification_icon` is not explicitly configured, the SDK falls back to the app's launcher icon (`context.getApplicationInfo().icon`). However, this default causes notifications to appear as white squares on Android 5.0+ because launcher icons are full-color images, not the silhouette-style icons required for modern Android notifications.

The SDK should improve its default behavior by checking for standard notification icon resources before falling back to the launcher icon. This would provide a much better out-of-the-box experience for developers using common frameworks like Expo, Firebase, and React Native, where notification icons are already configured through standard conventions.

**Proposed enhancement:**

Update the fallback logic in `IterableNotificationHelper.getIconId()` to check these standard locations before using the launcher icon:

1. `com.google.firebase.messaging.default_notification_icon` - Used by Firebase Cloud Messaging
2. `@drawable/notification_icon` - Standard convention used by Expo and many notification libraries  
3. `@drawable/ic_notification` - Common Android notification icon naming convention

This would allow the SDK to automatically utilize notification icons already configured through the Expo plugin and React Native SDK's app configuration, eliminating the white square issue without requiring additional manual configuration.

## 📋 Steps to Reproduce

1. Set up an Expo app with the Iterable React Native SDK and Expo plugin
2. Configure only the standard Expo notification icon in `app.json`:
   ```json
   {
     "expo": {
       "android": {
         "notification": {
           "icon": "./assets/notification-icon.png"
         }
       }
     }
   }
   ```
3. Do NOT manually add `iterable_notification_icon` metadata to AndroidManifest.xml
4. Send a push notification through Iterable
5. Observe notification displays a white square instead of the properly configured notification icon
6. In logs, see: `"No Notification Icon defined - defaulting to app icon"`

**Current behavior:** SDK defaults to launcher icon → white square on Android 5.0+

**Desired behavior:** SDK should detect and use the standard `@drawable/notification_icon` that Expo generates from app configuration

---

#### 📦 Iterable SDK versions:
- `@iterable/react-native-sdk`: 2.2.0
- `@iterable/expo-plugin`: 1.0.2

#### 📲 Android OS version: Android 15

PR #1038, opened 2026-04-07
feat: improve notification icon fallback to check standard Android conventions (#976)

franco-zalamena-iterable:
## Summary
Improves the notification icon fallback chain to check standard icon conventions before using the app launcher icon (which renders as a white square on Android 5.0+).

## Changes

Updated `IterableNotificationHelper.getIconId()` fallback order:
1. `iterable_notification_icon` resource (existing behavior, unchanged)
2. Firebase `com.google.firebase.messaging.default_notification_icon` meta-data
3. `@drawable/notification_icon` — Expo plugin and React Native SDK convention  
4. `@drawable/ic_notification` — common Android convention
5. App launcher icon (last resort, existing behavior)

## Motivation
Expo users who configure notification icons via `app.json` have `@drawable/notification_icon` generated automatically. Without this change, they see a white square unless they also manually add `iterable_notification_icon` to AndroidManifest.xml.

## Test plan
- [ ] Verify notifications use `iterable_notification_icon` when configured
- [ ] Verify notifications use Firebase meta-data icon when configured
- [ ] Verify notifications use `@drawable/notification_icon` when present
- [ ] Verify notifications use `@drawable/ic_notification` when present
- [ ] Verify fallback to launcher icon still works when none of the above are present

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableNotificationHelper.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableNotificationHelper.java
@@ -407,15 +407,22 @@ class IterableNotificationHelper {
         }
 
         /**
-         * Returns the iconId from potential resource locations
+         * Returns the iconId from potential resource locations.
+         *
+         * Fallback order:
+         * 1. {@code iterable_notification_icon} meta-data or icon set via {@link IterableApi#setNotificationIcon}
+         * 2. Firebase {@code com.google.firebase.messaging.default_notification_icon} meta-data
+         * 3. {@code @drawable/notification_icon} (Expo / React Native convention)
+         * 4. {@code @drawable/ic_notification} (common Android convention)
+         * 5. App launcher icon (last resort — renders as a white square on Android 5.0+)
          *
          * @param context
-         * @return
+         * @return resource id of the notification icon to use
          */
         private int getIconId(Context context) {
             int iconId = 0;
 
-            //Get the iconId set in the AndroidManifest.xml
+            // 1a. Check iterable_notification_icon set in AndroidManifest.xml meta-data
             if (iconId == 0) {
                 try {
                     ApplicationInfo info = context.getPackageManager().getApplicationInfo(context.getPackageName(), PackageManager.GET_META_DATA);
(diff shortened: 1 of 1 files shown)


franco-zalamena-iterable:
## PR Analysis

**Problem:** When `iterable_notification_icon` is not configured, the SDK falls back directly to the app launcher icon, which renders as a white square on Android 5.0+ because launcher icons are full-color rather than the required silhouette style. Expo/React Native users who already have proper notification icons at standard drawable resource names get no benefit from them.

**Ideal fix plan:**
- Extend the fallback chain in `getIconId()` to probe well-known drawable resource names (`notification_icon`, `ic_notification`) and the Firebase meta-data key before falling back to the launcher icon
- Add debug logging for each fallback step so developers can trace which icon was selected
- Avoid duplicating the `PackageManager.getApplicationInfo()` call that already exists a few lines above -- ideally read meta-data once and check both the Iterable key and the Firebase key from the same `Bundle`
- Add unit tests covering each fallback level (at minimum: Firebase meta-data present, `notification_icon` drawable present, `ic_notification` drawable present, none present -> launcher icon)

**What the PR did:**
- Added three new fallback steps in exactly the order proposed in the issue (Firebase meta-data, `@drawable/notification_icon`, `@drawable/ic_notification`)
- Added debug log lines for each new fallback
- Updated Javadoc on `getIconId()` to document the full fallback chain
- Did not add any tests
- Did not consolidate the duplicate `ApplicationInfo` lookup

**Assessment:**
- Root cause identified: **yes** -- the direct fall-through to the launcher icon with no intermediate checks is the cause
- Fix correctness: **correct** -- the new fallback order is sensible, resource lookups use the existing patterns already in the codebase, and the original behavior is preserved as the final fallback
- Missed:
  - **Unit tests.** The repo already has `IterableNotificationTest.java` which exercises icon resolution (it sets `context.getApplicationInfo().icon`). New tests covering each fallback step should be straightforward and would prevent regressions if the method is refactored later.
  - **Minor duplication.** The `PackageManager.getApplicationInfo(..., GET_META_DATA)` call is now made twice in `getIconId()` (once for the Iterable meta-data key, once for the Firebase key). These could be consolidated into a single lookup, reading both keys from the same `info.metaData` bundle. This is a nit, not a blocker.
- Wrong assessment: **none** -- the approach matches the issue request and is a reasonable enhancement
- Tests: **needed but missing** -- the existing test file demonstrates that icon resolution is testable in this codebase; coverage for the new fallback steps should be added

PR #1048, opened 2026-04-28
Fixing BCIT and in app tests in CI

franco-zalamena-iterable:
## 🔹 Jira Ticket(s) if any

* [SDK-180](https://iterable.atlassian.net/browse/SDK-180)

## ✏️ Description

This PR is meant to fix constant failures on the E2E-In-app testing and Instrumental tests.

The fix is centered around updating the engine the tests are running on and some minor lookup fixes in some tests.

We also have better diagnostics now for the failures on emulator

[SDK-180]: https://iterable.atlassian.net/browse/SDK-180?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sumeruchat (changes requested):
Thanks Franco — solid CI cleanup, the `setup-java@v4.8.0` (Temurin, SHA-pinned), `wrapper-validation` upgrade, and ubuntu+KVM migration all check out, and the `UiAutomator` switch to `Until.findObject(By.res(...), 30s)` is a real robustness improvement (not papering over a product bug). The `|| true`s correctly stay on diagnostics/cleanup — Gradle exit codes still propagate.

Requesting changes on **one blocker**; the rest can land as follow-ups if you'd rather not pile onto this PR.

---

### 🚫 Blocker

**1. Secret leak via new E2E diagnostics artifacts**

`run-e2e.sh:54-79` captures full logcat + `uiautomator dump` + screencap, and `.github/workflows/inapp-e2e-tests.yml:115-124` uploads them on every run. The screen the test navigates to logs key fragments in `integration-tests/src/main/java/com/iterable/integration/tests/activities/InAppMessageTestActivity.kt:209-225` as `first8...last4` (or the full value if ≤12 chars). GitHub Actions' secret masking matches the literal secret only — substrings won't be redacted — so the uploaded artifacts can leak stable BCIT/server-key material to anyone who can read the workflow run.

Please redact `InAppMessageTestActivity` the same way `MainActivity` does (length-only, or at most last4) and drop the composed `userInfo` log.

---

### Non-blocking (recommend addressing, but won't hold up the PR)

**2. APK secret exposure in `integration-tests.yml`**

`integration-tests.yml:278-286` and `:437-445` still upload `integration-tests/build/outputs/`. Those APKs embed `ITERABLE_API_KEY` and `ITERABLE_SERVER_API_KEY` as `BuildConfig` constants (see `integration-tests/build.gradle:35-40`). Pre-existing, but since this PR touches that workflow and documents the APK risk in the E2E workflow, it'd be nice to close it in the same sweep — or at least add a `TODO` / open a follow-up.

**3. 6h-hang mitigation has gaps**

The step-level timeouts help, but there are still unbounded `adb` calls upstream that can wedge the job:
- `integration-tests.yml:169-175` — `adb logcat -d | tail -20` in the wait-failure path
- `integration-tests.yml:226-230` — bare `adb wait-for-device` after `adb kill-server`
- `integration-tests.yml:411-415` — bare `adb wait-for-device` in the nightly job

If the emulator never comes up or adb gets stuck offline, these can still consume the full job timeout. Either wrap them in `timeout ... adb ...` or — better — migrate this workflow to `ReactiveCircus/android-emulator-runner` like the in-app E2E job.

**4. Diagnostic trap can itself hang**

`run-e2e.sh:71-79` runs `uiautomator dump`, `adb pull`, and `screencap` in the `EXIT` trap without timeouts. If Gradle fails because the emulator is unresponsive, the diagnostic capture can hang and replace the useful Gradle failure with a 6h job timeout. Short per-command `timeout`s would preserve diagnostics without re-introducing the bug we just fixed.

---

Static review only — I didn't run the workflows.

sumeruchat (approved):
Thanks for the thorough sweep, Franco — `71a251ba` cleanly addresses the original blocker plus all three non-blocking items from the previous round. Approving.

### ✅ Verified fixed
- **Secret leak in CI artifacts**: `maskedKey()` returns `****<last4> (length=N)` / `(length=N)` / `(empty)` — no `first8` prefix exposure, and the composed `userInfo` is no longer `Log.d`'d. Confirmed no raw `BuildConfig.ITERABLE_API_KEY` / `ITERABLE_SERVER_API_KEY` log or render paths remain under `integration-tests/src`.
- **APK upload**: `integration-tests/build/outputs/` removed from both main + nightly upload-artifact paths, with clear in-line comments. Confirmed no other workflow uploads that directory and the kept `reports/` / `test-logs.txt` paths don't transitively contain the APK.
- **Unbounded adb on the previously-discussed paths**: `timeout 10` on the wait-failure `logcat -d`, `timeout 60` on the main job's post-restart `wait-for-device`, `timeout 300` on the nightly `wait-for-device`. Sensible values for their contexts.
- **EXIT trap in `run-e2e.sh`**: every adb call inside `capture_post_test` (uiautomator dump, both pulls, both `rm`s, screencap) is wrapped with `timeout $ADB_TIMEOUT` (default 10s, configurable). `LOGCAT_PID` killed cleanly on EXIT.
- **Comment churn**: the SDK-170 explanatory comment removals are comments-only — `setAutoDisplayPaused(true)`, the pre-existing message drain, the `uiDevice.wait(Until.findObject(...))` pattern, and the length-only `BuildConfig` logs are all still present.

`./gradlew :integration-tests:assembleDebug :integration-tests:assembleDebugAndroidTest` passes after a minimal `google-services.json` stub (expected outside CI).

---

### Non-blocking follow-ups (suggest separate tickets so they don't get lost)

**1. (P2) Test user email is still raw in logcat / on-screen / UI hierarchy artifacts.**
If `ITERABLE_TEST_USER_EMAIL` is considered sensitive (it's a test account, but lives on a public repo and is captured into hierarchy.xml/screenshot.png/logcat artifacts on every E2E run), there are still several leak paths the same artifacts will pick up:
- `integration-tests/src/androidTest/java/com/iterable/integration/tests/InAppMessageIntegrationTest.kt:127, :136` — raw `Log.d` of `TestConstants.TEST_USER_EMAIL`
- `integration-tests/src/androidTest/java/com/iterable/integration/tests/BaseIntegrationTest.kt:113-114` — logged during setup
- `integration-tests/src/main/java/com/iterable/integration/tests/utils/IntegrationTestUtils.kt:93, :95, :146, :179` — `email` / `recipientEmail` logged raw; the MVP in-app test calls this path
- `integration-tests/src/main/java/com/iterable/integration/tests/activities/InAppMessageTestActivity.kt:98` — `"for user $userEmail"` to logcat; `:208` renders `"User: $userEmail"` into the on-screen `userInfoTextView`, which means hierarchy.xml + screenshot.png capture it on every run.
- Same pattern exists in `EmbeddedMessageIntegrationTest`, `DeepLinkIntegrationTest`, and `CampaignTriggerTestActivity`.

Cheapest fix is a shared `maskEmail()` helper alongside `maskedKey()` (keep only length + domain, or `f***@…`) and use it everywhere these get logged/rendered. If the test email is explicitly considered non-sensitive, fine to ignore — would be worth a comment in `TestConstants` saying so, so this doesn't get re-flagged next sweep.

**2. (P2) Residual unbounded adb calls that can still wedge the 6h job.**
The previously-discussed paths are closed, but several adb calls remain in steps with no step-level timeout. If one of these hangs the job can still consume the full 6h:
- `.github/workflows/integration-tests.yml:139, :156, :169` — boot-loop command substitutions
- `:186-:226` — setup / grant / health calls
- `:252-:253` — pre-test status checks

Either GNU `timeout`-wrap them, or add `timeout-minutes:` on the adb-heavy steps so the failure is fast.

**3. (P3) `run-e2e.sh` has a few adb calls outside the EXIT trap that aren't wrapped.**
- `:51` permission grant
- `:55` logcat clear
- `:56` background `adb logcat &` start (this one is inherently long-running, but the `wait` after `kill $LOGCAT_PID` on EXIT isn't itself bounded — add a TERM → KILL fallback if `wait` doesn't return promptly)

Minor next to (2); fold into the same follow-up.

---

Static review + targeted local build. Did not run the workflows themselves.

PR #1053, opened 2026-05-11
SDK-499 Fix NPE race condition in EmbeddedSessionManager.updateDisplayCountAndDuration()

### LINKED ISSUE #1052: NullPointerException in EmbeddedSessionManager.updateDisplayCountAndDuration() — thread-safety race condition
## SDK Version
`3.6.5`

## Platform
Android

---

## Description

We are seeing a production crash (`NullPointerException`) inside `EmbeddedSessionManager.updateDisplayCountAndDuration()` at line 114. The crash occurs when `endSession()` or `pauseImpression()` is called from a background thread (e.g. a `Dispatchers.Default` coroutine worker), and another thread concurrently modifies the impression state.

This is a **TOCTOU (time-of-check/time-of-use) race condition** in `EmbeddedImpressionData`. The field `start: Date?` is a plain Kotlin `var` with no `@Volatile` annotation and no synchronisation. The sequence that causes the crash is:

1. Thread A calls `updateDisplayCountAndDuration()` and passes the `if (start != null)` null check.
2. Thread B sets `start = null` (e.g. via a concurrent `endAllImpressions()` reset).
3. Thread A dereferences `start!!` → **NPE**.

---

## Stack Traces

**Crash 1** — via `endSession()`:
```
Fatal Exception: java.lang.NullPointerException
  at com.iterable.iterableapi.EmbeddedSessionManager.updateDisplayCountAndDuration(EmbeddedSessionManager.kt:114)
  at com.iterable.iterableapi.EmbeddedSessionManager.endAllImpressions(EmbeddedSessionManager.kt:91)
  at com.iterable.iterableapi.EmbeddedSessionManager.endSession(EmbeddedSessionManager.kt:41)
  at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:100)
  at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:586)
  at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.kt:829)
  at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:704)
```

**Crash 2** — via `pauseImpression()`:
```
Fatal Exception: java.lang.NullPointerException
  at com.iterable.iterableapi.EmbeddedSessionManager.updateDisplayCountAndDuration(EmbeddedSessionManager.kt:114)
  at com.iterable.iterableapi.EmbeddedSessionManager.pauseImpression(EmbeddedSessionManager.kt:86)
  at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:100)
  at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:586)
  at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.kt:829)
  at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:704)
```

Both crashes occur on `CoroutineScheduler$Worker` — i.e. `Dispatchers.Default` (thread pool), not the main thread.

---

## Root Cause

In `EmbeddedImpressionData`, the `start` field is declared as a plain mutable property:

```kotlin
var start: Date? = null
```

In `EmbeddedSessionManager.updateDisplayCountAndDuration()` (line ~110–115):

```kotlin
if (start != null) {
    // ... Thread B sets start = null here ...
    val duration = Date().time - start!!.time  // ← NPE
}
```

Because `start` is neither `@Volatile` nor guarded by a lock, the JVM is free to cache the null-check result in a register. A concurrent write from another thread makes the dereference unsafe.

---

## Suggested Fix

The simplest fix is to capture `start` in a local variable before the null-check, which is the standard safe-access pattern in concurrent Kotlin/Java:

```kotlin
val startSnapshot = start
if (startSnapshot != null) {
    val duration = Date().time - startSnapshot.time  // safe — local val is thread-confined
}
```

Alternatively, marking the field as `@Volatile` would ensure visibility, though a local snapshot is still needed to prevent the race window.

A broader fix would be to add `@MainThread` annotations to `EmbeddedSessionManager`'s public API to document that it is not thread-safe and must be called on the main thread.

---

## Workaround (client-side)

We are currently working around this by dispatching all `EmbeddedSessionManager` calls onto `Dispatchers.Main`, which serialises access to a single thread. However, this is a client-side workaround for a thread-safety issue in the SDK itself.

Shamyyoun:
## Jira Ticket

[SDK-499](https://iterable.atlassian.net/browse/SDK-499)

## Summary

Fixes a `NullPointerException` crash in `updateDisplayCountAndDuration()` caused by a TOCTOU (time-of-check/time-of-use) race condition.

Closes https://github.com/Iterable/iterable-android-sdk/issues/1052

---

## Problem

`EmbeddedImpressionData.start` is a plain `var Date?` with no synchronisation. The previous code checked `if (impressionData.start != null)` but then accessed `impressionData.start!!` in the body. When called concurrently from multiple threads (e.g. via `Dispatchers.Default` coroutines), another thread can set `start = null` between the null-check and the dereference, causing a crash:

```
Fatal Exception: java.lang.NullPointerException
  at com.iterable.iterableapi.EmbeddedSessionManager.updateDisplayCountAndDuration(EmbeddedSessionManager.kt:114)
  at com.iterable.iterableapi.EmbeddedSessionManager.endAllImpressions(EmbeddedSessionManager.kt:91)
  at com.iterable.iterableapi.EmbeddedSessionManager.endSession(EmbeddedSessionManager.kt:41)
  at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:704)
```

## Fix

Capture `impressionData.start` into a local `val` before the null-check. Because local variables are thread-confined (stack-allocated), the value cannot be changed by another thread after the null-check passes:

```kotlin
// Before
if (impressionData.start != null) {
    impressionData.duration = ... - impressionData.start!!.time  // ← NPE under concurrent access
}

// After
val start = impressionData.start
if (start != null) {
    impressionData.duration = ... - start.time  // ← safe, local val is immutable
}
```

This is the idiomatic Kotlin/Java pattern for safe null-check-then-use on shared mutable fields.

[SDK-499]: https://iterable.atlassian.net/browse/SDK-499?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/EmbeddedImpressionData.kt
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/EmbeddedImpressionData.kt
@@ -9,7 +9,7 @@ data class EmbeddedImpressionData(
     val placementId: Long,
     var displayCount: Int = 0,
     var duration: Float = 0.0f,
-    var start: Date? = null
+    @Volatile var start: Date? = null
 ) {
     constructor(
         messageId: String,
--- a/iterableapi/src/main/java/com/iterable/iterableapi/EmbeddedSessionManager.kt
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/EmbeddedSessionManager.kt
@@ -108,10 +108,11 @@ public class EmbeddedSessionManager {
     }
 
     private fun updateDisplayCountAndDuration(impressionData: EmbeddedImpressionData): EmbeddedImpressionData {
-        if (impressionData.start != null) {
+        val start = impressionData.start
+        if (start != null) {
             impressionData.displayCount = impressionData.displayCount.plus(1)
             impressionData.duration =
-                impressionData.duration.plus((Date().time - impressionData.start!!.time) / 1000.0)
+                impressionData.duration.plus((Date().time - start.time) / 1000.0)
                     .toFloat()
             impressionData.start = null
         }
--- a/iterableapi/src/main/java/com/iterable/iterableapi/EmbeddedSessionManager.kt
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/EmbeddedSessionManager.kt
@@ -108,13 +108,15 @@ public class EmbeddedSessionManager {
     }
 
     private fun updateDisplayCountAndDuration(impressionData: EmbeddedImpressionData): EmbeddedImpressionData {
-        val start = impressionData.start
-        if (start != null) {
-            impressionData.displayCount = impressionData.displayCount.plus(1)
-            impressionData.duration =
-                impressionData.duration.plus((Date().time - start.time) / 1000.0)
-                    .toFloat()
-            impressionData.start = null
+        synchronized(impressionData) {
+            val start = impressionData.start
+            if (start != null) {
+                impressionData.displayCount = impressionData.displayCount.plus(1)
+                impressionData.duration =
+                    impressionData.duration.plus((Date().time - start.time) / 1000.0)
+                        .toFloat()
+                impressionData.start = null
+            }
         }
         return impressionData
     }


franco-zalamena-iterable:
Closing this Pr due to Pr #1083

PR #1055, opened 2026-05-21
Drop racy parallel pre-build step from inapp-e2e-tests workflow

franco-zalamena-iterable:
## ✏️ Description

The merge for fixing ci tests missed a part regarding the setup, the pr had cache and this was not picked up for e2e messages tests


sumeruchat (approved):
Approving :)

Great job on the comment explaining why the pre-build only started racing after the ubuntu-latest + KVM migration. That's exactly the context that keeps someone from re-adding it later.

Confirmed the action's own `connectedDebugAndroidTest` builds both APKs transitively, so the removed step was redundant.

joaodordio (approved):

Issue #1056, opened 2026-05-25
In-app HTML messages ignore system light/dark theme on Android


AndrazP:
## ✍️ Issue Description

On Android 13+ (API 33+), in-app HTML messages always render the `@media (prefers-color-scheme: dark)` branch regardless of the actual system theme. A campaign whose HTML defines distinct light and dark styles always shows the dark variant, even when the host app and device are in light mode.

The same campaign HTML renders correctly on the **iOS Iterable Swift SDK** — the variant switches with the device theme as expected — so this is Android-specific.

The host app in our case is built with **Jetpack Compose** (Material 3, DayNight-aware theme at the application level). The activity theme correctly tracks the system theme; only the in-app message dialog does not, because the SDK overrides the theme on its `DialogFragment` and constructs the WebView under the host activity context.

We also verified the bug is still present on the `SDK-100-compose-support` branch (PR https://github.com/Iterable/iterable-android-sdk/pull/1015) — the Compose support work does not address it.

Root cause is in `IterableInAppFragmentHTMLNotification`:

1. **Line 122** — the `DialogFragment` style is hard-coded to a dark (non-DayNight) theme:
   ```java
   this.setStyle(DialogFragment.STYLE_NO_FRAME,
       androidx.appcompat.R.style.Theme_AppCompat_NoActionBar);
   ```
   `Theme.AppCompat.NoActionBar` extends `Theme.AppCompat`, which is the dark variant of AppCompat (the light one is `Theme.AppCompat.Light`, the system-following one is `Theme.AppCompat.DayNight`). Its `android:isLightTheme` is always `false`.

2. **Line 204** — the `WebView` is constructed with the host activity's context, not the dialog's themed context:
   ```java
   webView = createWebViewSafely(getContext());
   ```
   `Fragment.getContext()` returns the host activity context, so the DayNight fix on the dialog alone is not visible to the WebView.

On API 33+ (WebView 105+), `WebView` derives the CSS `prefers-color-scheme` value from the host theme's `isLightTheme` attribute (per the AndroidX WebView documentation). With (1) and (2) combined, the WebView always sees `isLightTheme=false` and reports `prefers-color-scheme: dark` to the loaded HTML, regardless of the system theme.

[...]

PR #1057, opened 2026-05-25
SDK-500 In-app HTML messages ignore system light/dark theme on Android

AndrazP:
In-app HTML messages always rendered the prefers-color-scheme: dark variant regardless of the system theme on API 33+. Two fixes:

- Switch DialogFragment style from Theme.AppCompat.NoActionBar (a dark theme: extends Theme.AppCompat, not Theme.AppCompat.Light) to Theme.AppCompat.DayNight.NoActionBar, so isLightTheme tracks the system mode.

- Build the WebView with the dialog's themed context instead of Fragment.getContext(), which returns the host activity context. On API 33+ WebView derives prefers-color-scheme from the hosting theme's isLightTheme attribute, so the DayNight fix only takes effect when the WebView is constructed under the dialog context.

## 🔹 Jira Ticket(s) if any

[SDK-500](https://iterable.atlassian.net/browse/SDK-500)

## ✏️ Description

See https://github.com/Iterable/iterable-android-sdk/issues/1056

[SDK-500]: https://iterable.atlassian.net/browse/SDK-500?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFragmentHTMLNotification.java
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFragmentHTMLNotification.java
@@ -120,7 +120,7 @@ public class IterableInAppFragmentHTMLNotification extends DialogFragment implem
         this.backgroundAlpha = 0;
         this.messageId = "";
         insetPadding = new Rect();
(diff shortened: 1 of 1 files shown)


franco-zalamena-iterable:
https://github.com/Iterable/iterable-android-sdk/pull/1084 open to address this issue

PR #1061, opened 2026-06-01
[SDK-115] Enable full BCIT on CI

franco-zalamena-iterable:
## 🔹 Jira Ticket(s) if any

* [SDK-115](https://iterable.atlassian.net/browse/SDK-115)

## ✏️ Description

We had two different test sets before for bcit, some functionality was not working, now with this pr we are adding all the bcit test suite.

The new test is called Integration Tests (BCIT), on this PR it runs what was already running on E2E in app tests.
We are then creating different PRs for each test suite and we can see it working independently

[SDK-115]: https://iterable.atlassian.net/browse/SDK-115?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sumeruchat (changes requested):
Nice stack split and good dated-user pattern. One non-blocking follow-up: the stable test email still reaches public CI diagnostics (MainActivity card, `DeepLinkIntegrationTest.kt:114`), worth masking to keep the dated-user hardening intact.

sumeruchat (approved):
Approving.

franco-zalamena-iterable:
Addressed the email-masking feedback in 37db7243.

Added a `maskEmail()` helper (`f***@iterable.com` — keeps first char + domain for debuggability) and applied it at every site that writes an email into logcat or the screen:

- `MainActivity` override card label + init log
- `DeepLinkIntegrationTest` / `InAppMessageIntegrationTest` / `EmbeddedMessageIntegrationTest` / `BaseIntegrationTest` sign-in logs
- `IntegrationTestUtils.ensureUserSignedIn` logs
- `PushNotificationTestActivity` / `InAppMessageTestActivity` logs

`CampaignTriggerTestActivity` is intentionally left unmasked — it's a manual-only tool with no instrumented test driving it, so it never reaches CI artifacts, and its on-screen log needs the real email to be usable.

4 unit tests cover the masking. Full module compiles, `:integration-tests:testDebugUnitTest` green.

PR #1065, opened 2026-06-09
[SDK-107] appAlreadyRunning flag parity

franco-zalamena-iterable:
## 🔹 Jira Ticket(s) if any

* [SDK-107](https://iterable.atlassian.net/browse/SDK-107)

## ✏️ Description

iOS is currently sending a flag "appAlreadyRunning" when tracking a push opening. the flag is always sent with the value **false** unless the customer app is sending a different value, but in the event of they not sending us that we are sending "false" to the backend.

Android now matches the behavior and we can follow up with a ticket to check if this is intentional

[SDK-107]: https://iterable.atlassian.net/browse/SDK-107?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


joaodordio (approved):
We should add some tests to this new surface but other than that LGTM

PR #1066, opened 2026-06-12
[SDK-419] handle oversized in-app HTML payloads

### LINKED ISSUE #638: `TransactionTooLargeException` when persisting in-app messages
We're seeing dramatic spikes in crashes caused by `TransactionTooLargeException`s when attempting to save activity state while this SDK is displaying an in-app message.

```
Fatal Exception: java.lang.RuntimeException: android.os.TransactionTooLargeException: data parcel size 579324 bytes
       at android.app.OurActivity$StopInfo.run(PendingTransactionActions.java:160)
       at android.os.Handler.handleCallback(Handler.java:873)
       at android.os.Handler.dispatchMessage(Handler.java:99)
       at android.os.Looper.loop(Looper.java:215)
       at android.app.ActivityThread.main(ActivityThread.java:6939)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:493)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:870)
```

It took a long time to identify the Iterable SDK as the cause because the exception is reported from the Android framework with no useful attribution hints. However we're now confident that the root cause was that we launched some Iterable campaigns which sent in-app messages with abnormally large HTML (>250kb), since turning them off has resulted in a dramatic reduction in crash reports.

We've now spent time going through our Iterable campaigns to ensure that all in-app HTML is a reasonable size (<100kb), but I wanted to suggest some further safeguards you could follow-up with:
1. Stop persisting the HTML upon activity destruction (it's not clear to me where this is happening, though `IterableWebView` through `IterableInAppFragmentHTMLNotification` seems a likely culprit)
2. Including a tool in the Iterable web console that warns against in-app messages with HTML above a reasonable size (say 100kb)

franco-zalamena-iterable:
## 🔹 Jira Ticket(s) if any

* [SDK-419](https://iterable.atlassian.net/browse/SDK-419)

## ✏️ Description

Intead of passing the html via bundle on recreation we are saving it in memory and refetching on recreation to prevent crashes 

| Before | After |
|--------|------|
|<video src="https://github.com/user-attachments/assets/4ff40ea3-1d39-46e0-adf1-4b593fb2d223" width="320" height="240" controls></video>|<video src="https://github.com/user-attachments/assets/0af8770e-f2ac-4295-a6f0-c1ac625fe527" width="320" height="240" controls></video>|

[SDK-419]: https://iterable.atlassian.net/browse/SDK-419?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


joaodordio (approved):
LGTM

Just checking that this will work on both Fragment and Compose views.

PR #1070, opened 2026-06-23
[SDK-487] is ok toshow parity

franco-zalamena-iterable:
## 🔹 Jira Ticket(s) if any

* [SDK-487](https://iterable.atlassian.net/browse/SDK-487)

## ✏️ Description

Android did not have a proper way to defer an in-app showing in a way that was not setting a flag to a specific value, the problem with that is the possibility of a racing condition. So this adds a way of having a check if it is deffered in the same place as we check for SHOW or SKIP.

The enum change is a public api change but the API never returns this value, it's intake only, so it should not be a big concern for integrators

| Before | After |
|--------|------|
|<video src="https://github.com/user-attachments/assets/02223964-37a6-4d06-b138-d5386f744ebd" width="320" height="240" controls></video>|<video src="https://github.com/user-attachments/assets/c9df6219-2342-4dd6-824d-cd5a198cb042" width="320" height="240" controls></video>|

[SDK-487]: https://iterable.atlassian.net/browse/SDK-487?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sumeruchat (approved):
Approving. Nice that DEFER hooks in at the same SHOW/SKIP decision point rather than a separate flag, which is what avoids the race the ticket was about, and keeping delivery tracking in onMessageAdded means a message re-evaluated on each pass doesn't double-track.

Two non-blocking notes:
1. resumeInAppDisplay() doesn't change the paused flag, so an app that used setAutoDisplayPaused(true) still has to call setAutoDisplayPaused(false) to resume. Worth spelling that out in the javadoc/changelog so integrators don't call resumeInAppDisplay() expecting it to unpause.
2. The new tests cover the core DEFER/resume behavior well. Two worth adding: trackInAppDelivery stays at one across repeated DEFER passes, and resumeInAppDisplay() while still paused is a no-op.

joaodordio (approved):
LGTM!

PR #1072, opened 2026-06-25
[SDK-497] Add unknown user criteria fetch callbacks

franco-zalamena-iterable:
## 🔹 Jira Ticket(s) if any

* [SDK-497](https://iterable.atlassian.net/browse/SDK-497)

## ✏️ Description

Adding two new callbacks for IterableUnknownUserHandler for success and failure on criteria fetch. These two callbacks have default implementation so they are optional of implementation

The Success callback fires on every fetch success not only once, implementation has to keep that in mind

This development was started by contributor @yinzcam-nick

| Before | After |
|--------|------|
|<video src="https://github.com/user-attachments/assets/5ea7d3b7-ecf0-421d-9ac5-27cdfc913911" width="320" height="240" controls></video>|<video src="https://github.com/user-attachments/assets/01046538-563c-4778-9e40-a2180c5deeec" width="320" height="240" controls></video>|

[SDK-497]: https://iterable.atlassian.net/browse/SDK-497?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


iterableapi/src/main/java/com/iterable/iterableapi/IterableUnknownUserHandler.java
```diff
@@ -1,5 +1,27 @@
...
     void onUnknownUserCreated(String userId);
-}
\ No newline at end of file
+
+    /**
+     * Called when the SDK fetches unknown user activation criteria successfully.
+     * <p>
+     * Fires on every criteria fetch, implementation should take that into consideration.
+     * @param criteria the criteria returned by the server
+     */
```
joaodordio, line 23: Tiny doc nit, but worth it for the integrator experience: the Javadoc says "fetches ... successfully" but not "after persisting locally". Since the whole point of this hook is to call track() inside it and have the new event be evaluated against the just-arrived criteria, can we add a line like: "Fires after the criteria response has been persisted to local storage; it is safe to call track, trackPurchase, etc. from this callback and have them evaluated against the new criteria set."

iterableapi/src/test/java/com/iterable/iterableapi/IterableApiCriteriaFetchTests.java
```diff
@@ -191,4 +194,76 @@ public void testForegroundCriteriaFetchWithCooldown() throws Exception {
         IterableActivityMonitor.getInstance().unregisterLifecycleCallbacks(getContext());
         IterableActivityMonitor.instance = new IterableActivityMonitor();
     }
+
+    @Test
+    public void testCriteriaCallbackOnSuccess() throws Exception {
```
joaodordio, line 201: Two small test additions worth doing in this PR:

1. In testCriteriaCallbackOnFailure, assert the reason passed to onCriteriaFetchFailed is non-empty so changes in IterableRequestTask.handleErrorResponse don't silently break the contract.
2. Add a test that triggers the criteria fetch twice (e.g. init + foreground simulation like testForegroundCriteriaFetchWhenConditionsMet) and asserts the callback fires twice. The CHANGELOG explicitly tells implementers the callback "may fire on every fetch", but right now no test pins that behavior.

joaodordio (approved):
We should also open a Docs PR highlighting the changes here

iterableapi/src/main/java/com/iterable/iterableapi/IterableUnknownUserHandler.java
franco-zalamena-iterable: Thanks for this point, i added it in the latest commit

iterableapi/src/test/java/com/iterable/iterableapi/IterableApiCriteriaFetchTests.java
franco-zalamena-iterable: New test created to deal with the double calls and also the non-empty reason

franco-zalamena-iterable:
The docs should be on the release right?

PR #1073, opened 2026-07-10
[SDK-417] Fail gracefully when managers accessed before init

### LINKED ISSUE #727: RunTime Exceptions are NOT documented
if I try and use the Iterable manager before its registered, I get a UNDOCUMENTED RuntimeException

java.lang.RuntimeException: IterableApi must be initialized before calling getInAppManager(). Make sure you call IterableApi#initialize() in Application#onCreate
                        	at com.iterable.iterableapi.IterableApi.getInAppManager(IterableApi.java:662)

WHY would do this ?
WHY throw Undocumented Runtime Exceptions like this ? 
just log a message and fail gracefully. 

Please fix this,

franco-zalamena-iterable:
## 🔹 Jira Ticket(s) if any

* [SDK-417](https://iterable.atlassian.net/browse/SDK-417)

## ✏️ Description

Fix to prevent calls before initialization to crash the app. We now log an error and have a new way to get the Iterable instance or null on the spot.

[SDK-417]: https://iterable.atlassian.net/browse/SDK-417?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/EmptyEmbeddedManager.kt
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/EmptyEmbeddedManager.kt
@@ -10,7 +10,7 @@ internal class EmptyEmbeddedManager : IterableEmbeddedManager() {
     private val emptySessionManager = EmbeddedSessionManager()
 
     private fun logNotInitialized(method: String) {
-        IterableLogger.e(TAG, "$method called before IterableApi was initialized; no-op. " +
+        IterableLogger.e(LOG_TAG, "$method called before IterableApi was initialized; no-op. " +
                 "Call IterableApi.initialize() in Application#onCreate.")
     }
 
@@ -58,5 +58,9 @@ internal class EmptyEmbeddedManager : IterableEmbeddedManager() {
     }
 
     override fun onSwitchToBackground() {
+    }
+
+    private companion object {
+        private const val LOG_TAG = "EmptyEmbeddedManager"
     }
 }
--- a/iterableapi/src/main/java/com/iterable/iterableapi/EmptyEmbeddedManager.kt
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/EmptyEmbeddedManager.kt
@@ -1,0 +1,62 @@
+package com.iterable.iterableapi
+
+/**
+ * No-op [IterableEmbeddedManager] returned by [IterableApi.getEmbeddedManager] when the SDK has not
+ * been initialized. Every method is a benign no-op so callers never crash for a pre-initialization
+ * usage error. Use [IterableApi.getEmbeddedManagerOrNull] to detect this state explicitly.
+ */
+internal class EmptyEmbeddedManager : IterableEmbeddedManager() {
+
+    private val emptySessionManager = EmbeddedSessionManager()
+
+    private fun logNotInitialized(method: String) {
+        IterableLogger.e(TAG, "$method called before IterableApi was initialized; no-op. " +
+                "Call IterableApi.initialize() in Application#onCreate.")
+    }
+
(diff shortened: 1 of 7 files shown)


iterableapi/src/main/java/com/iterable/iterableapi/EmptyEmbeddedManager.kt
```diff
@@ -0,0 +1,62 @@
+package com.iterable.iterableapi
+
+/**
+ * No-op [IterableEmbeddedManager] returned by [IterableApi.getEmbeddedManager] when the SDK has not
+ * been initialized. Every method is a benign no-op so callers never crash for a pre-initialization
+ * usage error. Use [IterableApi.getEmbeddedManagerOrNull] to detect this state explicitly.
+ */
+internal class EmptyEmbeddedManager : IterableEmbeddedManager() {
+
```
rtlsilva, line 9: `EmptyInAppManager` defines its own `private static final String TAG = "EmptyInAppManager";`, while this `EmptyEmbeddedManager` does not.
This means that a pre-init error on the embedded side will get tagged with the base `IterableEmbeddedManager` tag, while one in the in-app side will get tagged with the easier to find `EmptyInAppManager`.
Suggest adding an `EmptyEmbeddedManager` tag constant for consistency.

```suggestion
private const val TAG = "EmptyEmbeddedManager"
```

joaodordio (approved):
Nice fix, this catches up nicely with the Swift SDK's `EmptyInAppManager` / `EmptyEmbeddedManager` pattern and also quietly resolves the process-death recovery crash in `IterableInAppFragmentHTMLNotification` where the fragment could re-hydrate before `Application#onCreate` re-initializes the SDK.

Approving, with a few small non blockers worth considering before merge:

- **Log noise:** the empty stubs log an error on every method call, plus the getter also logs. In a render loop before init this gets loud fast. Consider logging once at the getter and either dropping the per-method logs or dropping them to `IterableLogger.d`.
- **Debug assertion parity with iOS:** Swift pairs the empty return with `assertionFailure`, which crashes debug builds and no-ops in release. A `BuildConfig.DEBUG` guarded throw in `getInAppManager()` / `getEmbeddedManager()` would give devs the same fail-fast signal locally while keeping prod crash-safe. This is standard operation in iOS world, not sure about Android.
- **Post-init coverage:** worth adding two tests that call `IterableApi.initialize(...)` and assert both getters return the real managers (not `instanceof Empty*`) and `*OrNull` return non-null. Locks in the wiring flip.
- **Failure-handler contract test:** `EmptyInAppManager.setRead(msg, read, success, failure)` and `removeMessage(..., success, failure)` invoke `failureHandler.onFailure("Iterable SDK is not initialized", null)`. That's an observable contract, cheap to lock down with a test.
- **Parent-class safety note:** a short comment on `IterableInAppManager` and `IterableEmbeddedManager` saying "any new method must be overridden in the Empty* subclass or must not touch `api` / `context` / `iterableApi`" would help future contributors avoid landmines, especially with the new `lateinit` fields.
- **`getMessages(placementId)` parity:** iOS returns `[]`, Android's empty stub returns `null` (mirrors the pre-existing nullable signature). Not from this PR, but returning `emptyList()` here would soften it without touching the public type.
- **Migration guide nit:** worth a one-liner that if anyone was catching `RuntimeException` from these getters to detect the uninitialized state, they should switch to the new `*OrNull` variants.

None of these are blockers, ship it whenever.

franco-zalamena-iterable:
I like the idea of crashing on debug and just no-op on production. But the problem is that the BuildConfig.Debug references the sdk BuildConfig from what i saw, so this would be just ignored in the final version

PR #1077, opened 2026-07-27
[SDK-547] Reduce excess JWT auth token requests (timer race)

franco-zalamena-iterable:
## Summary
Addresses an Android-specific race that can inflate `IterableAuthHandler.onAuthTokenRequested()` calls and backend JWT generation, reported by a customer seeing ~4x more Android JWT requests than iOS ([SDK-547]).

### Fix — One refresh timer at a time (race)
`scheduleAuthTokenRefresh` used a non-atomic `isTimerScheduled` check-then-set. The refresh timer (timer thread), app foreground (main thread), and 401 retry (network thread) can hit it concurrently; each could pass the guard and schedule its own `TimerTask`. `clearRefreshTimer()` only cancels the `Timer` it holds, so extra tasks could remain active and each request another token.

Scheduling and clearing are now synchronized, and the scheduler tracks the exact `TimerTask` that owns the refresh slot instead of a boolean. Only the owning task may fire or release that slot. A stale task that runs after cancellation cannot clear or execute in place of its replacement. Duplicate scheduling continues to retain the first scheduled task, callback, and retry policy.

Refresh call sites now provide an `IterableAuthRefreshReason`. Structured logs record each schedule, skip, fire, cancellation, stale-task rejection, and scheduling error with that reason.

## Testing
- `IterableAuthRefreshOwnershipTest` — 5 deterministic tests covering eight-thread concurrent scheduling, stale-task ownership, duplicate callback and policy retention, scheduling-failure cleanup, and lifecycle refresh behavior while failure retries are paused.
- Full `iterableapi` unit suite passed locally.
- `./gradlew :iterableapi:checkstyle` passed.
- `git diff --check` passed.

## Scope note
This PR is intentionally limited to the refresh timer ownership race and its observability. It does not change public APIs, callback contracts, existing retry priority, crypto timeout handling, stored-token restoration, or identity coordination. Those concerns should be handled independently so any intended behavior changes can be reviewed and tested separately.

The exact ~4x production ratio is a fleet-scale and device-dependent effect, and was not reproduced 1:1 with a long-lived token across configurations. Full attribution still requires backend request segmentation and customer configuration details.

One known related divergence is deliberately not included: when a token is already inside its refresh window, Android schedules a retry at `getNextRetryInterval()` while iOS waits for the next foreground or API call. Choosing the canonical behavior is a cross-platform product decision and should remain separate from this concurrency fix.

[SDK-547]: https://iterable.atlassian.net/browse/SDK-547


iterableapi/src/main/java/com/iterable/iterableapi/IterableAuthManager.java
```diff
@@ -309,11 +312,13 @@ public void run() {
                     } else {
                         IterableLogger.w(TAG, "Email or userId is not available. Skipping token refresh");
                     }
-                    isTimerScheduled = false;
+                    synchronized (IterableAuthManager.this) {
+                        isTimerScheduled = false;
+                    }
```
rtlsilva: Because this unconditionally resets `isTimerScheduled`, an already running task can overwrite the flag after `clearRefreshTimer()` installs a replacement timer, allowing duplicate scheduling.

Suggest replacing the flag with the actual scheduled `TimerTask` - to ensure stale tasks cannot alter replacement state, the task should clear/release ownership, not perform an unconditional state reset itself.

iterableapi/src/main/java/com/iterable/iterableapi/IterableKeychain.kt
```diff
@@ -120,6 +129,13 @@ class IterableKeychain {
         val encryptedValue = sharedPrefs.getString(key, null) ?: return null
         return try {
             encryptor?.let { runWithTimeout { it.decrypt(encryptedValue) } }
+        } catch (e: TimeoutException) {
+            // A crypto operation that times out is transient (slow/contended AndroidKeyStore), not a
+            // corrupt key. Don't wipe stored credentials or disable encryption over it — that would
+            // force a re-login (and a new auth-token request) on every slow launch. Return null for
+            // this read; the encrypted value stays intact for the next attempt. (SDK-547)
+            IterableLogger.w(TAG, "Crypto operation timed out; keeping encrypted data for retry.")
+            null
```
rtlsilva: If this returns `null` on timeout, is that `null` assigned to the auth token?
If it is, then a decryption timeout can still cause a refresh to be scheduled, which means it doesn't really reduce JWT auth token requests.

Suggest not representing both missing credentials and a timeout with `null` - retry reading the credentials after a timeout, but only request a new JWT after a successful read confirms that the token is genuinely absent.

Conceptually:
```kotlin
  internal sealed interface KeychainReadResult {
      data class Value(val value: String?) : KeychainReadResult
      data object TimedOut : KeychainReadResult
  }

  KeychainReadResult result = keychain.readAuthData();

  if (result.timedOut()) {
      scheduleAuthDataRestoreRetry();
      return; // Do not schedule JWT refresh.
  }

  _email = result.email();
  _userId = result.userId();
  _authToken = result.authToken();

  if (_authToken == null && isInitialized()) {
      authManager.scheduleAuthTokenRefresh(...);
  }
```

rtlsilva (changes requested):

iterableapi/src/main/java/com/iterable/iterableapi/IterableAuthManager.java
franco-zalamena-iterable: Good catch — you're right, and the unconditional reset was the actual lost update, not just a theoretical one.

Fixed in `1f0071b4` by tracking the `TimerTask` itself (`scheduledRefreshTask`) instead of a boolean, exactly as you suggested. The task now releases the slot only if it still owns it:

```java
private synchronized void releaseRefreshTask(TimerTask task) {
    if (scheduledRefreshTask == task) {
        scheduledRefreshTask = null;
    }
}
```

The release also moved into a `finally` block, so a throwing `requestNewAuthToken` can't leave the slot permanently claimed and wedge refreshing altogether.

Two related cleanups the identity check made possible:
- `clearRefreshTimer()` now clears the slot unconditionally instead of only inside the `timer != null` branch — previously a null `timer` left a stale claim behind.
- Dropped the redundant `if (!isTimerScheduled)` pre-check in `queueExpirationRefresh`; the guard inside `scheduleAuthTokenRefresh` covers it, and reading the flag outside the lock was itself racy.

Regression test: `IterableAuthTokenLifecycleTest.aStaleRefreshTaskDoesNotFreeTheReplacementsSlot`. Verified it actually catches this — reverting `releaseRefreshTask` to the old unconditional clear fails it with `expected:<1> but was:<2>`, i.e. the second refresh does get armed.

iterableapi/src/main/java/com/iterable/iterableapi/IterableKeychain.kt
franco-zalamena-iterable: Yes — that null was assigned to the auth token, and you're right that fix 2 therefore did not deliver the reduction the PR title claims. Traced it:

`IterableApi.retrieveEmailAndUserId()` read `keychain.getAuthToken()`, and on timeout that returned `null`, which the very next branch reads as "no token is stored" → `scheduleAuthTokenRefresh(...)` → `onAuthTokenRequested()`. So on a slow-keystore launch the SDK asked the app to mint a brand new JWT while a perfectly valid one sat in SharedPreferences. Not wiping credentials stopped the forced re-login, but the extra request survived.

Implemented your suggested shape in `f9e6b100`, with your sealed interface as written:

```kotlin
sealed interface KeychainReadResult {
    data class Value(val value: String?) : KeychainReadResult
    data object TimedOut : KeychainReadResult
}
```

`readValue()` returns `TimedOut` where it previously returned `null`, and the restore path only requests a new token once a *completed* read confirms none is stored, or the retries are exhausted (2 attempts, 1 s apart).

One deviation from the sketch worth flagging: the retry can't be inline. `retrieveEmailAndUserId()` runs from `initialize()` on the main thread, so re-reading there would add up to 500 ms of main-thread block per attempt. `retryAuthTokenRead()` posts to the main looper with a delay instead, so `initialize()` returns immediately and the retry lands later.

The existing `getEmail()`/`getUserId()`/`getAuthToken()` accessors keep their `String?` signature (`valueOrNull()`), so nothing outside the restore path changed.

Also: the restore logic moved to a new `IterableAuthDataRestorer` because `IterableApi.java` was one line under checkstyle's 2000-line file limit — that's why the diff touches an extra file.

Covered by `IterableKeychainTimeoutRecoveryTest` (4 tests: timeout is not reported as absent; a retry that succeeds returns the stored token and requests nothing; exhausted retries fall back to requesting; a genuinely absent token still requests immediately). Mutation-checked — 3 of the 4 fail without the fix.

iterableapi/src/main/java/com/iterable/iterableapi/IterableAuthManager.java
```diff
@@ -291,30 +327,109 @@ long getNextRetryInterval() {
...
+                    return;
+                }
+
+                IterableLogger.d(TAG, "auth_refresh action=fire reason=" + reason);
+                if (api.getEmail() != null || api.getUserId() != null) {
+                    api.getAuthManager().requestNewAuthToken(
+                            false,
+                            successCallback,
+                            reason.ignoresRetryPolicy()
+                    );
```
rtlsilva: `claimRefreshTask()` currently clears `scheduledRefreshTask` before the identity check and `requestNewAuthToken()` call.

If `queueExpirationRefresh()` replaces the token during that window, `clearRefreshTimer()` cannot invalidate the old task, and it still calls `onAuthTokenRequested` after the replacement.

In other words, there is still a narrower interleaving to verify here - in the window between a timer being canceled / replaced and a refresh being performed, the old task is no longer identifiable as stale, so it may still issue the refresh.

Suggest leaving `scheduledRefreshTask` assigned during the identity check, then atomically validating and consuming ownership immediately before calling `requestNewAuthToken()`. Conceptually, something like:
```java
if (!isCurrentRefreshTask(this)) {
     return;
}

boolean hasIdentity = api.getEmail() != null || api.getUserId() != null;
if (!hasIdentity) {
    releaseRefreshTask(this);
    return;
}

// this should synchronously confirm scheduledRefreshTask == task before clearing it
if (!consumeRefreshTask(this)) {
    return; // cancellation or replacement won.
}

api.getAuthManager().requestNewAuthToken(...);
```
Also suggest testing this interleaving where the task pauses during the identity check and is replaced before dispatch in `IterableAuthRefreshOwnershipTest`.

rtlsilva (changes requested):

rtlsilva (approved):
LGTM

PR #1079, opened 2026-08-10
[SDK-566] Accept fractional seconds for expiringAuthTokenRefreshPeriod

franco-zalamena-iterable:
# 📝 Summary
> Android now accepts fractional seconds for `expiringAuthTokenRefreshPeriod`, matching iOS/RN/Flutter, and validates the value instead of silently scheduling refreshes after expiry.

## 📖 Description

> **Stacked on #1077** — base is `feature/SDK-547-jwt-timer-race`, not `master`. Both branches modify `IterableAuthManager.java` and the same CHANGELOG section, so stacking avoids a conflict. Merge #1077 first; this will auto-retarget to `master`.

### ⚠️ The ticket's premise is wrong — please read before reviewing

SDK-566 states that Android interprets `expiringAuthTokenRefreshPeriod` in **milliseconds** while the other SDKs use **seconds**, and asks for a breaking unit change plus a coordinated version bump.

**Android already uses seconds.** Verified against all four codebases:

| Platform | Public API | Unit | Default |
|---|---|---|---|
| Android | `Builder.setExpiringAuthTokenRefreshPeriod(...)` | **seconds** (`* 1000L` internally) | 60 s |
| iOS | `config.expiringAuthTokenRefreshPeriod: TimeInterval` | seconds | 60 s |
| React Native | `IterableConfig.expiringAuthTokenRefreshPeriod` | seconds | 60 s |
| Flutter | `IterableConfig.expiringAuthTokenRefreshPeriod` | seconds | 60 s |

The `60000L` the ticket cites is the Builder's **internal millisecond representation** of 60 s, not a public seconds-valued default. The `* 1000L` conversion has been there since 2020 (`090bff97f`, `2cc5f9800`).

Corroboration: the RN and Flutter bridges pass seconds straight into the Android setter. If the premise were true, every hybrid app would already have a 1000x bug.

**So making the requested change would *introduce* the bug the ticket aims to prevent.** It is deliberately not done here. No breaking change, no coordinated version bump.

### What the real gap was

Android accepted **whole seconds only** (`Long`), while iOS/RN/Flutter accept fractional seconds. `0.5` was not expressible. That is a genuine parity gap, and it is what this PR fixes.

**1. Fractional seconds (`Added`)**
New `setExpiringAuthTokenRefreshPeriod(double)` overload. The `Long` overload is deprecated and delegates to it, so existing callers are unaffected — source and binary compatible. Overload resolution verified for `120L`, a boxed `Long`, a bare `int`, and `0.5`; none are ambiguous.

**2. Input validation (`Fixed`)** — two real bugs, both reproduced:
- **Negative** → `-60` became `-60000` ms, and the period is *subtracted* when computing the refresh time, so the refresh was scheduled **60 s after** the token had already expired.
- **Overflow** → `Long.MAX_VALUE * 1000L` wraps to `-1000`, same past-expiry path.
- **Null** → NPE on unboxing the `@NonNull Long`.

Now logged and corrected in the SDK's existing log-and-continue style (throwing would be a new crash surface for an init-time setter): `null`/`NaN`/negative fall back to the 60 s default; values above ~10 years are clamped. **Zero stays valid** — it means "refresh once expired" — and is covered by a test so it isn't "fixed" later.

**3. Internal naming (`Changed`)**
The core confusion is that one name meant seconds publicly and milliseconds internally. Package-private carriers renamed to `...Millis` (`IterableConfig`, `IterableAuthManager`, `RetryPolicy`). All `final`/package-private — invisible outside the SDK. Public setter names and parameters are untouched.

## Follow-ups (not in this PR)

- **React Native** — the Android bridge reads the value with `optLong`, so `12.7` truncates to `12` while iOS keeps `12.7`. Fix is committed on `feature/SDK-566-rn-fractional-refresh-period` but **cannot compile until an `iterableapi` release carries the `double` overload** — RN pins `3.6.2`, which publishes only the `Long` overload (`javac`: `double cannot be converted to Long`). Sequenced after this ships.
- **Flutter** — unverified; `Extensions.kt` suggests it truncates the same way.

[SDK-566]: https://iterable.atlassian.net/browse/SDK-566?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


iterableapi/src/main/java/com/iterable/iterableapi/IterableConfig.java
```diff
@@ -341,15 +351,62 @@ public Builder setAuthRetryPolicy(@NonNull RetryPolicy retryPolicy) {
...
+         * <p>
+         * Defaults to 60 seconds. Fractional seconds are supported, matching the iOS, React Native
+         * and Flutter SDKs.
+         * <p>
+         * A token handed to the SDK with less remaining lifetime than this period is already inside
+         * its refresh window, which causes the SDK to request another token right away. Keep the
+         * period comfortably below the lifetime of the tokens the auth handler returns.
+         * <p>
+         * Invalid values are logged rather than throwing. Meaningless values fall back to the 60
+         * second default ({@code null}, {@code NaN}, negatives); values above ~10 years are clamped
```
rtlsilva: There seems to be a mismatch between docs and behavior - the `isNaN` / negative branches return `this` without assigning the field, so the field only equals the 60s default if it was never set earlier in the builder chain: if I wrote `setExpiringAuthTokenRefreshPeriod(30.0).setExpiringAuthTokenRefreshPeriod(-60.0)`, the field would still be 30000ms.

Non-blocking, but probably worth a precision rewording in the javadoc and changelog to "Invalid values are ignored" or similar if this is the intended behavior - if it is not, we'd need to change it so that the value gets reset back to the default in those branches, matching the current docs.
Probably should add a test to exercise this set-then-invalidate case either way.

CHANGELOG.md
```diff
@@ -3,9 +3,19 @@ All notable changes to this project will be documented in this file.
...
+- `setExpiringAuthTokenRefreshPeriod` now validates its input instead of silently producing a broken refresh schedule. Previously a negative value was converted to a negative millisecond period and then *subtracted* when computing the refresh time, scheduling the refresh after the token had already expired; a very large value overflowed to a negative period with the same effect; and `null` threw a `NullPointerException` on unboxing. Invalid values are now logged and corrected — `null`, `NaN` and negative values fall back to the 60 second default, and values above ~10 years are clamped to that ceiling. Zero remains valid and means the token is refreshed only once it has expired.
```
rtlsilva: > `null`, `NaN` and negative values fall back to the 60 second default

See [related comment](https://github.com/Iterable/iterable-android-sdk/pull/1079/changes#r3765093358)

iterableapi/src/main/java/com/iterable/iterableapi/IterableConfig.java
```diff
@@ -8,6 +8,16 @@
  *
  */
 public class IterableConfig {
+    private static final String TAG = "IterableConfig";
+
+    static final long DEFAULT_EXPIRING_AUTH_TOKEN_REFRESH_PERIOD_SECONDS = 60L;
+
+    /**
+     * Ceiling for {@link Builder#setExpiringAuthTokenRefreshPeriod(Long)}, in seconds (~10 years).
```
rtlsilva: This javadoc links to the deprecated `Long` overload, pointing to the `double` one would be more apt.
```suggestion
     * Ceiling for {@link Builder#setExpiringAuthTokenRefreshPeriod(double)}, in seconds (~10 years).
```

rtlsilva (changes requested):

iterableapi/src/main/java/com/iterable/iterableapi/IterableConfig.java
franco-zalamena-iterable: Good catch, confirmed — `setExpiringAuthTokenRefreshPeriod(30.0).setExpiringAuthTokenRefreshPeriod(-60.0)` leaves 30000ms, not the 60000 the javadoc claimed.

Went with the wording fix rather than changing the behaviour: if someone deliberately sets 30s and a later computed value comes through as -1, silently resetting to 60s throws away their explicit intent, so ignoring the bad input seems more defensible. Javadoc now says invalid values are ignored and the period keeps whatever it had, and calls out that the >10y clamp *does* assign, unlike the ignore branches.

Also updated the three log messages — they had the same "using default of 60s" overstatement, and that's what someone actually reads while debugging. They now report the value being kept, e.g. `ignoring it and keeping 30.0s`.

Added `invalidExpiringAuthTokenRefreshPeriodKeepsThePreviouslySetValue` (30.0 → -60.0 → NaN, asserts 30000). Mutation-checked it: making the negative branch reset to the default instead fails it with `expected:<30000> but was:<60000>`.

CHANGELOG.md
franco-zalamena-iterable: Reworded to match — invalid values are now described as ignored, leaving the period at whatever it was before the call, with the >10y clamp called out separately since it does assign.

iterableapi/src/main/java/com/iterable/iterableapi/IterableConfig.java
franco-zalamena-iterable: Agreed — you committed the suggestion in 467a669c, thanks.

rtlsilva (approved):
LGTM

PR #1083, opened 2026-08-24
[SDK-499] Synchronize EmbeddedSessionManager to fix embedded session races

### LINKED ISSUE #1052: NullPointerException in EmbeddedSessionManager.updateDisplayCountAndDuration() — thread-safety race condition

franco-zalamena-iterable:
# 📝 Summary

Makes `EmbeddedSessionManager` internally thread-safe, fixing a reported `NullPointerException` plus two further races in the same class.

## 📖 Description

`EmbeddedSessionManager` had no synchronization at all, and its impression API has no SDK-internal callers — it is driven entirely by integrator code, so it can be entered from any thread. Three races: the reported NPE in `updateDisplayCountAndDuration()` (null-check then `!!` on `start`, which `pauseImpression()` nulls in between), concurrent modification of the impression map, and duplicate session tracking when `endSession()` raced with itself.

All access to `session`, `impressions` and the impression fields now goes through one private lock. A single class-wide lock instead of a concurrent map because the invariants span several fields — `endSession()` has to end all impressions, snapshot the list and reset both fields as one step. `trackEmbeddedSession()` is called after the lock is released, since it re-enters `IterableApi` and integrator code.

`endSession()` still leaves the session running when there are no impressions. That is pre-existing behavior, changing it changes what gets reported, so it is tracked separately in SDK-701. A test pins the current behavior so that change is explicit when SDK-701 lands.

No public API change and no behavior change on the single-threaded path.

This development was started by contributor @Shamyyoun

## 🧪 How to test?

`./gradlew :iterableapi:testDebugUnitTest --tests "com.iterable.iterableapi.EmbeddedSessionManagerThreadSafetyTest"`

`concurrentSessionAndImpressionUpdatesDoNotThrow` races 8 threads over the session and impression API. Reverting the two source files to master reproduces the reported crash:

```
java.lang.AssertionError: concurrent access failed: [java.lang.NullPointerException] expected:<0> but was:<1>
```

Full `iterableapi` suite passed locally.

## 🧾 Changelog

- Fixed a `NullPointerException` in `EmbeddedSessionManager.updateDisplayCountAndDuration()` that could crash apps calling embedded session methods off the main thread. `EmbeddedSessionManager` is now internally synchronized, which also fixes concurrent modification of its impression map and duplicate session tracking when `endSession()` raced with itself. Thanks to [@Shamyyoun](https://github.com/Shamyyoun) for the report and initial fix.

N/A

#### 🐞 Github Issues solved

Addresses #1052, supersedes #1053. Deliberately not using a closing keyword yet so merging doesn't auto-close the issue before we reply to the reporter.

#### 📚 Docs PR if applicable

N/A

[SDK-499]: https://iterable.atlassian.net/browse/SDK-499?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/iterableapi/src/main/java/com/iterable/iterableapi/EmbeddedImpressionData.kt
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/EmbeddedImpressionData.kt
@@ -9,7 +9,7 @@ data class EmbeddedImpressionData(
     val placementId: Long,
     var displayCount: Int = 0,
     var duration: Float = 0.0f,
-    var start: Date? = null
+    @Volatile var start: Date? = null
 ) {
     constructor(
         messageId: String,
--- a/iterableapi/src/main/java/com/iterable/iterableapi/EmbeddedImpressionData.kt
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/EmbeddedImpressionData.kt
@@ -9,7 +9,7 @@ data class EmbeddedImpressionData(
     val placementId: Long,
     var displayCount: Int = 0,
     var duration: Float = 0.0f,
-    @Volatile var start: Date? = null
+    var start: Date? = null
 ) {
     constructor(
         messageId: String,
--- a/iterableapi/src/main/java/com/iterable/iterableapi/EmbeddedSessionManager.kt
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/EmbeddedSessionManager.kt
@@ -108,10 +108,11 @@ public class EmbeddedSessionManager {
     }
 
     private fun updateDisplayCountAndDuration(impressionData: EmbeddedImpressionData): EmbeddedImpressionData {
-        if (impressionData.start != null) {
+        val start = impressionData.start
+        if (start != null) {
             impressionData.displayCount = impressionData.displayCount.plus(1)
             impressionData.duration =
-                impressionData.duration.plus((Date().time - impressionData.start!!.time) / 1000.0)
+                impressionData.duration.plus((Date().time - start.time) / 1000.0)
                     .toFloat()
             impressionData.start = null
         }
--- a/iterableapi/src/main/java/com/iterable/iterableapi/EmbeddedSessionManager.kt
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/EmbeddedSessionManager.kt
@@ -108,13 +108,15 @@ public class EmbeddedSessionManager {
     }
 
     private fun updateDisplayCountAndDuration(impressionData: EmbeddedImpressionData): EmbeddedImpressionData {
-        val start = impressionData.start
-        if (start != null) {
-            impressionData.displayCount = impressionData.displayCount.plus(1)
-            impressionData.duration =
-                impressionData.duration.plus((Date().time - start.time) / 1000.0)
-                    .toFloat()
-            impressionData.start = null
+        synchronized(impressionData) {
+            val start = impressionData.start
+            if (start != null) {
+                impressionData.displayCount = impressionData.displayCount.plus(1)
+                impressionData.duration =
+                    impressionData.duration.plus((Date().time - start.time) / 1000.0)
+                        .toFloat()
+                impressionData.start = null
+            }
         }
         return impressionData
     }
--- a/iterableapi/src/main/java/com/iterable/iterableapi/EmbeddedSessionManager.kt
+++ b/iterableapi/src/main/java/com/iterable/iterableapi/EmbeddedSessionManager.kt
@@ -6,6 +6,10 @@ public class EmbeddedSessionManager {
 
     private val TAG = "EmbeddedSessionManager"
 
+    // Callers reach this class from arbitrary threads (see issue #1052), so every access to
+    // impressions, session, and the impression fields happens under this lock.
+    private val lock = Any()
+
     private var impressions: MutableMap<String, EmbeddedImpressionData> = mutableMapOf()
 
     var session: IterableEmbeddedSession = IterableEmbeddedSession(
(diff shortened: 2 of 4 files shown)


iterableapi/src/main/java/com/iterable/iterableapi/EmbeddedImpressionData.kt
```diff
@@ -9,7 +9,7 @@ data class EmbeddedImpressionData(
     val placementId: Long,
     var displayCount: Int = 0,
     var duration: Float = 0.0f,
-    var start: Date? = null
+    @Volatile var start: Date? = null
```
jferrao-itrbl: The `start` var is marked with `@Volatile` so every thread always sees the latest value. That extra marking is not needed, because the class already uses one lock for all of this data. It can make a later reader think the marking `@Volatile` is what stops the crash, when the lock is what actually does.

nit:  Remove the extra `@Volatile` marking on start, or add a short comment that the lock is what keeps this safe.

iterableapi/src/test/java/com/iterable/iterableapi/EmbeddedSessionManagerThreadSafetyTest.java
```diff
@@ -0,0 +1,177 @@
...
+        sessionManager = new EmbeddedSessionManager();
+    }
+
+    @Test
+    public void endSessionWithoutImpressionsLeavesSessionRunning() {
+        sessionManager.startSession();
+        sessionManager.endSession();
+
+        assertTrue(sessionManager.isTracking());
+    }
```
jferrao-itrbl, line 39: This test checks that ending a session with no viewed messages still leaves the session running. Though that is a known bug, kept on purpose in this change, the test never says that in the file so someone could think this is the intended product behaviour.

nit: Add a one-line comment on the test: this pins a current bug; flip the assertion when ticket SDK-701 is done.

iterableapi/src/main/java/com/iterable/iterableapi/EmbeddedSessionManager.kt
```diff
@@ -108,13 +127,14 @@ public class EmbeddedSessionManager {
     }
```
jferrao-itrbl: Maybe slightly out-of-scope but these two private methods walk and update the in-memory list of viewed messages without taking the lock themselves. Today that is fine, because the only caller already holds the lock. A later change could call them without the lock and bring races back.

nit: Add a short comment on both helpers: caller must already hold the lock.

jferrao-itrbl (approved):
Approved with very minor comments.

joaodordio (approved):

PR #1084, opened 2026-08-25
[SDK-500] Follow the system light/dark theme in HTML in-app messages

### LINKED ISSUE #1056: In-app HTML messages ignore system light/dark theme on Android

franco-zalamena-iterable:
# Summary

> HTML in-app messages now follow the host activity's Android light/dark configuration instead of always rendering as if the device were in dark mode.

Jira ticket: [SDK-500](https://iterable.atlassian.net/browse/SDK-500)

## Description

A campaign with an `@media (prefers-color-scheme: dark)` block rendered its dark styles even in light mode. WebView takes `prefers-color-scheme` from the `isLightTheme` attribute of its context's theme, and both in-app containers were pinned to a non-DayNight AppCompat theme (always dark). On top of that, the fragment path built its WebView from the host activity instead of the dialog, so fixing the theme alone did nothing.

Both containers now use a DayNight theme and the WebView gets the dialog's context. In the default `AUTOMATIC` mode, the in-app follows the host activity's Android UI mode.

Apps that keep their theme only in Compose state can configure `IterableConfig.Builder.setInAppColorScheme()` for a fixed override or `setInAppColorSchemeProvider()` for a value that is evaluated whenever an in-app message is created.

Documentation is staged on the `Iterable/iterable-docs` branch
`docs/sdk/september-release` in commit `b6cfb52cc`. The documentation PR will
be opened with the September SDK release.

Worth calling out: in `AUTOMATIC` mode, an app with a hard-coded light theme running on a dark-mode device gets a dark in-app. Apps that need the in-app to follow a separately managed theme can use the explicit override.

This development was started by contributor @AndrazP.

Addresses #1056 and supersedes #1057, which only covered the fragment path. The Dialog path used by Compose apps (#1015) had the same problem.

Docs branch created for the september release: https://github.com/Iterable/iterable-docs/tree/docs/sdk/september-release

<video width="320" height="240" controls src="https://github.com/user-attachments/assets/e979f798-c860-47b8-bb98-871e9dcfede0" type="video/mp4" ng-if="option=='video'"></video>

[SDK-500]: https://iterable.atlassian.net/browse/SDK-500?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFragmentHTMLNotification.java
```diff
@@ -123,7 +123,7 @@ public IterableInAppFragmentHTMLNotification() {
         this.backgroundAlpha = 0;
         this.messageId = "";
         insetPadding = new Rect();
-        this.setStyle(DialogFragment.STYLE_NO_FRAME, androidx.appcompat.R.style.Theme_AppCompat_NoActionBar);
+        this.setStyle(DialogFragment.STYLE_NO_FRAME, androidx.appcompat.R.style.Theme_AppCompat_DayNight_NoActionBar);
```
joaodordio: This is the right fix for apps that follow the system or use `AppCompatDelegate.setDefaultNightMode`, but it leaves Compose apps that keep their theme in Compose state with no way to influence it. 

A Compose app on a light-mode device with the user set to dark currently gets a dark in-app and will get a light one after this. The Dialog path exists for exactly those apps (#1015), so I don't think we can ship this without either an `IterableConfig` override for the in-app color scheme, or an explicit docs note telling those integrators to call `setDefaultNightMode`. Which one do you think it's better? Or am I missing something on the Compose support side?

joaodordio (changes requested):
Left a comment regarding the composer support and we also need a docs PR to cover these changes plz! 

Other than that, all good!

iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppFragmentHTMLNotification.java
franco-zalamena-iterable: Good catch. I tested the documentation-only `AppCompatDelegate.setDefaultNightMode` option against the pure `ComponentActivity` path, and the dialog stayed light, so that would not reliably cover Compose-managed theme state.

I added an explicit config API in c34ff0176:
- `setInAppColorScheme()` for a fixed `AUTOMATIC`/`LIGHT`/`DARK` value
- `setInAppColorSchemeProvider()` for Compose state; it is queried whenever an in-app container is created

The tests cover explicit overrides on both the Fragment and Dialog renderers and changing the provider value between two Dialog in-apps. The documentation is staged in `Iterable/iterable-docs` on `docs/sdk/september-release` at b6cfb52cc; its PR will be opened with the September SDK release.

joaodordio (approved):
LGTM

PR #1086, opened 2026-08-25
[SDK-673] Add switchProject for runtime project switching

joaodordio:
# 📝 Summary
Adds `IterableApi.switchProject(context, project, callback)`, moving a running app between Iterable projects in place with no restart and no state carried over.

###### 🎟️ Jira Ticket: [SDK-673](https://iterable.atlassian.net/browse/SDK-673)

> ⚠️ **Stacked PR.** This targets `feature/SDK-675-offline-disable-device`, not `master`. Review [#1085](https://github.com/Iterable/iterable-android-sdk/pull/1085) first and merge it first. SDK-673 depends on it: the "queued device disables survive the switch" behaviour is inert without the `offlineApiSet` change in SDK-675, because a `disableDevice` is not offline-queueable without it. Re-target this to `master` once SDK-675 lands.
>
> The iOS counterpart is [iterable-swift-sdk#1091](https://github.com/Iterable/iterable-swift-sdk/pull/1091) (SDK-674). The two were written together and the behaviour is intended to match.

## 📖 Description

For multi-region apps that need to move between a US and an EU project without restarting. Previously the only reliable way to change projects was an app restart.

The call returns immediately and runs the whole sequence on the background executor: disable the push token on the previous project, reset the in-app, embedded and unknown-user managers, purge the offline queue apart from queued device disables, clear identity and the rest of the previous project's storage, then re-initialize against the new key. The auth manager is rebuilt against the new config and the request processor re-bound to it, which iOS gets for free from its instance swap.

### The project pair

`switchProject` takes an `IterableProject`, which pairs a project's API key with the `IterableConfig` to run it with:

```java
IterableProject euProject = new IterableProject(euApiKey, new IterableConfig.Builder()
        .setAuthHandler(euAuthHandler)
        .setDataRegion(IterableDataRegion.EU)
        .build());
```

The two halves are paired because `IterableConfig` carries the `dataRegion` and the `IterableAuthHandler`. Passing a loose key and config allowed one project's key to be combined with another project's region or auth handler, and that mismatch surfaces as an auth failure, or a request sent to the wrong region, that looks unrelated to the project it came from. Pairing them makes it unrepresentable.

The constructor rejects a null or blank key with `IllegalArgumentException`, so an unusable project is not constructible and a blank key can no longer reach a teardown. The switcher keeps its own blank-key guard anyway, because the package-private entry point is also reachable from initialize-time paths and from Kotlin callers whose platform types can carry a null through.

The `(apiKey, config)` form is now the package-private switcher entry point rather than public API. Replaced rather than overloaded, because this is unreleased: keeping both public would leave the loose form available forever and reintroduce exactly what the pair removes. iOS mirrors this with `IterableProject` and a failable initializer, so the guarantee matches while each platform keeps its own idiom for reporting an invalid value.

### Callback contract

`IterableProjectSwitchCallback` is a single-method interface, `onProjectSwitched(IterableProjectSwitchResult result)`, delivered on the main thread.

`true` means every teardown step completed cleanly. `false` means the SDK **is** on the new project but a cleanup step was noisy, or no device disable could be confirmed. **`false` never means the switch failed or was rolled back.** An app that does not use push always sees `false`, which is normal and not an error. The right response is the same either way: carry on and re-identify the user.

It is a single-method interface deliberately. Reusing `IterableInitializationCallback` does not work: its only abstract method takes no arguments, so a lambda would bind to that one and silently discard the boolean.

### The result type

The callback reports an `IterableProjectSwitchResult`, either `SWITCHED_CLEANLY` or `SWITCHED_WITH_WARNINGS`, rather than a boolean.

The boolean was the problem the guide kept having to write around: `false` did not mean the switch failed, did not mean it was rolled back, was the expected outcome for any app without push, and called for exactly the same handling as `true`. A value needing that much explanation to avoid being misread is the wrong shape. Both case names lead with "switched" so the part that holds in either case, that the SDK is on the new project, is visible at the call site instead of being something the reader has to recall.

Internal plumbing still carries the teardown boolean the individual steps set, and converts once at the callback boundary via `IterableProjectSwitchResult.from`, so the change stays at the edge.

Two cases and no reason detail, deliberately. Java enums can gain constants later without breaking callers, so there is no cost to leaving room, and inventing a warning taxonomy before anyone has asked for one would be guesswork. If you want the noisy step identified in the result rather than only in the logs, say so on this PR.

### Details worth a reviewer's attention

**The previous project's device disable.** The disable captures that project's API key *and* its region endpoint when it is initiated, because the FCM token lookup is asynchronous and the live key can change underneath it. Without the captured key the disable lands on the new project, leaving the old project still delivering push to the device. Without the captured endpoint a cross-region switch sends the old key to the new region and is rejected. The switch waits up to 2 seconds for the disable to reach the request layer before swapping; if it times out the switch still completes and reports `false`, and the disable still reaches the project it was created for.

**Region binding for offline tasks.** Offline tasks now persist the endpoint they were created for, so a rehydrated request goes to the region it was built for rather than whichever region is live at flush time. Tasks already on disk from an earlier version keep resolving the old way, so nothing queued is lost on upgrade.

**`trackPushOpen` is not queued behind the switch gate.** A push payload carries the sending project's `campaignId`, `templateId` and `messageId`, so replaying it after the switch would report it against a project where those IDs do not exist. It runs inline instead. This needed care because the switch gate and the background-init gate are the same state on Android, so `queueOrExecuteUnlessSwitching` splits them: still queued during initialization, inline during a switch. iOS does not gate push handling at all, for the same reason.

**Gate consistency.** The longest `track` and `updateEmail` overloads were public and ran inline while every shorter overload was queued, so mid-switch behaviour depended on which overload the caller used. Both now wrap private `*Internal` methods, matching the existing `setEmail` / `setUserId` treatment.

**Per-project state on the shared instance.** iOS drops this when it replaces its SDK instance; Android reuses `sharedInstance`, so it has to be explicit. The switch now clears the inbox session ID, stored push payload, notification data and device attributes. The inbox session ID was the one producing cross-project data: it would otherwise be attached to the new project's first in-app tracking call. Device ID and visitor consent are project-agnostic and deliberately kept.

**Concurrency.** The gate check and enqueue are a single atomic step, so a call cannot pass the check just before the gate is raised and then run against a half torn-down SDK. Twelve fields are now `volatile` and `getAuthManager()`'s rebuild is guarded by a lock. The switch also recovers if the background executor is shut down when the teardown or drain is submitted, which could happen when `switchProject` was called from inside a switch callback.

**Guards.** A null `context` or `apiKey` throws `IllegalArgumentException`, since both are `@NonNull` and a null is a programmer error. An empty or whitespace-only key is a runtime condition, so it is refused without tearing anything down and reported as `false`, matching iOS.

### Known limitations, called out deliberately

- Campaign-attributed events other than push opens are still queued. `trackPurchase` with a `campaignId`, and `track` with a `campaignId`, are replayed against the new project carrying the previous project's IDs if they are issued during the switch window. iOS queues its campaign-carrying `trackPurchase` too, so this is consistent across platforms rather than an Android-only gap. Fixing it properly means per-call key binding on both SDKs and belongs in its own ticket.
- iOS has no in-flight-initialize guard, Android does. On iOS an `initialize` immediately followed by `switchProject` is torn down underneath. That divergence needs an iOS follow-up.
- Neither SDK persists switch intent, so a process death mid switch leaves the app restarted with the previous project's identity cleared and no record a switch was attempted.

## 🧪 How to test?

759 tests, 0 failures, checkstyle clean.

- `IterableSwitchProjectTest`, 43 tests: the guard cases, each teardown step, identity and storage clearing, manager rebuild, keychain rebuild, offline queue region binding, rapid and nested switches, a throwing teardown step, callback delivery and thread, blank keys, per-project instance state, and gate consistency for the longest overloads.
- `IterableSwitchProjectQueueDrainTest`, 4 tests: a drain no executor will accept, a switch started off the main thread, and the two push-open cases (inline during a switch, still queued during initialization).
- `IterableSwitchProjectDisableRegionTest`: the disable dispatch timeout path, now asserting the warnings result, driven with an overridden timeout so it does not burn real time.
- `IterableOfflineTaskRegionTest`, 13 tests: persisted `baseUrl`, rehydrated task region, old-schema fallback, cross-region isolation, disable preservation.
- `IterablePushRegistrationTaskTest`: the disable carries the captured key and endpoint rather than the live ones, and is sent before the key swap.

Every new test was checked to fail with its fix reverted, not just to pass alongside it.

Manual check: initialize against project A, identify a user, then call `switchProject` with project B's key from the callback of a region lookup. Confirm the device is disabled on A and registered on B, the inbox is empty immediately after, and no event reaches A after the callback.

## 🧾 Changelog

Added to `CHANGELOG.md` under Unreleased. One `Added` entry for `switchProject` with sub-bullets covering the callback contract, the queued and inline call sets, the state that is cleared and preserved, and the edge cases. Seven `Fixed` entries covering offline endpoint persistence, disable preservation across a switch, the overload gating, the in-flight-initialize deferral, executor recovery, the duplicate auth listener, and the atomic gate check.

Not recorded.

#### 🐞 Github Issues solved

None known.

#### 📚 Docs PR if applicable

**A docs PR is required and does not exist yet.** This adds public API. There is an adoption guide written for the first customer team, which should be the basis for the `iterable-docs` entry, and it needs to cover the callback contract, the per-platform queued call sets, and the iOS in-flight-initialize caveat.

---

**Note on the base:** this stack sits on `5d726694` and `origin/master` has moved 4 commits ahead, including a 3.10.1 release prep and [SDK-547](https://iterable.atlassian.net/browse/SDK-547), which touches JWT auth timing. Worth rebasing both branches onto latest master before merge, since SDK-547 is adjacent to the auth manager rebuild here.

[SDK-673]: https://iterable.atlassian.net/browse/SDK-673?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SDK-547]: https://iterable.atlassian.net/browse/SDK-547?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


iterableapi/src/main/java/com/iterable/iterableapi/IterableBackgroundInitializer.java
```diff
@@ -310,6 +428,255 @@ public String getDescription() {
...
+     *
+     * @param cleanTeardown false when a teardown step was noisy. The SDK is on the new project
+     *                      either way; this never means the switch failed.
+     */
+    static void completeProjectSwitch(boolean cleanTeardown) {
+        final List<IterableProjectSwitchCallback> switchCallbacksToNotify = new ArrayList<>();
+        final List<IterableInitializationCallback> initCallbacksToNotify = new ArrayList<>();
+        final ExecutorService executor;
+        synchronized (initLock) {
+            isSwitchingProject = false;
```
rtlsilva: The switching flags are currently cleared before the asynchronous drain starts, which could potentially allow newer calls to run before calls queued during the switch.

Consider keeping the gate raised until all queued calls have finished running.

[Related comment](https://github.com/Iterable/iterable-swift-sdk/pull/1091/changes#r3873060963) on the iOS PR

iterableapi/src/main/java/com/iterable/iterableapi/IterableProjectSwitcher.java
```diff
@@ -0,0 +1,356 @@
...
+            if (IterableBackgroundInitializer.runWhenInitialized(() -> switchProject(context, apiKey, config, callback))) {
+                IterableLogger.w(TAG, "switchProject called while initialization was still in flight; "
+                        + "deferring the switch until initialization completes");
+                return;
+            }
+        }
+
+        // Step 2: raise the switch gate synchronously, so calls made after this method returns are
+        // queued rather than executed against a half torn-down SDK.
+        if (!IterableBackgroundInitializer.beginProjectSwitch(callback)) {
```
rtlsilva: A second `switchProject` call with a different API key is currently ignored, although its callback still runs.
For example, if ProjectA -> ProjectB is running and the app requests ProjectA -> ProjectC, both callbacks fire but the SDK ends on B.

This may be surprising behavior for a multi-region API, but it also seems to be intentional and covered by tests, so this comment might be more of a design discussion than a blocker.

Should we clarify or reconsider the behavior when a second call targets a different API key?
For example, only joining requests targeting the same project and either queueing or rejecting requests targeting a different one?

[Related comment](https://github.com/Iterable/iterable-swift-sdk/pull/1091/changes#r3873055725) on the iOS PR

iterableapi/src/main/java/com/iterable/iterableapi/IterableProjectSwitcher.java
```diff
@@ -0,0 +1,356 @@
...
+        }
+
+        try {
+            SharedPreferences.Editor editor = api.getPreferences().edit();
+            editor.remove(IterableConstants.SHARED_PREFS_CRITERIA);
+            // matchedCriteriaId is nested inside the unknown-session payload cleared above rather
+            // than being a key of its own, but remove it too so a build that starts writing it
+            // separately cannot carry a previous project's criteria id across a switch.
+            editor.remove(IterableConstants.SHARED_PREFS_CRITERIA_ID);
+            editor.remove(IterableConstants.SHARED_PREFS_ATTRIBUTION_INFO_KEY + IterableConstants.SHARED_PREFS_OBJECT_SUFFIX);
```
rtlsilva, line 384: A deep-link redirect started on the previous project can finish after the switch and write its attribution into the new project’s shared storage.

Maybe the redirect could capture the project / API key it belongs to and discard the result if that project is no longer active?

[Related comment](https://github.com/Iterable/iterable-swift-sdk/pull/1091/changes#r3873237862) on the iOS PR

rtlsilva (changes requested):

iterableapi/src/main/java/com/iterable/iterableapi/IterableBackgroundInitializer.java
```diff
@@ -310,6 +456,380 @@ public String getDescription() {
...
+     * @return true if this call owns the switch, false if a switch was already in progress (the
+     *         callback is either registered with the in-flight switch or parked with the request
+     *         that will follow it)
+     */
+    static boolean beginProjectSwitch(@NonNull String apiKey,
+                                      @Nullable IterableConfig config,
+                                      @Nullable IterableProjectSwitchCallback callback) {
+        synchronized (initLock) {
+            if (isSwitchingProject) {
+                if (apiKey.equals(inFlightSwitchApiKey)) {
```
rtlsilva, line 567: Once a different destination is pending, a later request for the in-flight key cannot safely join the earlier switch.
For B in flight with C queued, a subsequent B request joins B's callbacks; those callbacks fire, then C runs and the SDK finishes on C even though B was requested last.
The global scan in `parkPendingSwitchLocked` similarly reorders non-adjacent duplicates such as C → D → C.

Suggest preserving request order and only coalesce an adjacent equivalent request.

iterableapi/src/main/java/com/iterable/iterableapi/IterableBackgroundInitializer.java
```diff
@@ -310,6 +456,380 @@ public String getDescription() {
...
+            completeProjectSwitch(false);
+            return;
+        }
+        IterableLogger.d(TAG, "switchProject: running the switch that was requested while the "
+                + "previous one was in flight");
+        // Posted rather than run inline, so it lands behind the callbacks for the switch that just
+        // finished, which were posted to the same looper. Whatever the app does in those, in
+        // particular re-identifying the user, then runs against the project it was told it was on
+        // before this teardown starts.
+        new Handler(Looper.getMainLooper()).post(() ->
```
rtlsilva, line 719: The inherited gate is still raised while B's callback runs.
When C is pending after A → B and B's callback follows the public contract by calling `setEmail` or `setUserId`, that call is queued; C starts next, and C's completion drains the identity call against C. This sends B's intended identity to the wrong project.

Suggest ensuring callback work for B executes before C takes over, or delay B's callback until its calls cannot be replayed into C.

rtlsilva (changes requested):

iterableapi/src/main/java/com/iterable/iterableapi/IterableBackgroundInitializer.java
rtlsilva: The callback path now applies B's re-identification before C starts, but lowering and draining the shared operation queue at every handover creates the inverse routing failure.

In A → B, then request C, then call `track` or `setEmail`, that call is queued while B is in flight and B's `completeProjectSwitch` drains it while `_apiKey` is B, before C starts. This violates the documented contract that calls between C's request and callback replay against C.

Suggest preserving callback-scoped work on B without draining operations intended for a later destination into B, for example by segmenting queued operations by requested destination.

iterableapi/src/test/java/com/iterable/iterableapi/IterableSwitchProjectQueueDrainTest.java
```diff
@@ -0,0 +1,284 @@
...
+            allThree.countDown();
+        }, "second");
+
+        IterableBackgroundInitializer.completeProjectSwitch(true);
+
+        assertTrue("every queued call must run", allThree.await(5, TimeUnit.SECONDS));
+        assertEquals("the late arrival must go to the back of the queue, not jump the replay",
+                Arrays.asList("first", "second", "arrivedDuringTheDrain"), order);
+        assertFalse("the gate must be down once the drain is finished",
+                IterableBackgroundInitializer.isSwitchingProject());
```
rtlsilva: `allThree` only proves that the operation bodies ran; it reaches zero before the drain thread performs the locked empty-queue check that lowers the gate.
This assertion therefore races drain completion - the current CI failed here while the same test can pass locally.

Suggest waiting for an actual drain-completion signal or polling the gate with a timeout before asserting it is down.

rtlsilva (changes requested):

PR #1088, opened 2026-09-03
[SDK-701] End session for embedded with no impressions now reset the session

franco-zalamena-iterable:
# 📝 Summary

## 📖 Description
This PR fixes what was already modified on iOS for embedded sessions when they start and have no impressions when ending the session.

Before it would keep the open session until we got impressions, even on background or similar actions

Now if the endSession is called with no impressions we reset the session so it can start anew when necessary

## 🧪 How to test?
One test was modified from pinning the previous behaviour to the new one, the name is endSessionWithoutImpressionsEndsRunningSession

[SDK-701]: https://iterable.atlassian.net/browse/SDK-701?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rtlsilva (approved):
No blocking findings. LGTM

PR #1089, opened 2026-09-10
[SDK-717] Fix stale customAction replay when handler returns false

joaodordio:
# 📝 Summary
Fix stale `IterableCustomActionHandler` replay when handler returns `false`, a regression present in 3.6.5–3.10.1.

## 📖 Description
The `IterableCustomActionHandler` interface documents its boolean return value as *"Reserved for future use"*. Clients commonly return `false` (Kotlin default, follows the javadoc), and the React Native SDK wrapper always does too. Since the SDK-307 fix (3.6.5, [PR #975](https://github.com/Iterable/iterable-android-sdk/pull/975)), returning `false` was treated as "action not consumed", leaving `pendingAction` alive and causing the handler to fire again on every foreground and every `initialize()` call.

Three replay paths:
- `onForeground()` calls `processPendingAction` on every app foreground
- `initialize()` called it twice — before and after `loadLastSavedConfiguration()`
- The trampoline's `super.onResume()` triggers `onActivityResumed → onForeground → processPendingAction` on the stale action before `handlePushAction` sets the new one

**`IterableActionRunner.java`** — `callCustomActionIfSpecified` now invokes the handler and always returns `true` (consumed). Returns `false` only when the handler is `null` (SDK not yet initialized), keeping the SDK-307 retry intent intact.

**`IterableApi.java`** — removed the duplicate `processPendingAction()` call before `loadLastSavedConfiguration()` in `initialize()`. One call after full initialization is enough.

## 🧪 How to test?
Added `testCustomActionHandlerReturnFalseDoesNotReplayOnForeground` to `IterablePushActionReceiverTest`: sends a push with a handler that returns `false`, simulates a foreground event via `processPendingAction`, and asserts the handler fired exactly once.

Existing `testBackgroundCustomActionProcessedAfterSDKInit` continues to cover the SDK-307 scenario (handler null at push time, retried after initialize).

## 🧾 Changelog
Added a `Fixed` entry to `CHANGELOG.md` under `[Unreleased]` calling out affected versions 3.6.5–3.10.1.

N/A

#### 🐞 Github Issues solved
N/A

#### 📚 Docs PR if applicable
N/A

[SDK-717]: https://iterable.atlassian.net/browse/SDK-717?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


franco-zalamena-iterable:
This solution makes the openApp fallback work correctly, but customHandlerPresent does not necessarily mean the custom handler was invoked. From what i checked, an unhandled openUrl action is also cleared when an unrelated custom-action handler is configured, preventing its foreground retry.

I though about adding improving readability with something internal like this 

enum ActionDispatchResult {
    NOT_DISPATCHED,
    DISPATCHED_WITH_FALLBACK,
    HANDLED
}

This would make it easier to distinguish when to clear the pending action or when to run it.

I think having this boolean when it actually can represent more things can be a bit confusing to debug and to understand what is the intended behavior (like the bug itself)

franco-zalamena-iterable (approved):

franco-zalamena-iterable:
I was checking this PR with some tests and realized that we introduced a new problem, currently if the app is closed and a push url is passed, we silently fail it. 

Currently on the initializeForPush we just pass the android context, not the iterable config.
Previously, we would return false in this case and then when the app was fully initialized it would be processed, but now it is returning DISPATCHED_WITH_FALLBACK
This is more specific to the openApp=false config.

We could rename NOT_DISPATCHED to RETRY_LATER and return it instead in this case, so a failed URL before full initialization should produce RETRY_LATER, not DISPATCHED_WITH_FALLBACK.

iterableapi/src/main/java/com/iterable/iterableapi/IterablePushNotificationUtil.java
```diff
@@ -55,30 +55,23 @@ static boolean processPendingAction(Context context) {
...
      */
     private static ActionDispatchResult dispatchPendingAction(Context context, PendingAction action) {
-        // Automatic tracking
+        // Defer until initialize() is called with the full SDK config. initializeForPush only
+        // sets _applicationContext; _apiKey is only set by initialize(). Checking _apiKey covers
+        // both URL actions (need urlHandler / deep link config) and custom actions (need
+        // customActionHandler). When the SDK is fully initialized, always dispatch and let
+        // IterableActionRunner return false naturally if no handler is configured — that
+        // correctly allows the openApp fallback to fire without keeping pendingAction alive.
+        if (IterableApi.sharedInstance._apiKey == null) {
```
franco-zalamena-iterable: I'm a bit reluctant of changing this to here, this will change the behavior on many calls, even if they are changing to something "more correct" i think we should stick to fixing the reported bug and add visibility and readability with this PR. Let me know what you think, i pushed a sibling branch with a similar solution but less "breaking". Let me know what you think

[proposal branch](https://github.com/Iterable/iterable-android-sdk/tree/franco-zalamena/pr-1089-proposal)

franco-zalamena-iterable (changes requested):

iterableapi/src/test/java/com/iterable/iterableapi/IterablePushActionReceiverTest.java
```diff
@@ -24,10 +24,13 @@
 import static junit.framework.Assert.assertEquals;
 import static junit.framework.Assert.assertNotNull;
 import static junit.framework.Assert.assertNull;
+import static junit.framework.Assert.assertTrue;
```
franco-zalamena-iterable: This is not used anymore

iterableapi/src/main/java/com/iterable/iterableapi/IterablePushNotificationUtil.java
```diff
@@ -19,27 +19,28 @@ static void clearPendingAction() {
...
         IterableApi.sharedInstance.setPayloadData(action.intent);
         IterableApi.sharedInstance.setNotificationData(action.notificationData);
         IterableApi.sharedInstance.trackPushOpen(action.notificationData.getCampaignId(), action.notificationData.getTemplateId(),
                 action.notificationData.getMessageId(), action.dataFields);
 
-        return IterableActionRunner.executeAction(context, action.iterableAction, IterableActionSource.PUSH);
+        return IterableActionRunner.dispatchAction(context, action.iterableAction, IterableActionSource.PUSH);
+    }
+
+    static boolean executeAction(Context context, PendingAction action) {
```
franco-zalamena-iterable: This is not used anymore from what i checked

franco-zalamena-iterable (commented):
Two things making checkstyle fail

franco-zalamena-iterable (approved):

joaodordio:
I've created the task [SDK-753](https://iterable.atlassian.net/browse/SDK-753) to tackle the flaky-ness of the tests that are separate from the work on this PR. 

[SDK-753]: https://iterable.atlassian.net/browse/SDK-753?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1093, opened 2026-09-16
[SDK-705] IterablePushRegistrationTask as Runnable instead of AsyncTask

franco-zalamena-iterable:
# Summary

Jira: [SDK-705](https://iterable.atlassian.net/browse/SDK-705)

Replaces push registration AsyncTask orchestration with SDK-owned executors while preserving production behavior and rapid login/logout ordering.

## Behavior

- Push token acquisition, registration, disable requests, and their retries use a dedicated SDK-owned serial lane.
- A newer register or disable action invalidates stale retries from an earlier action.
- Push client callbacks continue to run on the Android main thread.
- General online API requests retain their existing thread-pool execution path.
- JWT and HTTP 5xx retry behavior remains supported on the push lane.

## Compatibility

- No public API is added or changed.
- Test injection uses package-private constructors and interfaces; no VisibleForTesting state is required.
- The final diff adds no deprecated AsyncTask invocation.

## Verification

- Focused push registration, public-flow integration, and IterableApi tests pass.
- iterableapi-ui unit tests pass.
- Checkstyle, lint, UI assembly, root consumer tests, and Android test compilation pass.
- The full core suite has only the repository-wide external notification-image failure that also reproduces on master.
- The standalone sample reaches SDK compilation and then hits its existing Kotlin 1.8 versus 1.9 data-object mismatch.

[SDK-705]: https://iterable.atlassian.net/browse/SDK-705?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


iterableapi/src/main/java/com/iterable/iterableapi/IterablePushRegistration.java
```diff
@@ -1,19 +1,19 @@
...
+    IterablePushRegistration(Executor executor) {
+        this.executor = executor;
     }
 
-    static class IterablePushRegistrationImpl {
-        void executePushRegistrationTask(IterablePushRegistrationData data) {
-            new IterablePushRegistrationTask().execute(data);
-        }
+    void executePushRegistrationTask(IterablePushRegistrationData data) {
+        executor.execute(new IterablePushRegistrationTask(data));
```
rtlsilva, line 17: The new serial executor stops ordering the operations once ENABLE calls `registerDeviceToken()`, because that method starts another thread and returns.
A following DISABLE task can therefore enqueue `/disableDevice` before `/registerDeviceToken`, so a rapid login/logout sequence may leave the device enabled when registration lands last.

Suggest keeping registration request submission on this serial execution path and adding a deterministic enable-to-disable request-order test.

rtlsilva (changes requested):

iterableapi/src/main/java/com/iterable/iterableapi/IterablePushRegistration.java
franco-zalamena-iterable: thanks for the catch!

franco-zalamena-iterable:
Stack review update after the rebase:

- The ordering concern remains addressed: push actions and their register/disable HTTP submissions use the same dedicated serial push lane.
- Deterministic coverage protects enable-to-disable ordering and prevents stale registration retries after a newer push action.
- The existing review discussion is preserved.
- The remaining deprecated request AsyncTask findings are intentionally removed by stacked SDK-707 rather than duplicating that migration into SDK-705.

Check, BCIT, changelog, Java analysis, and the main build job passed. I have rerun the two failed jobs after confirming their failures were the external notification-image assertion and one instrumentation response timeout.

Ready for another review.

franco-zalamena-iterable:
Rerun follow-up: the unit job now fails only on the repository-wide external IterableNotificationTest.testNotificationImage assertion. The instrumentation rerun reproduced the lower-stack AsyncTask response timeout in IterableApiResponseTest; that suite passes on SDK-706 once the deep-link lane is separated and passes on SDK-707 after request AsyncTask is removed. The push ordering coverage and BCIT remain green.

franco-zalamena-iterable:
@rtlsilva SDK-705 is ready for re-review in 8d7e9e35.

The ordering concern is now handled across the complete push flow: registration and disable submissions, HTTP work, 5xx retries, and JWT retries remain on the SDK-owned serial push lane, while newer actions invalidate stale retries. General online requests are restored to their original thread-pool behavior, and push callbacks remain on main.

The previous CodeQL findings were caused by newly added deprecated AsyncTask calls; those additions are gone from the final diff. Focused behavior tests, UI unit tests, static checks, root consumer tests, and Android-test compilation pass. CI is now running on the updated commit.

iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java
```diff
@@ -694,22 +695,26 @@ void setAuthToken(String authToken, boolean bypassAuth) {
     }
 
     protected void registerDeviceToken(final @Nullable String email, final @Nullable String userId, final @Nullable String authToken, final @NonNull String applicationName, final @NonNull String deviceToken, final Map<String, String> deviceAttributes) {
-        if (deviceToken != null) {
-            if (!checkSDKInitialization() && _userIdUnknown == null) {
-                if (sharedInstance.config.enableUnknownUserActivation) {
-                    unknownUserManager.trackUnknownTokenRegistration(deviceToken);
-                }
-                return;
+        registerDeviceToken(email, userId, authToken, applicationName, deviceToken, deviceAttributes, runnable -> new Thread(runnable).start());
```
rtlsilva: This default path still starts a separate thread before submitting the registration request, so an immediately following `disablePush()` can enqueue first and overtake a manual `registerDeviceToken()` call.
That leaves the public registration path outside the dedicated serial registration/disable lane described by this PR.

Suggest submitting it through the push executor and adding deterministic `registerDeviceToken()` → `disablePush()` ordering test coverage.

rtlsilva (changes requested):

